PR #6, opened 2016-02-03
Work on data model (Types module)

fisx:
This implements the result of yesterday's call.

This breaks compilation, because I haven't updated module `Frontend` yet.  I would still like to merge this to master so we can use it as a basis for further brainstorming.


src/Types.hs
```diff
@@ -88,38 +139,63 @@ data Category =
...
   deriving (Eq, Ord, Show, Read, Generic)
 
--- | "Stimme"
-data Vote = Vote
-    { _voteMeta  :: MetaInfo
-    , _voteValue :: Maybe Bool
+-- | "Stimme" for "Idee"
+data CommentVote = CommentVote
+    { _commentVoteMeta  :: MetaInfo CommentVote
+    , _commentVoteValue :: Bool
```
np: I would suggest using a different type that Bool, it's rather clear True is positive and False negative. A type such as `data UpDown = Up | Down` would further remove any ambiguity.

src/Types.hs
```diff
@@ -88,38 +139,63 @@ data Category =
   deriving (Eq, Ord, Bounded, Enum, Show, Read, Generic)
 
 -- | "Verbesserungsvorschlag"
+--
+-- 'Comments' are hierarchical.  The application logic is responsible for putting some limit on the
+-- recursion depth under which all children become siblings.  (FIXME: or should we represent this in
+-- the types?)
```
np: I think is better to simply make that restriction in the application logic it could then be easier to change.

src/Types.hs
```diff
@@ -27,38 +27,63 @@ import qualified Data.Csv as CSV
...
     , _metaCreatedAt :: Timestamp
-    , _metaChangedBy :: GUID
+    , _metaChangedBy :: GUID User
     , _metaChangedAt :: Timestamp
     }
   deriving (Eq, Ord, Show, Read, Generic)
 
-newtype Article = Article { fromArticle :: [ST] }
+-- | Markdown content.
+newtype Document = Document { fromDocument :: ST }
```
np: Document is the right name for the type.
However the constructor could reflect the fact that the document is in markdown format by being called Markdown, this then remains a newtype as long as we have only one kind of formatting.

src/Types.hs
```diff
@@ -27,38 +27,63 @@ import qualified Data.Csv as CSV
...
 -- for different object types?)
-type GUID = Integer
+newtype GUID a = GUID Integer
+  deriving (Eq, Ord, Show, Read, Generic)
 
-data MetaInfo = MetaInfo
-    { _metaId        :: GUID
-    , _metaCreatedBy :: GUID
+data MetaInfo a = MetaInfo
+    { _metaId        :: GUID a
```
np: I like this use of phantom types.

np:
Appart from these minor changs it LGTM.

src/Types.hs
```diff
@@ -27,38 +27,63 @@ import qualified Data.Csv as CSV
 
 -- | Globally Unique ID (for reference in the database).  (FIXME: should we have different id types
 -- for different object types?)
-type GUID = Integer
+newtype GUID a = GUID Integer
```
fhartwig: Bikeshedding: I think we should call this something other than GUID, since [GUID has a specific meaning](https://en.wikipedia.org/wiki/Globally_unique_identifier).

src/Types.hs
```diff
@@ -88,38 +139,63 @@ data Category =
...
+-- 'Comments' are hierarchical.  The application logic is responsible for putting some limit on the
+-- recursion depth under which all children become siblings.  (FIXME: or should we represent this in
+-- the types?)
 data Comment = Comment
-    { _commentMeta    :: MetaInfo
-    , _commentArticle :: Article
-    , _commentVotes   :: Set Vote
+    { _commentMeta    :: MetaInfo Comment
+    , _commentText    :: Document
+    , _commentVotes   :: Set Comment
```
Mikolaj: aren't theses two actually swapped?

src/Types.hs
```diff
@@ -88,38 +139,63 @@ data Category =
...
+    , _commentReplies :: Set CommentVote
     }
   deriving (Eq, Ord, Show, Read, Generic)
 
--- | "Stimme"
-data Vote = Vote
-    { _voteMeta  :: MetaInfo
-    , _voteValue :: Maybe Bool
+-- | "Stimme" for "Idee"
```
Mikolaj: I don't know German, but should not this haddock be different than for IdeaVote?

src/Types.hs
```diff
@@ -68,16 +93,42 @@ data IdeaSpacePhase =
...
+    , _feasibleValue  :: Bool
+    , _feasibleReason :: Document
+    }
+  deriving (Eq, Ord, Show, Read, Generic)
+
+-- | FIXME: Is there a better name for 'Like'?  'Star'?  'Endorsement'?  'Interest'?
+data Like = Like
+    { _likeMeta  :: MetaInfo Like
+    }
+  deriving (Eq, Ord, Show, Read, Generic)
```
kosmikus: I'm slightly inclined to call this `IdeaLike` to mirror the use of `IdeaVote` ...

src/Types.hs
```diff
@@ -88,38 +139,63 @@ data Category =
...
   deriving (Eq, Ord, Show, Read, Generic)
 
 data User = User
-    { _userMeta           :: MetaInfo
-    , _userName           :: ST
-    , _userPassword       :: EncryptedPass
-    , _userEmail          :: Maybe Email
+    { _userMeta     :: MetaInfo User
+    , _userLogin    :: ST
+    , _userFullName :: (ST, ST)
```
kosmikus: I think we should either have a `FullName` datatype or two inlined fields for first and last name.

src/Types.hs
np: Agreed `_commentVotes :: Set CommentVote` and `_commentReplies :: Set Comment`

src/Types.hs
np: :+1:

src/Types.hs
Mikolaj: Also, I think we've just agreed it's not globally unique (and MetaInfo is inlined in the schema). Am I right?

src/Types.hs
fisx: Agreed.

(@andorp, you had a nice hack yesterday for rendering markdown in the browser.  Could you paste that here for later reference?  Or even somewhere appropriate in the code?  Thanks!)

src/Types.hs
fisx: @fhartwig: agreed.

@Mikolaj: right.

So...  `UID`?  I'll make it `UID`?  Sounds like user id.  `Handle`?  No!  Any ideas?

src/Types.hs
fisx: Right, I missed that spot.  Fixed.

This raises the question of whether we should have German in the source code?  I did it because it references the specs we have in German, but if anybody thinks we shouldn't I'll remove it.

src/Types.hs
fisx: I remember why I didn't do it now: it forced me to re-align the entire block, because `First` is longer than `Full`.  (-:

Fixed!

fisx:
Good feedback, thanks everyone!  Everything should be fixed now, including the things I didn't comment on.

src/Types.hs
fisx: > I'm slightly inclined to call this `IdeaLike` to mirror the use of `IdeaVote` ...

oops, missed that one.  i'll fix it on master.

src/Types.hs
andorp: @fisx: Here it is

https://github.com/andorp/bead/blob/1ed6ccf1cbc73be0c7b519ffe2522cc1569bfc0c/src/Bead/View/Markdown.hs#L28-L32

    markdownToHtml :: String -> Html
    markdownToHtml = either (fromString . show) (writeHtml def) . readMarkdown def . replaceCrlf

PR #7, opened 2016-02-04
Update Types.hs, add schema.sql

fhartwig:
This should cover everything from `Types.hs`, modulo things marked as TODO. Not tested yet, so it probably has a few syntax errors.


schema/schema.sql
```diff
@@ -1,6 +1,133 @@
...
+    ('wild_ideas', 'edit_topics', 'feasibility', 'vote', 'finished');
+
+CREATE TYPE category AS ENUM
+    ('rule', 'equipment', 'class', 'time', 'environment');
+
+CREATE TYPE group_id AS ENUM
+    ('admin', 'moderator', 'principal', 'student', 'guest', 'in_class');
+
+-- plural table name because 'user' is a sql keyword
+-- TODO: problem: how creates the creators?
```
fisx: the first user can create themselves?  it's philosophically confusing, and may be technically untidy because we need the `AUID` in order to allocate it, but if it's the first user, it's easy enough to come up with a fresh `AUID`.

PR #8, opened 2016-02-05
html vs. blaze

fisx:
stuck with a type puzzle.  it would be easy to take `MetaInfo` apart in the phantom-parameterized part and the one that carries access information, and that would solve my problem.  i just wonder if it also works like this.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,7 +1,7 @@
 .phony:
 
 sensei: .phony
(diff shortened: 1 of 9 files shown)


np:
What's the status of this PR? Is it ready for review?

fisx:
Sorry, it should have been assigned to me.  I'll talk about it in the call in a moment.  You can still take a look if you're curious, of course.

PR #9, opened 2016-02-08
Simple Travis Build

andorp:
The build on [travis](https://travis-ci.org/liqd/aula) now tries to install aula. There is a build state indicator in the README file.

NOTE: Somehow the travis cache is not working as expected, but until the build does not reach critical timelimit or number of pushes/hour. We can live with it. (E.g. The resolution could be done via docker image, I've it done before)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,23 @@
+language: c
+
+sudo: false
+
+addons:
+    apt:
+        sources:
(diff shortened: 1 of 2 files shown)


andorp:
We should add `cabal test` when we have tests.

fisx:
LGTM.  thanks!

PR #10, opened 2016-02-08
Make Page types a lot simpler

fisx:
...  and slightly more verbose.

The previous way of keeping different, related pages in one type was a bit clumsy to use and required too many extensions.

(We should probably split Frontend.Html up into several modules later.)


fisx:
in line

```haskell
deriving instance Read (PageAdminSettings a)
```

```
src/Frontend/Html.hs:186:1: Warning:
    Couldn't match type ‘a’ with ‘'AdminSettingsEventsProtocol’
      ‘a’ is a rigid type variable bound by
          the instance declaration at src/Frontend/Html.hs:186:1
    Expected type: PageAdminSettings a
      Actual type: PageAdminSettings 'AdminSettingsEventsProtocol
    Relevant bindings include
      readPrec :: Text.ParserCombinators.ReadPrec.ReadPrec
                    (PageAdminSettings a)
        (bound at src/Frontend/Html.hs:186:1)
    In the first argument of ‘return’, namely
      ‘PageAdminSettingsEventsProtocol’
    In the expression: return PageAdminSettingsEventsProtocol
    When typechecking the code for  ‘GHC.Read.readPrec’
      in a derived instance for ‘Read (PageAdminSettings a)’:
      To see the code I am typechecking, use -ddump-deriv
```

src/Frontend/Html.hs
```diff
@@ -1,7 +1,9 @@
 {-# LANGUAGE DataKinds          #-}
 {-# LANGUAGE FlexibleInstances  #-}
+{-# LANGUAGE GADTs              #-}
```
np: Is this still required?

src/Frontend/Html.hs
```diff
@@ -1,7 +1,9 @@
...
 {-# LANGUAGE KindSignatures     #-}
 {-# LANGUAGE OverloadedStrings  #-}
+{-# LANGUAGE StandaloneDeriving #-}
```
np: Is this still required?

np:
I don't have any other remarks

fisx:
wasn't.  thanks.

PR #12, opened 2016-02-08
ToMarkup (fragment)

fhartwig:
Sorry to dump this on someone in this state, I haven't gotten very far and I suck at HTML :-/
Nevertheless, here it is in case someone wants to pick this up.


fisx:
Thanks, this is not done, but useful!

PR #13, opened 2016-02-08
app logic experiments

fisx:


src/Api/NoPersistent.hs
```diff
@@ -0,0 +1,52 @@
...
+import Control.Monad.IO.Class
+import Control.Concurrent.MVar
+import Control.Lens
+import Data.String (fromString)
+import Network.Wai.Handler.Warp (runSettings, setHost, setPort, defaultSettings)
+import Servant
+import Servant.HTML.Blaze
+import System.IO.Unsafe (unsafePerformIO)
+
+import Thentos.Frontend.Types
```
andorp: I am going to set up docker based testing on travis to avoid the recompilation of thentos and cabal packages.

src/Api/NoPersistent.hs
```diff
@@ -0,0 +1,48 @@
...
+{-# LANGUAGE RankNTypes            #-}
+{-# LANGUAGE ScopedTypeVariables   #-}
+{-# LANGUAGE TemplateHaskell       #-}
+{-# LANGUAGE TypeOperators         #-}
+{-# LANGUAGE ViewPatterns          #-}
+
+{-# OPTIONS_GHC #-}
+
+module Api.NoPersistent
+where
```
andorp: I suggest a rename this module to `Persistent` and create an abstract monad interface controlled via member visibility list, that Andres Löh mentioned. Even if we use this as prototyping, as it will help us to see how Persistent will fit into the big picture.

src/Api/NoPersistent.hs
```diff
@@ -0,0 +1,48 @@
...
+
+data AulaData = AulaData
+    { _dbIdeas :: [Idea]
+    }
+  deriving (Eq, Show, Read)
+
+makeLenses ''AulaData
+
+aulaState :: TVar AulaData
+aulaState = unsafePerformIO . newTVarIO $ AulaData []
```
andorp: The visibility list should hide this detail.

fisx:
@kosmikus, @np, @Mikolaj: could you all take a quick look and see if this is a good representation of our correspondences / conversations?  Thanks~

andorp:
Note for the future: we should write some Persistent tests.

andorp:
LGTM

Mikolaj:
Yes, it seems faithful vs the discussion. Thank you for summarizing!

PR #16, opened 2016-02-10
2016-02-09-ap-types-and-frontend-tests

andorp:
Contains:
 * Checks against the inverse property for hand written `TimeStamp`
 * Checks for `ToMarkup` rendering does not contain bottoms
 * Two fixes 
 * Placeholder `InitSpec.hs` is removed


fisx:
LGTM.

PR #18, opened 2016-02-10
Use lucid, not blaze

### LINKED ISSUE #14: replace blaze with lucid

fisx:
Fixes #14

### DIFF
--- a/aula.cabal
+++ b/aula.cabal
@@ -27,21 +27,19 @@ library
     , thentos-core
     , aeson
     , binary
-    , blaze-builder
-    , blaze-html
-    , blaze-markup
     , bytestring
     , case-insensitive
     , cassava
     , containers
     , cookie
     , digestive-functors
-    , digestive-functors-blaze
+    , digestive-functors-lucid
     , directory
     , filepath
     , functor-infix
     , http-types
     , lens
+    , lucid
     , mtl
     , postgresql-simple
     , pretty-show
@@ -55,7 +53,7 @@ library
     , safecopy
     , scrypt
     , servant
-    , servant-blaze
+    , servant-lucid
     , servant-server
     , string-conversions
     , text
@@ -103,14 +101,13 @@ executable aula-html-dummies
     , aula
 
     , binary
-    , blaze-html
-    , blaze-markup
     , bytestring
     , cassava
     , containers
     , cookie
     , directory
     , lens
+    , lucid
     , mtl
     , postgresql-simple
     , pretty-show
@@ -119,7 +116,6 @@ executable aula-html-dummies
     , safecopy
     , scrypt
     , servant
-    , servant-blaze
     , servant-server
     , string-conversions
     , text
@@ -132,7 +128,6 @@ executable aula-html-dummies
     , warp
     , process
     , filepath
-    , blaze-builder
   default-language:
       Haskell2010
 
@@ -151,7 +146,7 @@ test-suite tests
       aula
     , base >=4.8 && <4.9
     , binary >= 0.7.5 && <0.8
-    , blaze-markup >=0.7 && <0.8
-    , blaze-html >= 0.8 && <0.9
     , hspec
+    , lucid
     , QuickCheck >= 2.8 && <2.9
+    , text
--- a/exec/RenderHtml.hs
+++ b/exec/RenderHtml.hs
@@ -1,34 +1,24 @@
-{-# LANGUAGE DataKinds             #-}
-{-# LANGUAGE FlexibleContexts      #-}
-{-# LANGUAGE FlexibleInstances     #-}
-{-# LANGUAGE OverloadedStrings     #-}
 {-# LANGUAGE ScopedTypeVariables   #-}
-{-# LANGUAGE TemplateHaskell       #-}
-{-# LANGUAGE TypeOperators         #-}
-{-# LANGUAGE ViewPatterns          #-}
 
 {-# OPTIONS_GHC -Werror -Wall #-}
 
 
 module Main (main) where
 
-import Data.Maybe
 import Control.Exception
 import Control.Lens ((^.))
 import Control.Monad (forM_)
+import Data.Maybe
 import Data.String.Conversions
 import Data.Typeable
-import Test.QuickCheck
-import Text.Blaze
-import System.Process
+import Lucid
+import System.Directory
 import System.Environment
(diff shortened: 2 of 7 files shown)


fisx:
I ran into this issue: `ToHtml` has two methods, one for escape-then-render, one for render-without-escape.  The differences between the two are usually trivial, but I have to implement both for every instance.

Will look into it and then elaborate~

fisx:
To be more specific, lucid does this:

```haskell
instance ToHtml Text where
  toHtml    = build . Blaze.fromHtmlEscapedText
  toHtmlRaw = build . Blaze.fromText
```

If I write a `ToHtml` instance for some container type, it looks like this:

```haskell
instance (ToHtml body) => ToHtml (Frame body) where
    toHtml (Frame bdy) = do
        head_ $ do
            title_ "AuLA"
            link_ [rel_ "stylesheet", href_ "/screen.css"]
        body_ $ do
            headerMarkup >> toHtml bdy >> footerMarkup
```

and produces this error:

```haskell
aula/aula/src/Frontend/Html.hs:40:10: Warning:
    No explicit implementation for
      ‘toHtmlRaw’
    In the instance declaration for ‘ToHtml (Frame body)’
```

I want a way of saying: "do the obvious thing: either use `toHtml` or `toHtmlRaw`, whichever we're defining."

fisx:
Some good, but ineffective ideas:

```haskell
instance (ToHtml body) => ToHtml (Frame body) where
    (toHtml, toHtmlRaw) = undefined

aula/aula/src/Frontend/Html.hs:41:5:
    Pattern bindings (except simple variables) not allowed in instance declarations
      (toHtml, toHtmlRaw)
```

```haskell
instance (ToHtml body) => ToHtml (Frame body) where
    x = 43
    toHtml = undefined

aula/aula/src/Frontend/Html.hs:41:5:
    ‘x’ is not a (visible) method of class ‘ToHtml’
```

andorp:
One solution for this puzzle could be:

````haskell
instance ToHtml PageComment where
    toHtml    = pageCommentToHtml toHtml
    toHtmlRaw = pageCommentToHtml toHtmlRaw

pageCommentToHtml g (PageComment comment) =
    div_ $ do
        div_ $ do
            span_ . g . AuthorWidget $ comment ^. commentMeta
            span_ . g . VotesWidget  $ comment ^. commentVotes
        div_ $ do
            g $ comment ^. commentText
        div_ $ do
            span_ "[antworten]"
            span_ "[melden]"
````

**Comment by fisx** [comment]
Or this:

```haskell
class ToHtmlSimple a where
    toHtmlSimple :: Monad m => (a -> HtmlT m ()) -> a -> HtmlT m ()

instance ToHtmlSimple a => ToHtml a where
    toHtml = toHtmlSimple toHtml
    toHtmlRaw = toHtmlSimple toHtmlRaw
```

This triggers some overlapping issues, but it seems like this could work.

fisx:
@Mikolaj, @kosmikus, @np: any opinions?  better ideas?  I feel the safest bet would be @andorp's suggestion.

Mikolaj:
How about 

```toHtml = toHtmlRaw```

**Comment by fisx** [comment]
> ```toHtml = toHtmlRaw```

that means no escaping at all.  this works as long as we don't allow users to enter unparsed code, which we probably don't want to.

sounds like a good idea.  and much easier than ours!  :-)

**Comment by np** [comment]
I would say `toHtmlRaw` only makes sense for string-like types. For others `toHtmlRaw` can be the same as `toHtml`. As a matter of choice I would rather define properly `toHtml` and then have `toHtmlRaw = toHtml`.

In a way `toHtmlRaw = toHtml` could be a default definition in the `ToHtml` class. But this is a just guess by looking at the code given the lack of documentation.

**Comment by fisx** [comment]
Ok, I went for the `toHtmlRaw = toHtml` option.  Some string values wrapped in our types may now be unintentionally be escaped, but I expect less trouble from that then from forgetting to escape somewhere.

This should be ready for review.

**Comment by np** [comment]
I've had a quick look at the changes and while it all seems fine. It's not really striking as an improvement.

Pros:
* The attributes (using `[...]`) are a bit nicer than `!`.
* It's dispensing from the calls to `text`.

Cons:
* `toHtml` vs `toHtmlRaw`
* `Html ()` instead of `Html` (the transformer is `HtmlT` the monad could have been `HtmlM` leaving `Html = HtmlM ()`
* The massive use of overloading makes the types less useful as a documentation

**Comment by fisx** [comment]
More thoughts on blaze vs. lucid from `freenode#thentos`:

```
11:16 < fisx_> re. lucid, i think the biggest improvement is that it doesn't pretend it's a monad, but crashes if you try to use `>>=` instead of `>>`.
11:16 < npouillard> fisx_: I know but did you ever run into it?
11:16 < fisx_> i did have a bit of vertigo during the refactoring, but i feel the small syntactic improvements are also quite important here.
11:16 < fisx_> andor did.
11:17 < fisx_> it blocks a nice way of implementing multi-language.
11:17 < fisx_> namely, adding a reader to it.
11:17 < npouillard> good argument
11:18 < fisx_> for me the trade-off seems to be lucid=clean design blaze=slightly faster
11:18 < npouillard> the sadest part is this toHtmlRaw it really looks to not have been thought through
11:18 < fisx_> not sure about the "slightly", haven't seen the benchmarks.
```

andorp:
> 11:16 < fisx_> re. lucid, i think the biggest improvement is that it doesn't pretend it's a monad, but crashes if you try to use `>>=` instead of `>>`.
> 11:16 < npouillard> fisx_: I know but did you ever run into it?

It is not a big deal, until one has the idea to put it into a MonadTransformer, at that point bind will be evaluated and throw an error. So intead of `ReaderT r HTML a` one has to tackle with `Reader r HTML` which produces annoying boilerplate and non well composable code like:

````haskell
snippet :: Reader I18N HTML
snippet = do
  i18n <- ask
  return $ do
     p $ i18n "Translated text"
````

Also put the HTML templates into a reader transformer gives us the power to reuse the same templates for users with different roles, something like this:

````haskell
-- with pseudo typing
home :: Reader UserRole HtmlM ()
home = do
  role <- ask
  when (isAdmin role) $ div $ p "Admin Tasks"
  when (isStudent role) $ div $ p "Idea pool"
````

Which can be also defined in different ways. We will decide how to do that.

PR #19, opened 2016-02-10
Extend the DB with users and current user

np:
Additions to Api.Persistent:

* getDb
* addDb
* modifyDb
* getUsers
* addUser
* findUserByLogin
* loginUser
* dbUsers
* dbCurrentUser

New endpoints in Frontend:

* GET users/create_random
* GET users
* GET login/:login

Frontend.Html:

* Add PageShow which enables to write dev pages using the underlying Show instance.

Types:

* Add lenses for direct access to the meta data of users and ideas.

### DIFF
--- a/src/Api/Persistent.hs
+++ b/src/Api/Persistent.hs
@@ -9,12 +9,25 @@
 
 module Api.Persistent
     ( Persist
+    , AulaData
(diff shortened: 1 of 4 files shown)


fisx:
I like this!

PR #23, opened 2016-02-11
Some tweaks to make html hacking more convenient.

fisx:


src/Frontend/Core.hs
```diff
@@ -0,0 +1,16 @@
...
+where
+
+import Data.Typeable
+import Data.String.Conversions
+import Lucid
+import Lucid.Base
+
+
+semanticDiv :: forall m a. (Monad m, ToHtml a, Typeable a) => a -> HtmlT m () -> HtmlT m ()
+semanticDiv t = div_ [makeAttribute "data-aula-type" (cs . show . typeOf $ t)]
```
andorp: This will generate the following snippet:
````html
<div data-aula="PageIdea"> ... </div>
````

Which serves two purposes:
 * It helps the front-en developer to identify which part of the generated pages comes from which combinator
 * Later on when we write selenium suite, the semantic tags helps up to parse, identify and test elements on the page.

andorp:
````haskell
semanticDiv :: forall m a. (Monad m, ToHtml a, Typeable a) => a -> HtmlT m () -> HtmlT m 
semanticDiv t = div_ [makeAttribute "data-aula-type" (cs . show . typeOf $ t)]
````
This will generate the following snippet:
````html
<div data-aula="PageIdea"> ... </div>
````

Which serves two purposes:
 * It helps the front-en developer to identify which part of the generated pages comes from which combinator
 * Later on when we write selenium suite, the semantic tags helps up to parse, identify and test elements on the page.

fisx:
@andorp: right.  thanks.  will add that to the code.

np:
LGTM

andorp:
LGTM

PR #27, opened 2016-02-12
Create Idea Page

andorp:
It contains functionality to create an idea starting from the ideas page. After idea creation it redirects to the ideas page.

### DIFF
--- a/src/Api/Persistent.hs
+++ b/src/Api/Persistent.hs
@@ -1,11 +1,11 @@
-{-# LANGUAGE TypeOperators               #-}
 {-# LANGUAGE GeneralizedNewtypeDeriving  #-}
 {-# LANGUAGE ImpredicativeTypes          #-}
 {-# LANGUAGE OverloadedStrings           #-}
 {-# LANGUAGE ScopedTypeVariables         #-}
 {-# LANGUAGE TemplateHaskell             #-}
+{-# LANGUAGE TypeOperators               #-}
 
-{-# OPTIONS_GHC -Wall -Werror #-}
+{-# OPTIONS_GHC -Werror #-}
 
 module Api.Persistent
     ( Persist
@@ -23,11 +23,15 @@ module Api.Persistent
     , dbIdeas
     , dbUsers
     , dbCurrentUser
+    -- FIXME: Remove hack
+    , forceLogin
     )
 where
 
 import Data.Foldable (find)
+import Data.Maybe (fromJust)
 import Data.String.Conversions
+import Data.Time.Clock (getCurrentTime)
 import Control.Concurrent.STM
 import Control.Monad.Trans.Reader
 import Control.Lens
@@ -36,18 +40,20 @@ import Servant.Server ((:~>)(Nat))
 
 import Types
 
+import qualified Data.Set as Set (empty)
 
 data AulaData = AulaData
     { _dbIdeas :: [Idea]
     , _dbUsers :: [User]
     , _dbCurrentUser :: Maybe (AUID User)
+    , _dbLastId :: Integer
     }
   deriving (Eq, Show, Read)
 
 makeLenses ''AulaData
 
 emptyAulaData :: AulaData
-emptyAulaData = AulaData [] [] Nothing
+emptyAulaData = AulaData [] [] Nothing 0
 
 -- | FIXME: call this type 'Action'?  Or 'Aula'?  Or 'AulaAction'?  As of the time of writing this
 -- comment, it doesn't make sense to have separate abstractions for persistence layer (Transaction
@@ -67,19 +73,25 @@ getDb l = Persist . ReaderT $ fmap (view l) . atomically . readTVar
 modifyDb :: Lens' AulaData a -> (a -> a) -> Persist ()
 modifyDb l f = Persist . ReaderT $ \state -> atomically $ modifyTVar' state (l %~ f)
 
-addDb :: Lens' AulaData [a] -> a -> Persist ()
-addDb l a = modifyDb l (a:)
+addDb :: (HasMetaInfo a, FromPrototype a) => Lens' AulaData [a] -> Prototype a -> Persist a
+addDb l pa = do
+    u  <- currentUser
+    i  <- nextId
+    mi <- liftIO $ newMetaInfo u i
+    let a = (metaInfo .~ mi) $ fromPrototype pa
+    modifyDb l (a:)
+    return a
 
 getIdeas :: Persist [Idea]
 getIdeas = getDb dbIdeas
 
-addIdea :: Idea -> Persist ()
+addIdea :: Prototype Idea -> Persist Idea
 addIdea = addDb dbIdeas
 
 getUsers :: Persist [User]
 getUsers = getDb dbUsers
 
-addUser :: User -> Persist ()
+addUser :: User -> Persist User
 addUser = addDb dbUsers
 
 findUserByLogin :: ST -> Persist (Maybe User)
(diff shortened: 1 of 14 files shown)


src/Api/Persistent.hs
```diff
@@ -1,11 +1,11 @@
-{-# LANGUAGE TypeOperators               #-}
 {-# LANGUAGE GeneralizedNewtypeDeriving  #-}
 {-# LANGUAGE ImpredicativeTypes          #-}
 {-# LANGUAGE OverloadedStrings           #-}
 {-# LANGUAGE ScopedTypeVariables         #-}
 {-# LANGUAGE TemplateHaskell             #-}
+{-# LANGUAGE TypeOperators               #-}
 
-{-# OPTIONS_GHC -Wall -Werror #-}
+{-# OPTIONS_GHC -Werror #-}
```
andorp: Orphan instances: DB Hack needs to pass that.

src/Config.hs
```diff
@@ -19,7 +19,7 @@ makeLenses ''Config
 
 config = Config
     { _dbPath = "./aula.db"
-    , _listenerInterface = "127.0.0.1"
+    , _listenerInterface = "0.0.0.0"
```
andorp: Running aula-server from docker needs to listen on `0.0.0.0`. In my opinion we should use config files for even this small config options.

src/Frontend/Page/CreateIdea.hs
```diff
@@ -0,0 +1,77 @@
...
+            , (CatEquipment  , "Ausstattung")
+            , (CatClass      , "Unterricht")
+            , (CatTime       , "Zeit")
+            , (CatEnvironment, "Umgebung")
+            ]
+
+    p2 :: ProtoIdea -> ExceptT ServantErr IO ST
+    p2 idea = liftIO $ do
+        runPersist $ do
+            forceLogin 1 -- FIXME: Login hack
```
andorp: We should work on login functionality.

src/Persistent.hs
```diff
@@ -0,0 +1,12 @@
+{-# OPTIONS_GHC -Werror #-}
+
+module Persistent (module Api.Persistent, runPersist)
+where
```
andorp: I wanted to separate the active persistent handler from the API of Persistent. It had to be placed into a different module to avoid a circular dependency.

andorp:
Also, please fix my English :)

src/Api/Persistent.hs
fisx: You could also add `-fno-warn-orphans` and leave `-Wall` intact, but it's all good.  We have travis.

src/Config.hs
fisx: then you can remove the perl line from README.md

src/Frontend/Core.hs
```diff
@@ -12,6 +13,61 @@ import Lucid
 import Lucid.Base
 import Network.Wai.Internal (Response(ResponseFile, ResponseBuilder, ResponseStream, ResponseRaw))
 import Network.Wai (Middleware)
+import Text.Digestive.View
+
+----------------------------------------------------------------------
+-- building blocks
```
fisx: having this in Frontend.Core is fine.  alternatives:

1. leave in Frontend.Html
1. move to Frontend.PageCore
1. move to Frontend.Page.Core

i'm ok with all, but don't like the fact that Frontend.Page.Core makes it look like it's just another page.

src/Types.hs
```diff
@@ -23,6 +24,14 @@ import Lucid
 import qualified Database.PostgreSQL.Simple.ToField as PostgreSQL
 import qualified Data.Csv as CSV
 
+----------------------------------------------------------------------
+-- prototypes of types
+
+type family Prototype type_ :: *
+
+-- HACK
```
fisx: not really a hack?

src/Types.hs
```diff
@@ -23,6 +24,14 @@ import Lucid
 import qualified Database.PostgreSQL.Simple.ToField as PostgreSQL
 import qualified Data.Csv as CSV
 
+----------------------------------------------------------------------
+-- prototypes of types
+
+type family Prototype type_ :: *
```
fisx: call this `Proto`?

src/Types.hs
```diff
@@ -159,6 +179,9 @@ data User = User
     }
   deriving (Eq, Ord, Show, Read, Generic)
 
+-- FIXME: Temporal hack to be able to save users.
```
fisx: english: "temporary".  "temporal" would mean you are hacking time.  (-:

src/Frontend/Page/CreateIdea.hs
fisx: agreed.  i suggest to think about how to do session state first, and then do a refactoring around `userLogin`.

regarding `forceLogin`, we don't really need to worry about that, though.  with the current implementation of `userLogin`, we can easily replace this line with a `redirect "/login"` as soon as that is available.

src/Frontend/Html.hs
```diff
@@ -119,7 +84,7 @@ instance ToHtml PageIdeasOverview where
         h1_ "Was soll sich verändern?"
         p_ $ "Du kannst hier jede lose Idee, die du im Kopf hast, einwerfen und kannst fuer die "
             <> "Idee abstimmen und diese somit \"auf den Tisch bringen\"."
-        div_ $ button_ "+ Neue Idee" -- FIXME: should link to idea creation form
+        div_ $ button_ [onclick_ "location.href='/ideas/create'"] "+ Neue Idee" -- FIXME: should link to idea creation form
```
andorp: The route should be calculated instead of magic `/ideas/create` constants.

src/Frontend/Core.hs
```diff
@@ -12,6 +13,61 @@ import Lucid
...
+
+-- | Render Form based Views 
+class FormPageView p where
+    -- | Generates a Html view from the given view, form action, and the @p@ page
+    formPageView :: (Monad m) => View (HtmlT m ()) -> ST -> p -> HtmlT m ()
+
+-- | The page after submitting a form should be redirected 
+class RedirectOf p where
```
fisx: just an observation: this forces us to have every page live under only one path.  this is probably going to work, as we can always add more newtypes if we need to.

fisx: sratch my last comment.  i didnt' read the comment, and thought each page had its own path stored here, which is not the case.

src/Frontend/Page/CreateIdea.hs
andorp: At the point when the user submits the form, it will have credentials. No need to redirect here.

src/Api/Persistent.hs
```diff
@@ -88,3 +100,56 @@ findUserByLogin login = find (\u -> u ^. userLogin == login) <$> getUsers
...
+forceLogin :: Integer -> Persist ()
+forceLogin x = modifyDb dbCurrentUser (const (Just (AUID x)))
+
+nextId :: Persist (AUID a)
+nextId = do
+    modifyDb dbLastId (+1)
+    AUID <$> getDb dbLastId
+
+currentUser :: Persist (AUID User)
+currentUser = fromJust <$> getDb dbCurrentUser
```
fisx: I prefer` (\(Just v) -> v)`, as it gives a better error message and a compile-time non-totality warning if -Wall is enabled.

src/Api/Persistent.hs
```diff
@@ -88,3 +100,56 @@ findUserByLogin login = find (\u -> u ^. userLogin == login) <$> getUsers
...
+instance FromPrototype Idea where
+    fromPrototype i =
+           (ideaTitle    .~ (i ^. protoIdeaTitle))
+         . (ideaDesc     .~ (i ^. protoIdeaDesc))
+         . (ideaCategory .~ (i ^. protoIdeaCategory))
+         $ emptyIdea
+
+emptyIdea :: Idea
+emptyIdea = Idea
+    { _ideaMeta = Prelude.error "blah"
```
fisx: please elaborate on "blah"  (:

fisx:
closing in favour of #38

PR #33, opened 2016-02-15
2016 02 12 np topic refinement

np:
I recommend reviewing the commits independently.


fisx:
LGTM.  missing:

- move code from Frontend.Html to Frontend.Page.TopicOverview
- some of the HTML
- adding new topics.
- tests

but i agree that all of this can be done elsewhere.

PR #36, opened 2016-02-15
2016 02 15 np topic tests

np:


Mikolaj:
Looks fine. 

Personally I find type signatures of definitions very useful. Especially in top level defs and sufficiently large or unobvious local defs. In this particular case I did wonder whether the `change` argument of `findAllInBySpec` was pure and a quick look at the type would answer that at once.

fisx:
LGTM.  i made two trivial commits, will wait for travis and merge.

PR #40, opened 2016-02-15
2016 02 15 np move topic page

np:
I will rebase it onto master.

### DIFF
--- a/src/Frontend/Topics.hs
+++ b/src/Frontend/Topics.hs
@@ -1,0 +1,137 @@
+{-# LANGUAGE DataKinds           #-}
+{-# LANGUAGE FlexibleInstances   #-}
+{-# LANGUAGE KindSignatures      #-}
+{-# LANGUAGE LambdaCase          #-}
+{-# LANGUAGE OverloadedStrings   #-}
+
+{-# OPTIONS_GHC -Werror -Wall #-}
+
+module Frontend.Topics
+where
+
+import Control.Lens
+import Control.Monad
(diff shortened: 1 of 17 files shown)


Mikolaj:
LGTM. Better to rebase soon --- such commits get into conflicts a lot.

src/Frontend/Topics.hs
```diff
@@ -0,0 +1,137 @@
+{-# LANGUAGE DataKinds           #-}
+{-# LANGUAGE FlexibleInstances   #-}
+{-# LANGUAGE KindSignatures      #-}
+{-# LANGUAGE LambdaCase          #-}
+{-# LANGUAGE OverloadedStrings   #-}
+
+{-# OPTIONS_GHC -Werror -Wall #-}
+
+module Frontend.Topics
```
andorp: What if we move it to `Frontend.Page.Topics` or should I move `Frontend.Page.CreateIdea` to `Frontend.CreateIdea`? I prefer collect every page under the Page submodule.

fisx: > What if we move it to `Frontend.Page.Topics` or should I move `Frontend.Page.CreateIdea` to `Frontend.CreateIdea`? I prefer collect every page under the Page submodule.

Yes, i would also like this to go to `Frontend.Page.Topics`.

np: I don't mind but can we do this later?

andorp: I can do this. My new change will depend on it.

Issue #20, opened 2016-02-10
Suggestion how to separate Views for Page related types.


andorp:
The `Frontend.Html` should contain some view related functionality.
````haskell
-- | Page View for Digestive Functors
class PageView a where
    pageView :: View Html -> ST -> a -> Html

instance PageView PageHomeWithLoginPrompt where
    pageView v formAction PageHomeWithLoginPrompt =
        DF.form v formAction $ do
            DF.div $
                DF.h1 "Willkommen bei Aula"
            DF.div $ do
                DF.inputText "login" v
                DF.inputText "pass" v
                DF.inputSubmit "ANMELDEN"
            DF.div $
                DF.p "Solltest du dein Password nicht mehr kennen, melde dich bitte bei den Admins eurer Schule."
````
The `Frontend` module could import and use the views.
````haskell
loginForm :: Server (FormH HTML Html ())
loginForm = formH "/login" validate process render
  where
    render :: View Html -> ST -> ExceptT ServantErr IO Html
    render v formAction = pure $ pageView v PageHomeWithLoginPrompt

    validate :: DF.Form Html (ExceptT ServantErr IO) (ST, ST)
    validate = (,)
        <$> ("login" .: DF.text (Just "Bein benutzername"))
        <*> ("pass"  .: DF.text (Just "Dein Password"))

    process :: (ST, ST) -> ExceptT ServantErr IO Html
    process _ = do
        return . toMarkup . Frame . text $ "home"
````

Please share your thoughts.

Edited by @andorp

Mikolaj:
I'm only starting to grok the magic of our tech stack, but the proposal looks very promising. Nitpicks: there is a missing ``v`` argument in ``render`` and you also failed to pass ``formAction`` through, which is currently unbound in ``pageView`` (do we need another parameter there?).

andorp:
@Mikolaj You are right, I missed to pass the argument v, I had just sketched the proposal.

andorp:
@Mikolaj You are right about the form action as well. I am not sure if the form action would come from the caller side of the `pageView`. Probably it is related to the paths we give in the servant path description.

fisx:
> Probably it is related to the paths we give in the servant path description.

you mean in the type-level routing table?  yes, it's actually the same in most cases.

[...]

PR #41, opened 2016-02-15
Test form rendering.

andorp:
* New make command: `test-repl`
 * `FormPageView` has an associated type and a new method to define forms
 * Test suite, to check if the form renders and uses the same parameters as it was given in the view

### DIFF
--- a/tests/Frontend/HtmlSpec.hs
+++ b/tests/Frontend/HtmlSpec.hs
@@ -1,22 +1,28 @@
-{-# LANGUAGE GADTs #-}
+{-# LANGUAGE GADTs             #-}
+{-# LANGUAGE OverloadedStrings #-}
 module Frontend.HtmlSpec where
 
 import Arbitrary
+import Frontend.Core
 import Frontend.Html
 import Frontend.Page.CreateIdea
 
 import Control.Applicative ((<$>))
+import Control.Monad.Trans.Except
+import Data.Either (either)
 import Data.Typeable (Typeable, typeOf)
 import Lucid (ToHtml, toHtml, renderText)
-import Test.Hspec (Spec, describe, it)
+import Test.Hspec (Spec, context, describe, it)
 import Test.QuickCheck (Arbitrary(..), Gen, Testable, forAll, property)
+import Test.QuickCheck.Monadic (assert, monadicIO, run)
 
+import Text.Digestive.View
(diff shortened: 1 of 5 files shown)


tests/Frontend/HtmlSpec.hs
```diff
@@ -59,5 +68,19 @@ data HtmlGen where
...
+    it (show $ typeOf g) . property . forAll g $ \pageSource -> LT.length (renderText (toHtml pageSource)) > 0
+
+data FormGen where
+    F :: (Show m, Typeable m, FormPageView m) => Gen m -> FormGen
+
+-- | Checks if the form rendering does not contains buttons and
+-- the view has all the fields defined for GET form creation.
+renderForm (F g) =
+    it (show $ typeOf g) . property . forAll g $ \page -> monadicIO $ do
+        length <- run . fmap failOnError . runExceptT $ do
```
np: I'd rather avoid using `length` as a variable name.

np:
apart from the comment it LGTM

tests/Frontend/HtmlSpec.hs
andorp: I am going to chage it.

tests/Frontend/HtmlSpec.hs
```diff
@@ -59,5 +68,19 @@ data HtmlGen where
...
 -- | Checks if the markup rendering does not contains bottoms.
-renderMarkup (H g) = describe (show $ typeOf g) .
-    it "renders" . property . forAll g $ \pageSource -> LT.length (renderText (toHtml pageSource)) > 0
+renderMarkup (H g) =
+    it (show $ typeOf g) . property . forAll g $ \pageSource -> LT.length (renderText (toHtml pageSource)) > 0
+
+data FormGen where
+    F :: (Show m, Typeable m, FormPageView m) => Gen m -> FormGen
+
+-- | Checks if the form rendering does not contains buttons and
```
Mikolaj: buttons or bottoms?

andorp: Good point. Bottoms. Fixing...

Mikolaj:
Does it subsume #20 (for forms?) Anyway, it's more readable than the old version and the extra test doesn't hurt either. LGTM>

PR #45, opened 2016-02-16
Add `nil` (alias of `mempty`) and `isNil`.

np:


fisx:
LGTM, as discussesd in irc.

I guess now we should also replace all occurrances of mempty and *.empty to be consistent, but I'll leave that to you @np.  (-:

PR #48, opened 2016-02-16
Use `nil` and `isNil`

np:


src/Frontend/Page/CreateIdea.hs
```diff
@@ -32,8 +32,8 @@ instance FormPageView PageCreateIdea where
 
     makeForm PageCreateIdea =
         ProtoIdea
-        <$> ("title"         .: DF.text Nothing)
-        <*> ("idea-text"     .: (Markdown <$> DF.text Nothing))
+        <$> ("title"         .: DF.text nil)
```
andorp: I would use explicit `Nothing` here as we are not build abstraction over the monoid laws when we are dealing with form definition. It introduces mental juggling over types, and searching for the definition of `DF.text` .

src/Frontend/Page/CreateIdea.hs
```diff
@@ -32,8 +32,8 @@ instance FormPageView PageCreateIdea where
...
+        <*> ("idea-text"     .: (Markdown <$> DF.text nil))
```
andorp: See my previous comment.

Mikolaj:
Uhm, nifty.

fisx:
I think I agree with @andorp, otherwise LGTM.  @np, can you make a call and merge?

src/Frontend/Page/CreateIdea.hs
np: The line is thin and thus I don't feel strongly about it. In a way `nil` is not meant to make you think about which type it is since it is the point of this abstraction. One additional advantage is when a type change `nil` will remain the empty one (lists, maps...).

I revert those.

andorp: > One additional advantage is when a type change nil will remain the empty one (lists, maps...)
I like this :) In this case there is not a big chance for a code change.

PR #50, opened 2016-02-16
Testing forms

### LINKED ISSUE #43: test forms
there is some cool stuff in #41 already, but there are form properties still to be tested.  we need polymoprhic tests like the one in #41 for:

1. post with arbitrary correct input
1. post with arbitrary incorrect input (fuzzed correct input?)
1. ...?

fisx:
Fixes #43.


fisx:
test case still fails, but that may be due to a bug in the test or to a bug in the code.  will investigate!

fisx:
Need document `selectCategoryValue` better and merge with master.  Will do that tomorrow.

src/Frontend/Page/HomePageWithLoginPrompt.hs
```diff
@@ -28,10 +28,13 @@ instance ToHtml PageHomeWithLoginPrompt where
     toHtmlRaw = toHtml
     toHtml p = semanticDiv p "PageHomeWithLoginPrompt"
 
+newtype LoginFormData = LoginFormData (ST, ST)
```
Mikolaj: Why `newtype` with a pair, and not a curried `data`?

src/Frontend/Page/HomePageWithLoginPrompt.hs
```diff
@@ -28,10 +28,13 @@ instance ToHtml PageHomeWithLoginPrompt where
...
+  deriving (Eq, Ord, Show)
+
 instance FormPageView PageHomeWithLoginPrompt where
-    type FormPageResult PageHomeWithLoginPrompt = (ST, ST)
+    type FormPageResult PageHomeWithLoginPrompt = LoginFormData
 
-    makeForm _ = do
+    makeForm _ = LoginFormData <$> do
```
Mikolaj: Is this `do` needed?

src/Frontend/Page/HomePageWithLoginPrompt.hs
fisx: > > +newtype LoginFormData = LoginFormData (ST, ST)
>
> Why `newtype` with a pair, and not a curried `data`?

you are right, will fix.

Mikolaj:
Really shrewd. Re `pendingWith`, in what sense will the input be invalid? I'm worried types get in the way and prevent us from generating something really outrageous. :) Anyway, LGTM.

fisx:
@Mikolaj: good question.  I will add a comment.

PR #55, opened 2016-02-17
Introduce Action Monad

andorp:


andorp:
> fst3 = view _1

I learn a new trick every day. Thanks :)

Mikolaj:
On one hand, it's great we switch from various ad-hoc partial monad stacks to our expanded Action monad, the same in all places. And defining the monad it absolutely necessary, it will be the crucial point of our implementation. 
On the other hand, could we try to start with constraints instead of a fixed Monad? It might be easier than refactoring afterwards, as we did with Thentos. A few primitive operations may mention the Action monad, but for others it would make sense to have some constraint shorthands and use them instead. I don't grok the details, but usually apart of the full power-constraint `ActionLike`, it makes sense to define some restricted ones, like `ActionRO`, `ActionNoError`, or, the other way around, `ActionState`, `ActionPersistent`, `ActionError`. 
It doesn't need to be very fine-grained, but it's nice to see in the types that some function can't result in an error and another can read the state (or DB), or both read and alter it. I guess we can start with the full-power constraint everywhere and relinquish the power only after the code is (locally) mostly complete and we have a better idea what code needs what power level.

fisx:
@Mikolaj: sounds like a good idea.  Don't think i'd be hard to do, right?

np:
LGTM

Mikolaj:
Perfect. We don't hardwire `Action` (apart of a few primitives), but put `ActionMonad m` constraint, which we can refine to our hearts' content when the code in question is stable or when we want to express some very strongly felt design invariant. LGTM.

Makefile
```diff
@@ -32,3 +32,6 @@ test-repl:
 hlint:
 	$(HLINT) --version
 	find src exec -name '*.hs' | xargs $(HLINT)
+
+ghci-no-type-errors:
+	cabal exec -- ghci $(FULL_SOURCES) -fdefer-type-errors
```
fisx: we can add this if it makes your life easier, but what about adding it to ~/.ghci?  (that's what i am doing.)

andorp: I prefer to have as a separate option.

andorp:
@np:
> This is one location which expects all the constraints. For some pages one might not need all the
> constraints. One option but I'm not sure if it's worthy would be to have another associated type 
> FormPageConstraints p :: (* -> *) -> Constraint.

@andorp
> I won't merge it until Monday. We can have a discussion on this idea. Actually I like it very much.

@fisx
> Do we have a reason to do this, besides that it can be done? If I'm not missing anything, the 
> simpler approach is to have too many constraints, and once this is getting in our way, it would be 
> easy to refactor something like this in.
> It's really beautiful, admittedly, but I'm worried that we are getting distracted...

andorp:
Please review and merge. I had enough rebase today. :)

andorp:
https://github.com/liqd/aula/issues/64

HLint.hs
```diff
@@ -3,6 +3,8 @@ import "hlint" HLint.Dollar
 import "hlint" HLint.Generalise
 import "hlint" HLint.HLint
 
+ignore "Avoid lambda" = Action.mkRunAction
```
fisx: > > +ignore "Avoid lambda" = Action.mkRunAction
>
> If we use `(Nat . run)` instead of `\ s -> Nat (run s)` the type checker misses the `forall a` quantification, which is needed for the Natural transformation.

I'm ok with this, but I'm wonderung: could we leave "Avoid lambda" enabled and add an explicit type signature there?

fisx:
> https://github.com/liqd/aula/issues/64

please add "Fixes #<issue number>" to the description of the PR instead.  this will add a note here that the issue has been referenced in the PR, and the issue will be closed automatically once the PR has been merged.

fisx:
(note: the "Fixes ..." needs to be in the original description of the PR, not some comment.)

fisx:
My mistake.  #55 is the PR (not an issue fixed by the PR), and #64 is a new issue created for an earlier comment from here.  Sorry, please ignore the noise!

HLint.hs
fisx: just tried this quickly, and failed.  your solution is good, too.  i was just curious.

src/Arbitrary.hs
```diff
@@ -8,14 +8,14 @@
 
 {-# OPTIONS_GHC -fno-warn-orphans -Werror #-}
 
-module Arbitrary (topLevelDomains, loremIpsum) where
+module Arbitrary (topLevelDomains, loremIpsum, generate, arbitrary) where
```
fisx: I am not that fond of re-exports.  They seem convenient early-on, but they increase overall complexity of the import graph, and make reading the code more challenging (unless we stick with explicit import lists all over).

I don't want to change anything here and now, just a general remark.

fisx:
Despite my wining and the many changes, I am quite happy with this!  (-:

Please look at my changes and either merge or assign back to me with complaints.  (To speed things up, you can also merge now and open isuses for changes you feel need to be made / un-made.)

Thanks!

PR #56, opened 2016-02-17
Add ideas to topics

np:
This simply reflects the status so far.
Still need to be done:
* formRedirectH' needs to be moved to Servant.Missing, moreover we should decide if
  it can replace formRedirectH.
* redirectFormHandler' should replace redirectFormHandler since the cost is only
  to wrap pure pages with `pure`.
* split in smaller commits
* write tests

### DIFF
--- a/src/Api/Persistent.hs
+++ b/src/Api/Persistent.hs
@@ -36,7 +36,7 @@ module Api.Persistent
     , getTopics
     , addTopic
     , modifyTopic
-    , moveIdeaToTopic
+    , moveIdeasToTopic
     , findTopic
     , findUserByLogin
     , findIdeasByTopicId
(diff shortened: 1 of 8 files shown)


fisx:
The `Servant.Missing` was what I was hoping would work.  This looks to me like it's ready to move there and be used as is, and polished later.

The page looks exciting, but I have only looked at the code.

This should probably go to some US page, but let me just write it down here: On saving a newly created topic, we will probably have to show a page saying "your topic has been created", which redirects on button click or with a delay of 3 seconds.  Or something like that.  I don't see anything there that we didn't already plan on doing (the re-direct can be done in the HTML header or with a line of javascript, or just on button click), just letting everybody know.

andorp:
LGTM

PR #58, opened 2016-02-18
Add hlint to travis build.

andorp:
HLint checks are added to travis build. It won't break the build for the time being, until we figure out the common rules. Please contribute to `src/HLint.hs` or comment on this thread.


Mikolaj:
LGTM.

PR #59, opened 2016-02-18
Add '-Wall -Werror' to travis build.

### LINKED ISSUE #30: travis should build with -Wall -Werror and run hlint.
see https://github.com/liqd/thentos on how we could do it.

andorp:
Fixes #30


fisx:
hlint gives lots of warnings, but travis succeeds anyway.  (not really related to this PR, but i suggest to fix it here anyway.  close enough.)

andorp:
I propose to have hlint at an evaluation period and everyone can add there objections to hlint ruleset and next week we are going to make the build fail if hlint fails. You just have to move from `before_install` section to `script` section, and remove the `exit 0` from `./travis/docker-hlint.sh`

andorp:
Anyway thanks for taking care of hlint complaints. +1

fisx:
Ok, I if travis doesn't surprise me now I'll be happy with this PR.

I also think everybody should take a look at the rules I disabled in `HLint.hs` and make sure they can live with this.  Some rules I feel strongly about, some I don't really care.  Please comment!

I'm also ok with merging this as we can always changes the rules again later.

I'll un-assign myself.  @andorp, your move.  (Also, thanks! :-)

andorp:
Please comment +1 if you are ok with the current HLint rules.

np:
* I'm keen on the use of `<>` instead of `++`.
* I can live with redundant `do`, `$`, no `const` but I would turn them on.
* While I prefer `map` over `fmap` (shorter), if infix `<$>` it's even better.

Mikolaj:
Personally, to save time, I tend to run hlint only before (frequent) releases. But I'm OK with the rules and since travis checks hlint in PRs for me, I won't at least have to run it by hand (though in emacs it nicely jumps to the hlint error, so I may end up doing so for larger PRs). +1

andorp:
>  I'm keen on the use of <> instead of ++.
>  I can live with redundant do, $, no const but I would turn them on.
>  While I prefer map over fmap (shorter), if infix <$> it's even better.
I don't have any preferences on them. @fisx please discuss it with @np

fisx:
@np: I don't know of a rule enforcing `<$>` over `map`, but I think I wouldn't like that anyway.  So if I understand you correctly, you want to enforce `<>` over `++`.  I changed that.

I don't know how to interpret your other points (does "turn off" mean to ignore, or to turn off the ignore?).  I am going to merge this.  If you still feel the need to change the rules, please open a new issue / PR, and we can discuss it.

fisx:
> Personally, to save time, I tend to run hlint only before (frequent) releases. But I'm OK with the rules and since travis checks hlint in PRs for me, I won't at least have to run it by hand (though in emacs it nicely jumps to the hlint error, so I may end up doing so for larger PRs). +1

yes, this is also my work-flow, but sometimes i also like to have travis do it for me.

PR #42, opened 2016-02-15
Use sensei to drive exec/RenderHtml.

fisx:
I just checked: on my laptop, I can set up sensei to watch the repo and on changes in src/Frontend/Page/*.hs, refresh of all html will take <3s.

The only problem is that sensei can't un-watch files.  See https://github.com/hspec/sensei/pull/28 for a discussion what needs to be done there.

PR #61, opened 2016-02-19
Get rid of unsafePerformIO in RenderHtml

np:
Only the last commit is part of this PR.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,9 +8,9 @@ before_install:
     - git show | head -1  # (for matching against commit hash given on the travis log web page)
     - docker pull fisx/aula
     - docker run --rm -it -v `pwd`:/root/aula/ fisx/aula /bin/sh -c "/root/aula/travis/docker-list-sandbox-packages.sh"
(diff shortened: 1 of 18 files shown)


Mikolaj:
I can't say I understand everything, but the result has definitely an improved readability (and is safer).

BTW, I know it's hard to foresee in advance, but may I ask for a separate commit with style improvements next time? Thank you!

andorp:
LGTM

fisx:
Yes, this is an improvement!

I will add this to #42, where I have touched RenderHtml quite a bit.

fisx:
Manually moved to #42.  Please check and re-open if I missed anything.

fisx:
> BTW, I know it's hard to foresee in advance, but may I ask for a separate commit with style improvements next time? Thank you!

+1

my habit is to work a bit on things, and then commit them in sematic
packages, but not necessarily self-contained.  by allowing for the
latter, it's not that much more work, but the down side is that `git
bisect` won't work any more.  the biggest upside for me is that some
commits can clearly be checked by travis and require almost no human
review at all.

i am in favor of giving up on `git bisect` and aiming for clean,
readable commits.

PR #62, opened 2016-02-20
Add User as a Frame parameter.

andorp:
Preparation, before merging Action monad. The username will come from the session.

### DIFF
--- a/exec/RenderHtml.hs
+++ b/exec/RenderHtml.hs
@@ -1,4 +1,5 @@
 {-# LANGUAGE BangPatterns          #-}
+{-# LANGUAGE OverloadedStrings     #-}
 {-# LANGUAGE ScopedTypeVariables   #-}
 
 {-# OPTIONS_GHC -Werror -Wall #-}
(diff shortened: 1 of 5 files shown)


andorp:
@fisx please review

fisx:
Should we just add the entire user, while we are at it?  (The avatar will also be needed, and probably the user id for the link to the profile page.)

(This also conflicts with #42, but I think it will be easy enough to resolve.)

fisx:
(@andorp: i'll hang out on irc for a bit.)

fisx:
I would have preferred two commits here, since I was already finished reading the previous changes.  Also: sorry for making you write `frameUserHack`!  :-)

Otherwise LGTM!  merging~

andorp:
@fisx: let's discuss offline the two commits suggestion. I would like to learn :)

andorp:
> Also: sorry for making you write frameUserHack! :-)

No problem, I just wanted to make it visible for everyone that we should remove it as soon as possible. :)

PR #68, opened 2016-02-21
Add specs module that imports src/**/*.hs

fisx:
(for exhaustive type checking in sensei.)

### DIFF
--- a/aula.cabal
+++ b/aula.cabal
@@ -28,6 +28,7 @@ library
     , Frontend.Page.Static
     , Frontend.Page.Topic
     , Frontend.Page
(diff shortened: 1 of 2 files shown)


fisx:
(The hlint error is fixed in #55.)

andorp:
In selenium?

fisx:
re. travis: I think what happens here is that the cabal file is missing the entry for `Frontend.Prelude`.  I could fix this, but it would only cause conflicts with #69, where the problem doesn't exist any more.  I will wait for #69 to be merged, then rebase this, then it should be ok.

PR #75, opened 2016-02-22
Api types

fisx:
- Give static content its own sub-directory.
- Refactor: move some primitive api types to more suitable modules.
- move existing end-points to /testing.
- add new api with all pages from click dummy.
- add handlers throwing "not implemented".

From here on, we can either work on the quick&dirty routing table as before, move features from there to the final application, or work on the latter from start.

I like servant!


Mikolaj:
After a quick look: you've got to love Servant. :) LGTM.

andorp:
LGTM

PR #81, opened 2016-02-22
Refine action abstraction

### LINKED ISSUE #67: Use `MonadIO` instead of `ActionIO`?
https://github.com/liqd/aula/commit/a4d695a7799d9e1b3193e20a881b73f373453116

I introduced that in #55, but @andorp has objections and has reverted it.

What are the pros and cons for each?  I like `MonadIO` because it is isomorphic and because I can't think of any reason to duplicate its structure.  But I'm sure there is at least one.

fisx:
Fixes #67.

(Some of this stuff I am inclined to call bug fixes.)

### DIFF
--- a/src/Action.hs
+++ b/src/Action.hs
@@ -44,8 +44,6 @@ class ( ActionLog m
       , ActionPersist m
       , ActionUserHandler m
       , ActionError m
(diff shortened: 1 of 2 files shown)


np:
LGTM

andorp:
LGTM

Mikolaj:
Perfect.

PR #82, opened 2016-02-22
Remove the last unsafePerformIO

np:


andorp:
On https://github.com/liqd/aula/pull/82/files#diff-3c928bdddfffa6d17f5d2f2e68f32b81R163 :

In the future when Form Generation depends on some persistent/action state we need to make this smarter to be able to add context for the tests.

andorp:
LGTM

PR #84, opened 2016-02-22
Initial DB using an arbitrary item of each type

np:


andorp:
What if we move `createRandom` into Action module? Is it ok if I do it by myself?

fisx:
> What if we move `createRandom` into Action module? Is it ok if I do it by myself?

note that this code (not just the module) will move to ./tests/ at some point, but if you have a reason to move it now go ahead.

andorp:
Please review my changes.

fisx:
Ah, you want to eliminate the `CreateRandom` module so we don't have to export `persistIO`.  That's indeed a good reason, but I still have no strong opinion either way.

fisx:
Ah, even better: you just moved a small part to `Persistent`, not the entire `CreateRandom` module.

I like this a lot now, both your changes.  I added to small commits.  @np, can you take a last look and merge if happy?

fisx:
(Argh, forgot to self-assign/un-assign.  How is this so hard?)

andorp:
Please review.

fisx:
LGTM.

Minor remark: I would have done this change as a separate commit for two reasons.

- it keeps track of what we learned, and how much things imrproved in the process.
- it's easy to review.

PR #85, opened 2016-02-22
Keep paths in sync between routing tables, handlers, pages.

### LINKED ISSUE #77: Hard coded paths in HTML templates.
There are similar code snippets in the code base:

````haskell
div_ $ button_ [onclick_ "location.href='/ideas/create'"] "+ Neue Idee"
````

We don't have QA personel to check the page accessibility. Links to pages should not be hard coded in this form. I see two possible solutions:

 * Have one ADT which enumerates all the pages in parallel with the Page types. Which is bad as it leads us to double bookkeeping.
 * Have a type family with type level Strings and use those strings as paths on the value level.

### PR DESCRIPTION
**Pull request by fisx**
Fixes #77.

### CODE REVIEW

**Comment by fisx** [comment]
The last commit should demonstrate that the `ADT` approach is at best unpleasant.  It will require a lot more work in getting all paths passed through our form handlers.

Giving up for today.

**Comment by andorp** [comment]
I have another solution for this. Let's discuss it when you are awake.

**Comment by fisx** [comment]
This solution suffers the same problem as `RedirectOf`, namely that it moves the information where a link points from the routing table (where it is created) to the link target.

But this may be our best option.  If we allow for orphans, we can still place the `Page` instances next to the routing table.  Perhaps it's not so bad.

@np, @Mikolaj, @kosmikus: thoughts?

nit-picking:
- the syntax is `Proxy :: Proxy PageCreateIdea`
- the name should be `HasPath`, or `HasUriPath`.  I'm leaning towards the former.
- the name for `RedirectOf` should be `HasRedirectTarget`.

**Comment by fisx** [comment]
still got it wrong: the `pth` primitive doesn't compose the path up to `Top`, but only gives you the sub-path.  not very useful.  i'll fix that!

**Comment by fisx** [comment]
i would like to use `P.pth P.MainSpace` instead of `P.pth $ P.TopMain P.MainSpace`, which works because `Main` has no prefix.  However, this will make it harder to move `Main`.

Thinking about this, I realized that having `HasPath` instances for sub-api path types opens up Page developers to the risk of generating suffixes instead of full paths.  Solution: instance of `HasPath` instance, create `foo` function for every `Foo` path type as I did earlier in 4ffd7fb, and only write `HasPath` instances for `Main`, `Top`.  Then we can take away the `Main` instance if we want to move `Main` to a non-empty prefix.

**Comment by fisx** [comment]
I tried this:

```haskell
(</>) :: HasPath p =>  -> UriPath -> UriPath
(</>) a b = pth a <//> b
```

but it messes things up.  with overloaded strings in place, it would force us to annotate all string literals with a type.  this problem is not local to `Frontend.Path`, either, and will clutter the caller's code, too.  it would also force us to use the type synonym instance extension.

will abandon this idea for now.

fisx:
(corrected typo in last comment.)

np:
Nothing blocking on my side, in particular since we might want this to be merged soon to make progress.

Mikolaj:
Clear, explicit. LGTM.

src/Frontend/Path.hs
```diff
@@ -0,0 +1,122 @@
...
+
+(</>) :: UriPath -> UriPath -> UriPath
+(</>) = (<//>)
+
+href_ :: HasPath p => p -> Lucid.Attribute
+href_ = Lucid.href_ . pth
+
+
+class HasPath p where
+    pth :: p -> UriPath
```
andorp: Why not `path`?

np:
LGTM

PR #90, opened 2016-02-23
Handle missing trailing slashes in addresses

### LINKED ISSUE #70: Handle missing trailing slashes better.
`serveDirectory` expects its path to have a trailing slash, or the relative references it lists won't work.

```
16:33 < fisx_> so, if you are on page /samples, then your working directory is /
16:33 < fisx_> then relative link x.html points to /x.html
16:33 < fisx_> if you are on page /samples/
16:34 < fisx_> then your working directory is /samples
16:34 < fisx_> then it works.
16:34 < fisx_> it's not nice, but it's not clear to me how to handle this better either.
16:35 < fisx_> perhaps if we offer "samples" :> "index.html", but not "samples"?
16:35 < fisx_> no redirect or anything?
16:35 < fisx_> (the redirect hack would make sure that nobody ever clicks a link from /samples,
16:35 < fisx_> while the url would still get everybody to /samples/, where they can be happy.
16:36 < fisx_> i've actually seen this on well-made pages before.
16:36 < fisx_> if you don't add a trailing slash, it is added for you.
16:36 < fisx_> prehaps there is even a middleware for that?
16:36 < fisx_> i'll open an issue.
```

I like the idea of adding a middleware, and letting servant work on the assumption that all urls have a trailing slash in the path.  Not sure though if I'm missing something here.

Mikolaj:
Fixes #70.

### DIFF
--- a/aula.cabal
+++ b/aula.cabal
@@ -59,6 +59,7 @@ library
     , time
     , transformers
     , wai
+    , wai-app-static
     , warp
   exposed-modules:
       Action
@@ -270,6 +271,7 @@ test-suite spec
     , time
     , transformers
     , wai
+    , wai-app-static
     , warp
     , aula
     , base >=4.8 && <4.9
--- a/src/Frontend.hs
+++ b/src/Frontend.hs
@@ -12,10 +12,13 @@ where
 
 import Control.Monad.Trans.Except
 import Lucid
+import Network.Wai (Application)
 import Network.Wai.Handler.Warp (runSettings, setHost, setPort, defaultSettings)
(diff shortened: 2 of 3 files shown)


fisx:
Thanks, this is much nicer than my hack.

I would have put `waiServeDirectory` into `Forntend.Core` so we can use it in other places as well, but we can do that once those other places pop into existence.

I changed whitespacing a bit and will merge when travis is green.

Mikolaj:
Thank you.

andorp:
LGTM, if the travis is green

PR #87, opened 2016-02-23
README.md Fix the instructions for dummies, etc.

Mikolaj:
A fix to docs. Pushing, in particular, to show how I'm invoking dummies and ask on IRC for help in debugging them.

PR #91, opened 2016-02-23
Fix: sensei make rules; also add a little documentation.

### LINKED ISSUE #92: Fix README
- Do not mention THENTOS_ROOT_PATH, or mention setting it to thentos-core to match new Makefile contents.
- fix rendering of shell commands.

fisx:
Fixes #92.


Mikolaj:
Should we also change the README.md? Now it asks to set ` THENTOS_ROOT_PATH=`pwd`/thentos`, while you seem to use ` THENTOS_ROOT_PATH=`pwd`/thentos/thentos-core`.

fisx:
Good point.  `$THENTOS_` is only needed with `sensei-full`, though.  just pushed that out of #87.

fisx:
@Mikolaj: I understood your comment now.  You are right, will fix!

PR #95, opened 2016-02-24
2016 02 24 mk misc minor fixes

Mikolaj:
These are misc fixes from testing page creator workflow with docker. Just a few hunks and I think nothing dangerous, but a quick review would be welcome.

Please merge before Page Creator commences. I will be available to provide additional guidance, on top of what I added to README (and elswhere to HOWTO, already pushed).


np:
LGTM

fisx:
Thanks!  I added support for the AULA_SAMPLES volume, elaborated a little on the points you identified, and will merge now.  (Please yell if I un-improved anything.)

PR #100, opened 2016-02-24
Re-align hpack's package.yaml file with aula.cabal.

fisx:


Mikolaj:
How does it differ from the previous one?

PR #101, opened 2016-02-24
Move pages to Page.Comment and Page.Idea

np:


fisx:
@np hlint wants you to use `isNothing` over `== Nothing`.  I guess the motivation is it doesn't require an `Eq` instance, but I don't see a problem with that, and `== Nothing` requires less context, makes things more explicit.  You can either change the code or disable the rule.

fisx:
(or assign to me and i'll do it.)

np:
The argument of context would go away with a prelude module.

Mikolaj:
LGTM.

fisx:
LGTM, too!

PR #102, opened 2016-02-24
Tweak html hacking tooling and docs

### LINKED ISSUE #94: [Feature request] Under docker 'make click-dummies-refresh' needs to be prodded after .hs change in aula-samples/
[Edit: actualy, README doesn't cover the .hs files inside aula-samples/. They can be sometimes changed manually to make the random data look nicer, but it's a minor issue anyway.]

02:17 < mikolaj> fisx: I need to prod 'make click-dummies-refresh' in a docker with Enter (and then it works fine); I assume this is the intended/acceptable behaviour, so I don't file a github issue
02:18 < mikolaj> just letting you know, so that Caroline knows to prod it
02:19 < mikolaj> however, I don't have to prod it whenever any file under aula/ changes; only when I tweak the .hs files under aula-samples/
02:20 < mikolaj> (with 'nano' editor copied to the docker, because I can't use vi)
02:27 < mikolaj> hmm, actually README does suggest you never need to touch the `make click-dummies-refresh` terminal
02:27 < mikolaj> too bad; will file an issue

This may or may not be related to 'make click-dummies-refresh' rebuilding stuff all the time whenever  aula-samples/ is inside ~/aula/. Tweaking timezone doesn't help. I'm just putting aula-samples/ outside ~/aula/ and then editing its contents inside the docker with a copied-over `nano` (but editing our normal source code in ~/aula/ with my trusty emacs; it works fine).

fisx:
Fixes #94


fisx:
Ok, to merge, coverage is not our main concern as of now.

(I also modified the coverall settings, hope we won't be warned again.)

andorp:
LGTM

Mikolaj:
Splendid!

PR #104, opened 2016-02-24
Fix: semanticDiv does not use ToHtml for p.  Drop bogus ToHtml instances for forms.

fisx:


fisx:
argh, not as easy as i thought.  but doable.  will take another look later.

Mikolaj:
LGTM

PR #107, opened 2016-02-24
User Settings Page

andorp:


src/Action.hs
```diff
@@ -76,6 +82,8 @@ instance ActionUserHandler Action where
         put $ UserLoggedIn username "session"
         persistent $ loginUser username
 
+    userState = get
+
     logout = do
         gets _username >>= persistent . logoutUser
```
np: This is not related to your commit but `gets _username` can be replaced by `use username`.

src/Frontend.hs
```diff
@@ -208,6 +211,8 @@ type AulaTesting =
   :<|> "topics" :> GetH (Frame (PageShow [Topic]))
   :<|> "users"  :> GetH (Frame (PageShow [User]))
 
+  :<|> "user" :> "settings" :> FormH HTML (Html ()) ST
+
```
np: Given the new state of AulaTesting, I would avoid adding handlers we have elsewhere.

src/Frontend/Page/User.hs
```diff
@@ -0,0 +1,67 @@
...
+    , profileNewPass1 :: Maybe ST
+    , profileNewPass2 :: Maybe ST
+    }
+    deriving (Eq, Show)
+
+instance FormPageView PageUserSettings where
+    type FormPageResult PageUserSettings = UserSettingData
+
+    formAction _ =
+        P.path . P.TopTesting $ P.path P.UserSettings
```
np: ... and thus adjusting this to the main frontend

src/Frontend/Page/User.hs
```diff
@@ -0,0 +1,67 @@
...
+            DF.form v fa $ do
+                DF.inputText "email" v >> br_ []
+                DF.inputText "old-password" v >> br_ []
+                DF.inputText "new-password1" v >> br_ []
+                DF.inputText "new-password2" v >> br_ []
+                DF.inputSubmit "ANDERUNGEN SPEICHERN"
+
+-- FIXME: Redirect to the right place
+instance RedirectOf PageUserSettings where
+    redirectOf _ = P.path $ P.TopTesting "/ideas"
```
np: Likewise

src/Frontend/Path.hs
```diff
@@ -70,6 +71,7 @@ main SpaceAll               = "space/"
 main (SpaceOne sid p)       = "space/" </> sid </> space p
 main UserAll                = "user/"
 main (UserOne (AUID uid) p) = "user/" </> cs (show uid) </> user p
+main UserSettings           = "user/settins/"
```
np: "settins", typo?

src/Action.hs
andorp: I got the following error:
````
src/Action.hs:88:13:
    Not in scope: ‘username’
    Perhaps you meant ‘_username’ (line 125)
````

Mikolaj: The lenses are defined in the same file, much below.

andorp: It just behaves strange if I put it just before the class instance.

andorp:
I merged it accidentally. :( If there are issues, please feel free to revert.

PR #108, opened 2016-02-24
2016 02 cc html hack frame

local-girl:
(everything except the last space from @local-girl.  i just rebased.)

### DIFF
--- a/src/Frontend/Core.hs
+++ b/src/Frontend/Core.hs
@@ -18,7 +18,6 @@ import Data.Set (Set)
 import Data.String.Conversions
 import Data.Typeable
 import Lucid
-import Lucid.Html5
 import Lucid.Base
 import Servant
 import Servant.HTML.Lucid (HTML)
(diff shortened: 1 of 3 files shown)


src/Frontend/Core.hs
```diff
@@ -18,6 +18,7 @@ import Data.Set (Set)
 import Data.String.Conversions
 import Data.Typeable
 import Lucid
+import Lucid.Html5
```
fisx: This line can go away; it's implied by one of the other imports.

exec/RenderHtml.hs
```diff
@@ -182,4 +182,4 @@ dynamicRender s = do
         no = return Nothing
 
         pf :: User -> a -> Html ()
-        pf user = pageFrame' [meta_ [httpEquiv_ "refresh", content_ "1"]] (Just user) . toHtml
+        pf user = pageFrame' [] (Just user) . toHtml
```
fisx: I've done this (with comment, and calling a slightly simpler function) in #104.

PR #113, opened 2016-02-25
Add readMaybe to our "Prelude"

np:
This is actually a re-export from Text.Digestive.Util.

### DIFF
--- a/src/Types.hs
+++ b/src/Types.hs
@@ -10,6 +10,8 @@
 {-# OPTIONS_GHC -Wall -Werror #-}
 
 module Types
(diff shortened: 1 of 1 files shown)


andorp:
LGTM

kosmikus:
What is wrong with `Text.Read.readMaybe` which is in `base`?

andorp:
Sorry I was to quick merging this.

andorp:
@np wdyt?

np:
Good catch @kosmikus my mental base was too old indeed.

PR #114, opened 2016-02-25
Fix readability of #104

fisx:
better source comments based on comments to #104 by @np.  thanks!

### DIFF
--- a/exec/RenderHtml.hs
+++ b/exec/RenderHtml.hs
@@ -96,7 +96,7 @@ instance (ToHtml p) => ToHtml' (ToHtmlDefault p) where
 instance (FormPageView p) => ToHtml' (ToHtmlForm p) where
     toHtml' (ToHtmlForm p) = unwrap2 $ do
         v <- unwrap1 $ getForm "" (makeForm p)
-        formPage v "formAction" p
+        formPage v "/pseudo/form/action" p  -- (action doesn't matter here)
       where
         unwrap1 = return . runIdentity
         unwrap2 = HtmlT . return . runIdentity . runHtmlT
--- a/exec/RenderHtml.hs
+++ b/exec/RenderHtml.hs
@@ -75,23 +75,23 @@ pages f =
     ]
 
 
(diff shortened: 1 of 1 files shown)


np:
LGTM

exec/RenderHtml.hs
```diff
@@ -75,32 +75,38 @@ pages f =
...
 
 
+-- | We write 'ToHtml' for pages that contain no forms, and 'FormPageView' for pages that do.  In
+-- this module, we need to render both into html for viewing only, and 'ToHtml'' is introduced for
+-- this.  We can instantiate wrapper types for 'ToHtml' instances and 'FormPageView', resp., and get
+-- a uniform way of rendering html for either.
 class ToHtml' p where
     toHtml' :: Monad m => p -> HtmlT m ()
 
+-- | A wrapper type to make all 'ToHtml' instances 'ToHtml'' instances.
```
Mikolaj: a typo

exec/RenderHtml.hs
```diff
@@ -75,32 +75,38 @@ pages f =
     ]
 
 
+-- | We write 'ToHtml' for pages that contain no forms, and 'FormPageView' for pages that do.  In
+-- this module, we need to render both into html for viewing only, and 'ToHtml'' is introduced for
```
Mikolaj: Haddock accepts the @ToHtml'@ form as well and it may look better here. Haddock doesn't generate links in that case, which is actually good, because we refer to the definition immediately below. Also one link per entity in one haddocks is usually enough, so the @foo@ form could be used below as well.

Mikolaj:
Other than the notes, yep, good docs and clearer code.

exec/RenderHtml.hs
Mikolaj: Here @ToHtml'@ would really help --- I thought these are the same. BTW, if anybody has a good idea, a name without a prime would be better (though I see how @ToHtml'@ makes sense).

fisx:
thanks @Mikolaj, good points.  but since i personally like the primed name better, and it's just local, and this is already the second round, this time only to optimizise code readability, and this code is not relevant for production, i would like to keep it.

merge?

PR #118, opened 2016-02-25
2016 02 cc add sass and fonts+2

local-girl:

PR #119, opened 2016-02-25
2016 02 cc add sass and fonts+3

local-girl:
rebase of #118


fisx:
not that i could tell, but LGTM.  (-:

waiting for travis.

PR #123, opened 2016-02-25
2016 02 cc style frame

local-girl:
Some basic styling, still looks crap - sorry !


np:
LGTM

local-girl:
<i> is very common these days for icons

Issue #31, opened 2016-02-15
use uri-bytestring for urls instead of ST


fisx:
Also relevant: http://hackage.haskell.org/package/url-generic-0.1/docs/Web-URL-Generic.html (similar to the idea of `Frontend.Path`; unmaintained as of 2016-02-25)

fisx:
also relevant

fisx:
edited description.

fisx:
I am currently leaning towards writing more documentation for `Data.UriPath` and `Frontend.Path`, and not draw in any more libraries, but only comment on what we are doing differently.

If nobody objects, I will make a PR to that end soonish.

fisx:
More comments from brief off-line discussion with @np:

- It may still be nice to replace `ST` with `uri-bytestring` as a target type in `Data.UriPath`.
- It may be overkill, too, as we are only dealing with paths.  But `uri-bytestring` supports paths, relative refs, and we may still need more of the full URI structure in the future.

fisx:
@np: I just tried to change the avatar field in `User` from `ST` to `Maybe UriPath`, but that would have destroyed `Eq` and `Show` derivability.

What do you think about changing the implementation of `UriPath` to something slower, but more manageable and straight-forward?  (Alternatively, we could just use uri-bytestring here.)

fisx:
[irc]

```
10:52 < npou> one easy option is to say that the Eq instance will be slow, (==) = (==) `on` 
              (`appendUriParts` [])
```

fisx:
I am about to commit this in `Frontend.Path`:

```haskell
main (Login mb)       root = root </> (fromString ("login" <> if fromMaybe True mb then "" else "?status=False"))
```

it works, but it's certainly not as nice as uri-bytestring.  so when deciding where to go with `Data.UriPath`, we should thing about queries.

fisx:
Last comment has been resolved in #406.  Closing for now, a clear need for uri-bytestring does not appear to be emerging any time soon.

PR #125, opened 2016-02-25
All the LINKS are belong to us

np:
This PR depend on #120 so for now it compares to 2016-02-25-np-refactor-ideas-topics.


src/Data/UriPath.hs
```diff
@@ -0,0 +1,72 @@
...
+
+import Thentos.Prelude
+import Data.String.Conversions
+
+import qualified Data.Text as ST
+import qualified Lucid
+
+newtype UriPart = SlashFreeUriPart { fromUriPart :: ST }
+
+instance IsString UriPart where
```
andorp: Why do we need `IsString`?

src/Data/UriPath.hs
```diff
@@ -0,0 +1,72 @@
...
+    fromString s
+        | '/' `elem` s = error $ "UriPart.fromString: unexpected literal '/' in URI part: " <> s
```
andorp: I have concerns about the `error`

andorp:
Two minor comments. Otherwise LGTM

fisx:
`IsString` is for `UriPath` literals; the error is to force use of `(</>)` for path segment separation.

fisx:
@np, did you take a look at `uri-bytestring` (#31)?  the first commit looks like you are re-implementing this?

src/Data/UriPath.hs
```diff
@@ -0,0 +1,72 @@
...
+    mempty = DiffUriParts id
+
+    DiffUriParts ps `mappend` DiffUriParts qs = DiffUriParts (ps . qs)
+
+infixl 7 </>
+
+(</>) :: UriPath -> UriPart -> UriPath
+DiffUriParts ps </> p = DiffUriParts (ps . (p :))
+
+instance IsString UriPath where
```
fisx: This makes it legal to write:

```haskell
p = "no/slashes/" </> "ohno"
```

I don't want to change this, just an observation.

fisx:
More thoughts, good and bad mixed:

- I added a few trivial changes.
- Now that I've seen the entire branch, I have to admit it looks very differnt from `uri-bytestring`.
- I noticed that `UriPath` concatenation allows anybody now to construct invalid paths from valid ones.
- Overall I'm worried about over-engineering.  It's not easy to see your way through this api.
- Tests would be great (especially in light of the above).

In conclusion, LGTM!  I enjoyed the mind-boggling, and I'm looking forward to seeing how this will help us and/or bite us in the future.  :-)

Merging.  Thanks!

src/Data/UriPath.hs
np: Better would be to statically check the string literals.

Another option would be to use URI encoding.

andorp: For the time being, we can have it this way. If we encounter issues coming from here, we will change it.

np:
@fisx I didn't look thoroughly into uri-bytestring. The module was short and easy to setup, my goal was to avoid using an alias to ST.

What I don't like:

* UriPath.fromString: this was to support the long static paths for TopStatic.
* Absolute vs relative

Improvements:

* UriPath.fromString: we could at least do some cleaning of leading and trailing slashes.
* Absolute vs relative: it might be worth trying to index UriPath by a type-level Boolean.
  The Monoid instance of UriPath could be restricted to relative.
* The best solution should make use of the routing table.
* I'm not sure the `Space` layer is really helping, for instance the page types are all considered top-level.
  For instance `U.ViewIdea space ideaId` instead of `U.Space space (U.ViewIdea ideaId)`.

Relative paths are supposed to help with relocation, but imagine that for some reason we want to invert the nesting of some paths/captures then our data types reflecting the old scheme would look inconsistent.

My conclusion is that safe links really have a benefit for us and that relative paths add unneeded complexity.

PR #126, opened 2016-02-25
Show Current User in the Frame

andorp:


src/Frontend/Core.hs
```diff
@@ -76,16 +78,20 @@ class RedirectOf p where
...
+  | otherwise       = return $ PublicFrame p
 
 instance (ToHtml body) => ToHtml (Frame body) where
-    toHtmlRaw                = toHtml
-    toHtml (Frame usr bdy)   = pageFrame (Just usr) (toHtml bdy)
-    toHtml (PublicFrame bdy) = pageFrame Nothing (toHtml bdy)
+    toHtmlRaw = toHtml
+    toHtml = renderPageFrame toHtml
+
+renderPageFrame :: Monad m => (t -> HtmlT m a) -> Frame t -> HtmlT m ()
```
andorp: It was necessary to create this functions, becuase I wanted to use ToHtml in the redirectFormHandler but got:

````
Could not deduce (ToHtml (HtmlT Identity ()))
````

**Comment by np** [comment]
I'm wondering if we could move the Frame wrapping out of the handlers?

For instance:

```
viewIdea :: ActionPersist m => IdeaSpace -> AUID Idea -> m (Frame ViewIdea)
```

Would become:

```
viewIdea :: ActionPersist m => IdeaSpace -> AUID Idea -> m ViewIdea
```

and then only in Frontend.hs we deal with the Frame.

P.S. I've written such a comment earlier on but I can't find it anymore so...

fisx:
> P.S. I've written such a comment earlier on but I can't find it anymore so...

I think this is a good idea, found your previous comment, and linked it here: #129.

src/Frontend/Core.hs
fisx: I found a solution for that involving an orphan.  Will push when I'm done; please review and revert if you have any objections.  (Your solution was perfectly fine, I just wanted to understand how things work here.)

fisx:
LGTM, thanks!

I made a few changes; please review, revert anything you don't like, and merge.

Issue #129, opened 2016-02-26
Move frame out of the handlers


fisx:
- https://github.com/liqd/aula/commit/4ea328c18312062f7d42afa5184abbd7cf1c0e78#commitcomment-16353983
- https://github.com/liqd/aula/pull/126#issuecomment-189389039

andorp:
#132

fisx:
first attempt: https://github.com/liqd/aula/commit/a95525b1f69ce3dd8d723e73e9d36b832b40ece2

Since it doesn't reduce the number of calls to `makeFrame`, I reverted that.  Other options:

1. Use custom end point `GetH` and call `makeFrame` it in the `HasServer` instance.
2. For forms, call it in our generic form handler.
3. do it in `enter`?
4. new type class `HasFrame` with instances for all servant combinators that pushes calls to `makeFrame` from the top of the routing table down to the endpoints.  (like `enter`, but go more of the way on foot, i think.)

I like 3 most, even though I'm not sure if it can be done; followed by a combination of 1, 2.

PR #130, opened 2016-02-26
Frame refactor attempt - DO NOT MERGE

andorp:
**I create this to summarize my learnings**

Using `HTML` from the `Servant.Lucid` module forces us to use Identity monad in HtmlT but ToHtml instances forces us to use `HtmlT m ()` when we tackle with the Frame wrapping.

### DIFF
--- a/src/Frontend.hs
+++ b/src/Frontend.hs
@@ -11,13 +11,10 @@ module Frontend
 where
 
 import Control.Monad.Trans.Except
-import Lucid
 import Network.Wai (Application)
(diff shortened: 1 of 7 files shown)


andorp:
https://github.com/liqd/aula/issues/129

andorp:
When I used `type FormHandler a = FormH HTML FormHtml a` and applied the `FrameHtml` constructor to the `formPage v fa page` GHC started to behave wierdly.

fisx:
I don't think I'll look into this today.  Is it urgent?  If so, please re-assign to me.

andorp:
Please do not look into this :) I think I have another approach.

PR #132, opened 2016-02-27
Type Safe Forms, makeFrames in Frontend module

andorp:
As a partial result for the Frame Refactor, Type Safe Forms are created. Some of the `makeFrame` functions called in the `Frontend` module.


andorp:
````haskell
type AulaSpace =
       -- browse wild ideas in an idea space
       "idea" :> GetH (Frame PageIdeasOverview)
       -- view idea details (applies to both wild ideas and ideas in topics)
  :<|> "idea" :> Capture "idea" (AUID Idea) :> "view" :> GetH (Frame ViewIdea)

aulaSpace :: IdeaSpace -> ServerT AulaSpace Action
aulaSpace space = makeFrame =<< (
       Page.viewIdeas  space
  :<|> Page.viewIdea   space)
````
````
src/Frontend.hs:162:19:
    Couldn't match type ‘AUID Idea -> Action (Frame ViewIdea)’
                   with ‘Frame (AUID Idea -> m0 ViewIdea)’
    Expected type: ServerT AulaSpace Action
      Actual type: Action PageIdeasOverview
                   :<|> Frame (AUID Idea -> m0 ViewIdea)
    In the expression:
      makeFrame =<< (viewIdeas space :<|> viewIdea space)
    In an equation for ‘aulaSpace’:
        aulaSpace space
          = makeFrame =<< (viewIdeas space :<|> viewIdea space)
````

fisx:
the `:<|>` operator doesn't give you an end point, but a routing table consisting of more than one end point.  you can't make a frame from a routing table, you can just serve it, or create swagger docs from it.  not sure if that's a helpful explanation.  reading in servant sources helps, though!  (-:

andorp:
I have a partial result. I would stop here with this.

andorp:
rebasing...

andorp:
@fisx If you don't like the last commit a95525b please, revert  it. I doubt about that change myself as well.

Mikolaj:
I really like the result. It's effectively a phantom type so far, right? Still, very useful for safety and documentation via types. No opinion about the last commit: both variants have similar number of the `makeFrame` tags.

andorp:
@Mikolaj 

> It's effectively a phantom type so far, right?

Not really, it is used for page creation in the line: https://github.com/liqd/aula/pull/132/files#diff-f9c8fb0392f16adce432e902ecba7b78R233 . The diff is hard to read at that part, sorry about that.

Mikolaj:
@andorp: I was looking at individual commit diffs this time and they are reasonably readable. I only had to verify manually that the code of formRedirectH' was moved over without any tricky changes. Other than that it's clear how commit descriptions match the diffs.

Re phantom type `p`, yes, I see it's really used in redirectFormHandler, but it was used so before, as well. You just tied it into FormHandler and so to the many functions that use FormHandler, which is great. It's "phantom" in the sense that it's neither a container inside FormHandler nor are there any extra class dispatches on it that weren't there before (but there may be in the future, perhaps).

andorp:
@Mikolaj  I got your point. Thanks for the teaching. :)

fisx:
LGTM.

- I agree that a95525b is not much of an improvement and reverted it.
- Two trivial commits.
- I'm not sure if the comment on `formRedirectH'` regarding the servant PR is really obsolete.  @np?

waiting for travis.  will merge when green.

PR #136, opened 2016-02-28
User Profile: Created Ideas and Delegated Votes

andorp:


fisx:
I still think it would be more consistent to have `settings` under `/user/<id>`, not under `/user`.  But I guess it's mostly a style question.

Otherwise LGTM.

PR #137, opened 2016-02-29
Redirect from `/` to `/space`.

fisx:
- Export `redirect` from `Frontend.Core`.
- Add instance for `ToHtml ()`.
- Add path `Top -> "/"`.
- Add route `/` redirecting to `/space`.

### DIFF
--- a/src/Frontend.hs
+++ b/src/Frontend.hs
@@ -61,6 +61,7 @@ type AulaTop =
        (AulaMain :<|> "testing" :> AulaTesting)
   :<|> "samples" :> Raw
   :<|> "static"  :> Raw
+  :<|> GetH (Frame ())
 
 
 aulaTop :: (Action :~> ExceptT ServantErr IO) -> Server AulaTop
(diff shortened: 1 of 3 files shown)


src/Frontend.hs
```diff
@@ -69,6 +70,7 @@ aulaTop (Nat runAction) =
   :<|> (\req cont -> getSamplesPath >>= \path ->
           waiServeDirectory path req cont)
   :<|> waiServeDirectory (Config.config ^. htmlStatic)
+  :<|> redirect "/space"
```
andorp: This should be changed to `/login` in the future.

src/Frontend/Core.hs
```diff
@@ -59,6 +60,11 @@ instance (ToHtml (HtmlT Identity ())) where
     toHtmlRaw = toHtml
     toHtml = HtmlT . return . runIdentity . runHtmlT
 
+-- | FIXME: Could this be a PR for lucid?
+instance (ToHtml ()) where
```
andorp: Extra parenthesis.

andorp:
LGTM

src/Frontend.hs
fisx: I disagree.  If you try to hit a page and don't have access, you should be redirected to `/login`.  If you are loading `/`, you should be pointed to some page that is most likely to be usedful directly **after** login.  Even better than `/space` would be the current year's class idea space for students, the admin workbench for admins, and so on.  but this is still sci-fi...

andorp: Got your point. It is valid.

PR #138, opened 2016-02-29
Improve idea space handling.

fisx:
- add some idea spaces to initial db
- make idea space overview clickable

### DIFF
--- a/src/Api/Persistent.hs
+++ b/src/Api/Persistent.hs
@@ -160,8 +160,13 @@ getSpaces = getDb dbSpaces
 getIdeas :: Persist [Idea]
 getIdeas = getDb dbIdeas
 
-addIdeaSpace :: IdeaSpace -> Persist ()
-addIdeaSpace ispace = modifyDb dbSpaceSet $ Set.insert ispace
+-- | If idea space already exists, return 'False'.  Otherwise, create it and return 'True'.
+addIdeaSpace :: IdeaSpace -> Persist Bool
+addIdeaSpace ispace = do
+    exists <- (ispace `elem`) <$> getSpaces
+    if exists
+        then return False
+        else modifyDb dbSpaceSet (Set.insert ispace) >> return True
 
 addIdea :: Proto Idea -> Persist Idea
 addIdea = addDb dbIdeaMap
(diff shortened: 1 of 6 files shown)


src/Api/Persistent.hs
```diff
@@ -159,6 +160,14 @@ getSpaces = getDb dbSpaces
 getIdeas :: Persist [Idea]
 getIdeas = getDb dbIdeas
 
+-- | If idea space already exists, return 'False'.  Otherwise, create it and return 'True'.
+addIdeaSpace :: IdeaSpace -> Persist Bool
```
andorp: This should has type `IdeaSpace -> Persist ()` to hide information, and thow an error on data duplication.

fisx: but if it throws an error, it won't hide information, it will just be more complicated to call it if you don't care whether it's there or not.

I don't think hiding the information whether an idea space exists is interesting from a security perspective.

I'm just thinking that I want to rename it to `addIdeaSpaceIfNotExists`.

andorp: The runPersist should have `Persist :~> ErrorT PersistError IO` instead of `Persist :~> IO` . And with that you can throw an error if the entity exist already. Or connection error happened, etc.

fisx: yes, but i don't want exceptions here.  even if we had them i didn't want them.

are you sure you do?  if so perhaps we should talk about this tomorrow.

tests/Api/PersistentSpec.hs
```diff
@@ -125,6 +125,25 @@ spec = do
     let getIdeasWithTopic = filter (isJust . view ideaTopic) <$> getIdeas
     findAllInBySpec "findIdeasByTopicId" getIdeasWithTopic findIdeasByTopicId (ideaTopic . _Just) changeAUID
 
+    describe "addIdeaSpace" $ do
+        let test :: (Int -> Int) -> IdeaSpace -> SpecWith (Persist :~> IO)
+            test upd ispace = do
+                it ("can add " <> showIdeaSpace ispace) $ \(Nat rp) -> do
```
andorp: This is like BDD: Given something, do something, test someting based on the previous and after state. No if in testcase. Just high level combinators! Big like! :)

andorp:
LGTM

PR #139, opened 2016-02-29
Fix: sensei didn't type check AllModulesSpec without a test case.

fisx:


fisx:
will merge myself when travis is green.

PR #140, opened 2016-02-29
2016 02 29 mf fix type checking with sensei+1

fisx:


fisx:
(didn't wait on #139 after all, and of course screwed it up...)

PR #142, opened 2016-02-29
File upload forms

fisx:
Upload user csv and crash with the parsed contents. (Requires thentos PR#501.)


fisx:
@andorp re. the last commit with the test, i'd be interested in your ideas how to make this test better.  apart from the fact that it's not done yet and fails, that is.  (-:

fisx:
This PR can't pass travis because the docker image does not contain thentos-0.3.0.  In order to be able to upgrade the docker image, I will merge this branch and re-run travis tomorrow on the new docker image.

Please review post-merge!

Mikolaj:
will do :)

fisx:
hum, travis re-run doesn't appear to do any good.  everything works on my laptop, though!

src/Frontend/Page/FileUpload.hs
```diff
@@ -0,0 +1,77 @@
...
+
+type CsvUserRecord = [SBS]
+
+batchCreateUsers :: forall m. (ActionTempCsvFiles m, ActionM m)
+      => ServerT (FormHandler BatchCreateUsers ST) m
+batchCreateUsers = redirectFormHandler (pure BatchCreateUsers) q
+  where
+    q :: BatchCreateUsersFormData -> m ()
+    q (BatchCreateUsersFormData _clname Nothing) =
+        throwError $ err500 { errBody = "uploaded failed: no file!" }
```
Mikolaj: typo; should be: "upload FAILED:"

src/Frontend/Page/FileUpload.hs
```diff
@@ -0,0 +1,77 @@
...
+    q (BatchCreateUsersFormData clname (Just file)) = do
+        let schoolcl = SchoolClass theOnlySchoolYearHack clname
+        eCsv :: Either String [CsvUserRecord] <- popTempCsvFile file
+        case eCsv of
+            Left msg -> throwError $
+                err500 { errBody = "parsing upload FAILED: "    <> cs msg }
+            Right records -> throwError $
+                err500 { errBody = "parsing upload SUCCEEDED: " <> cs (show (schoolcl, records)) }
```
Mikolaj: if we tweak the texts anyway, perhaps put "parsing of CVS upload" in both places?

src/Frontend/Core.hs
```diff
@@ -247,13 +251,15 @@ redirectFormHandler getPage processor = getH :<|> postH
...
         page <- getPage
         let fa = absoluteUriPath $ formAction page
+            env = getFormDataEnv formData
         (v, mpayload) <- postForm fa (processor1 page) (\_ -> return $ return . runIdentity . env)
-        case mpayload of
+        (case mpayload of
             Just payload -> processor2 page payload >>= redirect
-            Nothing      -> renderer page v fa
+            Nothing      -> renderer page v fa)
+            `finally` cleanupTempCsvFiles formData
```
Mikolaj: Indeed, seems like a good place to clean up. But we need to be sure the CSV files are already fully read by this point (lazy IO). I can't deduce it from the type of the `processor` argument, so I'd add a contract to the haddock ("The @processor@ argument should read any opened files eagerly, because they are closed at the end of the call"). Fortunately, I can see the code of our processor for CSV files *is* strict enough, because it parses the whole file to decide whether to throw an error. We should be fine as long as we don't let users upload files for future use (e.g., if they upload an avatar, we should display it while still serving the same request or at least calculate its size, or something).

Mikolaj: Note that as long as we read (or write to) files eagerly and close them in obviously safe places (e.g., a parent thread of all potentially file-opening threads, after they all terminate), we don't need to use `resourceForkIO`, which is one of the main complexities of the `resourcet` engine and it's use pattern. Let's keep it that way.

fisx: Interesting!  I will add the second comment to the haddocks.

As for the strictness, what about another constraint?:

```haskell
redirectFormHandler
    :: (FormPageView p, Page p, ActionM m, NFData (FormPageResult p))
    => [...]
```

Then we can call `deepseq` on `payload`; probably after the processor has been called.  Not in the spirit of squeezing the best possible performance out of the system, but much safer.

Mikolaj: Am I right it's the `processor` argument in which the call to `popTempCsvFile` place?

It would be great if we could ensure non-laziness of the IO with `deepseq`, but I think we would need to hunt down any data resulting from `process` and call `deepseq` on it, or ensure strictness of all such data via types. The latter usually involves changing tons of code to only use strict types. I don't know the details of our data flow, but I think the type doesn't prevent the `process` argument from reading files and placing their content in state (persistent or not) of the system lazily without any relation to `payload`. Do we have guarantees the fields of our app state, whether the persistent part or some transient state, is strict? Additionally, are state updates in our monad strict enough (they usually are not)?

Mikolaj:
All resolved in #171. Thank you!

PR #144, opened 2016-03-01
Admin Pages

andorp:

PR #149, opened 2016-03-02
Generics based Arbitrary instances.

andorp:
During the development if we change the structure of our data, we had to adjust the arbitrary instances, which is kind of boilerplate. Usage of generics based arbitrary speeds up the development process.

### DIFF
--- a/aula.cabal
+++ b/aula.cabal
@@ -28,6 +28,7 @@ library
       base >=4.8 && <4.9
     , thentos-core >=0.3.0
     , aeson
+    , basic-sop
     , binary
     , bytestring
     , case-insensitive
@@ -39,6 +40,7 @@ library
     , directory
     , filepath
     , functor-infix
+    , generics-sop
     , http-types
     , lens
     , lucid
@@ -100,6 +102,7 @@ executable aula-server
       base >=4.8 && <4.9
     , thentos-core >=0.3.0
     , aeson
+    , basic-sop
     , binary
     , bytestring
     , case-insensitive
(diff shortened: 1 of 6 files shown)


kosmikus:
Generally, I'm obviously in favour of using generics. But you're changing current behaviour in a few places, aren't you? In particular where you're replacing code using `arbWord` and `arbPhrase`?

andorp:
@kosmikus You are right, I did two changes unintentionally. I am going to fix them.

fisx:
thanks @kosmikus, all good points!

We just talked about your suggestion to give up on sop and decided against it.  The idea is that we have two kinds of random data: true noise, and consistent system states generated by arbitrary valid user behavior.  The former is what we do in `src/Arbitrary.hs` (which btw will go to `tests/` at some point); the latter will happen in the future.

I just opened related issue #152.

Does that make sense to you?

fisx:
@andorp, could you fix the other comments, wait for #144 to go to master, and then rebase / merge?

thanks!

kosmikus:
@fisx You misunderstand me. I certainly don't advocate to get rid of sop or generics. I say that the two concrete instances which deviate from the generic form should just be given explicitly. Alternatively, introduce a newtype for titles and give it the appropriate `Arbitrary` instance.

fisx:
@kosmikus: ah, ok.  then i just rephrased your suggestion in #152 without noticing?  that means we agree.  good.  (-:

fisx:
LGTM.  Thanks!

PR #153, opened 2016-03-03
Fix vector version to 0.10.12.3

andorp:


np:
LGTM

np:
The plan is that this might fix the docker image

andorp:
Thanks

Issue #15, opened 2016-02-10
quay.io vs dockerhub


andorp:
Quay.io could be an alternative docker hub for us.

Tasks:
 * Make a comparison on dockerhub vs quay.io
 * Figure out the necessary work to change to quay.io
 * Write a plan how we can do the migration

Related repositories:
 * https://github.com/liqd/aula
 * https://github.com/liqd/aula-docker

andorp:
The `https://quay.io/repository/liqd/` organization is created.

PR #155, opened 2016-03-03
travis now uses docker images from quay.io/liqd/aula

andorp:


andorp:
Fixes #15

andorp:
Apart from the hlint failure, the build is ok.
LGTM

PR #158, opened 2016-03-04
Fix hspec versionn to 2.2.2

andorp:


andorp:
Agreed with @fisx to merge it.

PR #159, opened 2016-03-04
Add generics to stack.yml

andorp:


kosmikus:
Any particular reason why we're not using `lts-5` which has `generics-sop-0.2.1.0`?

np:
We'll switch to a more recent stackage version (such as lts-5), however the number of dependencies to bump is not negligible so we postpone that.

kosmikus:
@np Ok, fair enough.

PR #168, opened 2016-03-07
Make form payload strict.

fisx:
Also, various bits and pieces.

See also: https://github.com/liqd/aula/pull/142/files#r55120481


fisx:
copying https://github.com/liqd/aula/pull/142/files#r55197705:

> Am I right it's the processor argument in which the call to popTempCsvFile takes place?

> It would be great if we could ensure non-laziness of the IO with deepseq, but I think we would need to hunt down any data resulting from process and call deepseq on it, or ensure strictness of all such data via types. The latter usually involves changing tons of code to only use strict types. I don't know the details of our data flow, but I think the type doesn't prevent the process argument from reading files and placing their content in state (persistent or not) of the system lazily without any relation to payload. Do we have guarantees all the fields of our app state, whether the persistent part or some transient state, are strict? Additionally, are state updates in our monad strict enough (they usually are not)?

fisx:
irc exerpt:

```
16:24 < mikolaj> let me check my understanding:
16:24 < mikolaj>  it's the processor argument in which the call to popTempCsvFile takes place, right?
16:26 < mikolaj> 2. if so, it's @processor@ that opens files and reads from them (possibly lazily)
16:27 < mikolaj> 3. if so, `payload` is not the result, but the input of the computations that opens files
16:27 < fisx_> ah.
16:27 < fisx_> yes, what i did doesn't work yet.
16:27 < fisx_> or at all.
16:27 < fisx_> you're right.
16:27 < fisx_> payload contains a 'Maybe FilePath'.
16:27 < fisx_> forcing that is not the solution!  (:
16:27 < mikolaj> heh
16:28 < mikolaj> fisx_: but generally, I'd not overengineer for now, we just have one case and it's
simple, so an English contract may be OK; unless you have some really smart solution based on
types (this one was smart, just not enough:)
16:30 < mikolaj> and defnitely going all the way with strictness for the sake of lazy IO alone 
                 is not worth it
16:30 < mikolaj> (but generally, some strictness is goog; e,g., with the new GHC we may set all 
                 datatype fields to be strict by default and save some memory and latency for 
                 free)
```

andorp:
LGTM

fisx:
Closing in favour of #171.

PR #171, opened 2016-03-07
Bit and pieces

fisx:
This obsoletes #168.  I will take another shot at the `NFData` idea in a separate PR.  @Mikolaj, could you take a quick look if the haddocks for `redirectFormH` make sense now?  (Part of that text may look familiar.   :-)

### DIFF
--- a/src/Arbitrary.hs
+++ b/src/Arbitrary.hs
@@ -36,6 +36,7 @@ import qualified Frontend.Path as P
 
 
 -- | FIXME: push this upstream to basic-sop.
+-- See also: https://github.com/well-typed/basic-sop/pull/1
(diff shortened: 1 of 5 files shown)


fisx:
I think the `NFData` idea is not interesting.  I have tweaked `instance ActionTempCsvFiles Action`.  This time I `deepseq`ed the parsed file contents, not the file name, but while I did I realized that (a) strictness of the monadic values inside the monad are still tricky to reason about, and (b) `deepseq` doesn't help with that.  It looks like it's ok, and @Mikolaj seems convinced of it, too.

I therefore propose to see if the haddocks are plausible enough as they are, and then merge this and move on?

Mikolaj:
Good decision. Indeed, one can get weary of lazy IO. :) Merging.

PR #179, opened 2016-03-09
Abstract over persistence implementation

Mikolaj:
Please feel free to review and contribute while still in progress.

### DIFF
--- a/src/Action.hs
+++ b/src/Action.hs
@@ -1,4 +1,5 @@
 {-# LANGUAGE FlexibleContexts           #-}
+{-# LANGUAGE FlexibleInstances          #-}
 {-# LANGUAGE FunctionalDependencies     #-}
 {-# LANGUAGE GeneralizedNewtypeDeriving #-}
 {-# LANGUAGE LambdaCase                 #-}
@@ -81,13 +82,13 @@ class ( ActionLog m
       , ActionTempCsvFiles m
       ) => ActionM m
 
-instance ActionM Action
+instance ActionM (Action Persist)
 
 class Monad m => ActionLog m where
     -- | Log events
     logEvent :: ST -> m ()
 
-instance ActionLog Action where
+instance ActionLog (Action r) where
     logEvent = Action . liftIO . print
 
 class (MonadPersist r, Monad m) => ActionPersist r m | m -> r where
@@ -97,13 +98,13 @@ class (MonadPersist r, Monad m) => ActionPersist r m | m -> r where
     -- complex computations.
     persistent :: r a -> m a
 
-instance ActionPersist Persist Action where
+instance MonadPersist r => ActionPersist r (Action r) where
     persistent r = Action $ ask >>= \(Nat rp) -> liftIO $ rp r
 
-instance MonadIO Action where
+instance MonadIO (Action r) where
     liftIO = Action . liftIO
 
-instance MonadPersist Action where
+instance MonadPersist r => MonadPersist (Action r) where
     -- getDb :: AulaGetter a -> m a
     getDb = persistent . getDb
     -- modifyDb :: AulaSetter a -> (a -> a) -> m ()
@@ -117,7 +118,7 @@ class Monad m => ActionUserHandler m where
     -- | Make the user log out
     logout :: m ()
 
-instance ActionUserHandler Action where
+instance MonadPersist r => ActionUserHandler (Action r) where
     login user = do
         put $ UserLoggedIn user "session"
         persistent $ loginUser user
@@ -130,9 +131,9 @@ instance ActionUserHandler Action where
 
 class MonadError ActionExcept m => ActionError m
 
-instance ActionError Action
+instance ActionError (Action r)
 
-instance GenArbitrary Action where
+instance GenArbitrary r => GenArbitrary (Action r) where
     genArbitrary = Action . liftIO $ generate arbitrary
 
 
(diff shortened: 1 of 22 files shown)


src/Api/PersistentImplementation.hs
```diff
@@ -0,0 +1,47 @@
+{-# LANGUAGE GeneralizedNewtypeDeriving  #-}
+{-# LANGUAGE ImpredicativeTypes          #-}
+{-# LANGUAGE OverloadedStrings           #-}
+{-# LANGUAGE ScopedTypeVariables         #-}
+{-# LANGUAGE TypeOperators               #-}
+
+{-# OPTIONS_GHC -Wall -Werror -fno-warn-orphans #-}
+
+module Api.PersistentImplementation
```
andorp: What about `Api.Persistent.STM` ?

Mikolaj: You mean "How about `Api.Persistent.STM`?" in the sense that you propose to rename the module? A good point, given that we already know we will have more than one implementation. Normally, I'd even move the implementation to a distant directory called `Implementations`, somewhere close to the top-level, together with an implementation(s) of `ActionM`, etc. to make sure people are aware they are mucking around with implementation details. How about `Api.PersistentImplementation.STM` as a compromise?

andorp: Sorry for the short question. I would say, as we have a `Persistent` module already we could move the implementation to the `Persistent.Implementations.STM` or `Persistent.STM`. And expose the current implementation in the `Persistent` module? What do you think?

Mikolaj: Oh, cool, `Persistent/Implementations/STM` is even better, because it's distant from `Api/Persistent.hs`. BTW, what is our convention about singular vs plural in names? Would not `Implementation/` be better (no 's')?

But I'd not expose the implementation in `src/Persistent.hs`, precisely because I'd like to clearly mark (by having to import *.Implementation.*) any code that reaches beyond the abstraction barrier and so depends on one of the actual implementations (even if it's the best one, the chosen one), not on our common API for it.

andorp: You are right in both cases.

I think we don't use plural in module names. The `Persistent.Implementation.STM` is better.
I agree with you to not to expose the implementation in `src/Persistent.hs`. :)

Mikolaj: Splendid, ta, will fix before merging.

src/Action.hs
```diff
@@ -67,8 +70,12 @@ data UserState
 
 makeLenses ''UserState
 
+-- TODO: here @ActionPersist Persist@ is hardwired, so either
+-- parameterize ActionM by the persistance implementation
+-- or remove ActionPersist from ActionM and add by hand elsewhere
+-- (messy and lots of places need to be touched).
 class ( ActionLog m
-      , ActionPersist m
+      , ActionPersist Persist m
```
Mikolaj: @andorp, @fisx: this is still a major abstraction leak. As soon as @fisx generally approves of the effort (earlier, we would risk wasting time), could you have a look here and suggest something? I'd lean towards parameterizing `ActionM` but this is the most important monad ever, so probably I have no clue what I'm proposing.

andorp: I am ok with the further parameterizing of the ActionM monad. I had the same idea, for me it clearly says that for any Action we implement has to be independent of which persistent implemantation we use.

fisx:
Regarding

```haskell
 class ( ActionLog m
-      , ActionPersist m
+      , ActionPersist Persist m
```

Just talked this through with @andorp.  I agree with him that `Action` should get an extra type parameter:

```haskell
newtype Action p a = Action (ExceptT ActionExcept (RWST (p :~> IO) () UserState IO) a)
```

This way, we are forced to settle for both the persistence implementation and the action implementation in `runFrontend`, and make both choices explicit.

Then the `p` in `Action p a` needs to instantiate `MonadPersist`, but I think it's enough to add that constraint to the code that uses the `Action` type rather than to the type itself.

Btw, shouldn't `MonadPersist` be called `PersistM`?

Mikolaj:
Yes, `PersistM` makes sense. We don't define any `MonadFoo` in our code other than `MonadPersist`.

Agreed about the extra parameter both to `Action` and `ActionM`. Will see how it fits the code (quite a few places affected, but that's good). I call the paramter `r`, not `p`, because `p` is already taken for `Page`, etc., and sometimes conflicts.

Will go though with this, without moving the location of the `Implementation` file yet.

Mikolaj:
Done. Only renaming the implementation file remains (and doing the same leak detection for the implementation of ActionM, but this is future work and surely not in this commit, which is about PersistM implementation leak detection).

fisx:
I found these:

```
$ git grep Persistent.Implementation
Frontend.hs:aulaTop :: (Action Persistent.Implementation.STM.Persist :~> ExceptT ServantErr IO) -> Server AulaTop
Frontend.hs:catchAulaExcept :: (m a ~ (ServerT api (Action Persistent.Implementation.STM.Persist)))
```

Should we generalize that, too?  Or add a comment why we can't / don't want to?  Make that a FIXME?

Mikolaj:
Yes, there are just the two I commented above about. Nothing deep, I think. Will FIXME.

fisx:
> No, unfortunately, in this case I couldn't easily fix this. But I agree, this can be parameterized further, just with some extra work that is hard to justify at this stage.

Ah, I see.  Another case for monad-base?

fisx:
Nice work, thanks!  Please re-assign to me when you are done with my comments.

fisx:
Thanks!

PR #183, opened 2016-03-10
Steal & mutate thentos release script.

fisx:


make-version.sh
```diff
@@ -0,0 +1,25 @@
+#!/bin/bash
+
+set -e
+cd `dirname $0`
+export NEW_VERSION="$1"
+export ONLY_RUN_ON_BRANCH="2016-03-10-mf-make-release-script"
```
andorp: I think this should be "master".

make-version.sh
```diff
@@ -0,0 +1,25 @@
...
+
+test -z "$NEW_VERSION" \
+    && ( echo "Please provide a version number to move to."; exit 1 )
+git status -b | grep -q "On branch $ONLY_RUN_ON_BRANCH" \
+    || ( echo "Please only run this script on master."; exit 1 )
+
+perl -i -pe 's/(version:\s*).*/${1}'"$NEW_VERSION"'/' package.yaml
```
andorp: I don't speak perl. @Mikolaj could you take a look, maybe you have better knowledge about the topic. :)

make-version.sh
fisx: oops.  thanks, will fix.

make-version.sh
Mikolaj: I'm worked in Perl for a few months, but I was happy to forget it. :) Looks legit, does it work? `(version:\s*)` finds and skips `version:` and all spaces, `.*` finds the number to replace, extending to the end of the line, which is replaced by the first argument to the expression, which is `$NEW_VERSION`.

Mikolaj:
Haven't tested, but LGTM.

PR #184, opened 2016-03-10
Separate pure code and effects better in MetaInfo creation.

fisx:


fisx:
@Mikolaj: assigning to you, as it may get in your way, and you may have an opinion on this.

Related from irc:

```
15:09 < fisx> ...  we will need to pass in 'currentUser' for virtually all Persist business.
15:10 < fisx> it can't be in Persist, but..
15:11 < fisx> it is required for everything that creates or changes things that have MetaInfo.
15:11 < fisx> to keep track of 'last changed by'.
```

Mikolaj:
Thank you, will see where I get with that and then report back.

Mikolaj:
Yes, my experiments with user session indicate this will be very useful to me. Merging.

PR #185, opened 2016-03-10
2016 03 cc create idea style

local-girl:
I'm not sure if the image seletor works. code wise it does everything its supposed to but if you unhide the radio buttons:
Inspect (as in chrome's code inspector) the label "Kann deine Idee einer..."
click the element ```<div class="category-radios">```
remove the style ```position:absolute```
Now click the image selector - you can see the radios don't highlight properly. I'm interested to see if the data is correct, I'll come back to it as I want to move on.


fisx:
@local-girl: looks great, thanks!

- I added very few nit-picks.
- I didn't test the category assignment, but elaborated on what I want to do there in the FIXME.
- I reverted the hidden-field thing after all because i think ideally we never want to have to use it.  if you decide you do you can dig it out of the commit history (i made an explicit revert commit).

@andorp: could you take a quick look at the 'changed files' tab (just the haskell code), and merge if you and travis are happy?

fisx:
(did the review-of-review offline.)

PR #188, opened 2016-03-11
Admin Page: Edit User

andorp:


Mikolaj:
@andorp : Your hpack apparently generates cabal sections in different order. Could you upgrade hpack and try again and commit aula.cabal that has a smaller diff wrt our versions? Thank you.

Mikolaj:
Otherwise, LGTM, though I've done a rather quick scan this time. Will merge as soon as the aula.cabal is tweaked.

fisx:
@Mikolaj I think I've observed before that even the same hpack version gets the order different each time.  I think we should just ignore that noise as long as we keep aula.cabal under version control.

Mikolaj:
Oh, a pity, let me merge then. I though it was the version difference (0.9 vs. 0.10).

Mikolaj:
Huh, actually, we got conflicts here now. So, a good place to stop for today. :)

src/Frontend/Page/Admin.hs
```diff
@@ -342,15 +402,36 @@ adminSettingsGaPUsersCreate =
...
+    editUser (EditUser role clss) = do
+        let isSelectedClass (Student clss')    = clss == clss'
+            isSelectedClass (ClassGuest clss') = clss == clss'
+            isSelectedClass _                  = False
+
+        let addGroup gs = case role of
+                RoleStudent -> Student clss : gs
+                RoleGuest   -> ClassGuest clss : gs
+
+        persistent $ modifyUser uid (userGroups %~ addGroup . filter (not . isSelectedClass))
```
fisx: This is confusing, could you write a comment what the application logic is that you had in mind?  I thought we'd probably want to just overwrite 'userGroups' and always keep a singleton in it.  Or, if that's not too much refactoring, change userGroups to `userGroup :: Group`.

fisx:
Except for the one comment this is great, thanks!  @andorp back to you.  Please re-assign me before merging?

src/Frontend/Page/Admin.hs
andorp: We agreed that the userGroup list will contain classes from the past. The semantics is not well defined here. I would leave the actual implementation as it is now, and work on the full solution in a different PR. Of course I am going to write a documentation comment what is happening now. Maybe some tests, which are also convenient here.

andorp: Relates #197

fisx:
Nice.  Made a few nit-picks; can be merged if travis is green.  Thanks!

PR #189, opened 2016-03-11
Scrap some boilerplate form FormHandler

andorp:


Mikolaj:
Uhm, nice, LGTM. Merging.

fisx:
Why is this an improvement?  (I have no real objections and I'm outvoted so it's fine; just wondering if it wasn't fine before and not worth the noise.)

Mikolaj:
I may be missing something, but I think it's just a lot of tiny snippets of code eliminated and also the only couple of places where the snippets are different, stand out now (not so much via the different combinator name, but by the fact, that you actually mention the argument, which is omitted elsewhere).

andorp:
I removed the `ST` parameter, because It happened me third time, that I had to answer the following question: "Which parameter should I pass to formHandler to make the typechecker happy?", and `ST` had no real information. Anyway there are some cases when we pass different parameter than `ST`. We should check those cases, probably they work differently, than the standardized ones, but I am not sure, if they should work differently.

PR #194, opened 2016-03-11
Get fake session from a TVar that acts like a fake cookie, instead of from Persist

Mikolaj:


Mikolaj:
Have a look at commit "Take UserState from fake cookie (has no effect due to adminUsernameHack)". It has no effect at the (re)start of the server, due to 

```
    -- FIXME: Login shouldn't happen here
    runActionForceLogin = Nat $ \action -> runAction $ do
        Action.login adminUsernameHack
        action
```

which is probably needed for our demo server to be friendly. Otherwise it would require the user to login first, before being able to browse the site.

If the server doesn't restart too often (e.g., once per request --- I have no clue how it multiplexes under load), subsequent logouts and logins should work OK, though (not tested yet, neither manually nor automatically).

fisx:
it's ok if the users have to login, since they have to do that in the production server as well.  for now, i would expect login to work without any checks, if you enter a user name and any password, you are logged in under that name.  once authentication actually works, we just need to create a few scaffolding users and circulate the passwords for those.

we will have a dedicated demo server in addition to the staging server we are using now, and it will be restarted less regurlarly.  for those, it's ok to lose all state (even the database).

Mikolaj:
Just tested manually with the current state of this PR: 
* we indeed start logged in as 'admin'
* logout works fine
* login as 'admin' works fine with any password
* login as 'foo' results in "Something went wrong", which is OK (we do check that the user is in DB, because we take his data from DB, needed to update Action UserState)
* somehow going to *any* other page after logout results in auto-login as 'admin', which puzzles me, because I thought the auto-login occurs only at server restart; looking again at the code I see `enter runActionForceLogin`, which probably means that each action is prepended by the auto-login; not just at server restart

So I removed the auto-login (not pushed yet):
* now one needs to login as 'admin' to browse the demo server
* since we start at page `space`, not at `login`, the starting screen of the server is "Something went wrong"
* but logout is now persistent and logging as other users should be as well (will test, but I'd have to add other users to the initial DB first)
* some tests fail now, will investigate

fisx:
> I see enter runActionForceLogin, which probably means that each action is prepended by the auto-login; not just at server restart

yes, that's indeed not good.  i think we can now remove that hack, since logging in as admin should work fine without.  (fin*er*, in fact.)

now we want the "authentication" error (I think we somewhere check `isPrivatePage` and that fails) to result in a redirect to the login page.  if you can do this in this PR, great, otherwise, could you open an issue?

Mikolaj:
@fisx: I think I'm done, apart of redirecting to the login page when no user is logged in and we try to access a private page. I tried and failed in the last commit; help! :)

Mikolaj:
@fisx: I think the redirect should be completed in this PR so that our staging server is not in a state that puzzles the user (currently, after this PE, it would be --- internal server error at the very start). After the redirection is done, it should be user friendly enough, redirecting to login page from the start, 'admin' can log in without password, 'admin2' replaces him, when needed, and everything work. Though, perhaps we could hint on the login page to login as admin? To be super user friendly, until we really deploy?

src/Action.hs
```diff
@@ -104,12 +110,6 @@ instance PersistM r => ActionPersist r (Action r) where
 instance MonadIO (Action r) where
     liftIO = Action . liftIO
 
-instance PersistM r => PersistM (Action r) where
```
andorp: This instance should stay. @np idea is to remove all the non necessary `persistent` calls from the codebase. I want to do that, after this PR is merged. Please let this instance live. :)

src/Types.hs
```diff
@@ -258,7 +258,6 @@ phaseName = \case
 ----------------------------------------------------------------------
 -- user
 
--- | FIXME: introduce newtypes 'UserLogin', 'UserFirstName', 'UserLastName'?
```
andorp: Good catch.

src/Action.hs
```diff
@@ -167,14 +181,27 @@ type ActionExcept = ServantErr
...
-mkRunAction :: (r :~> IO) -> UserState -> (Action r :~> ExceptT ServantErr IO)
-mkRunAction persistNat = \s -> Nat (run s)
+-- - The ability to change the state is missing. FIXME: which state? login modifes UserState all right
+-- - The state should be available after run. FIXME: which state? what for?
+mkRunAction :: forall r. PersistM r
+            => (r :~> IO) -> Action r :~> ExceptT ServantErr IO
+mkRunAction persistNat = Nat run
   where
-    run s = ExceptT . fmap (view _1) . runRWSTflip persistNat s . runExceptT . unAction
+    run = ExceptT . fmap (view _1) . runRWSTflip persistNat UserLoggedOut . runExceptT . (setCurrentUser >>) . unAction
```
andorp: Good catch.

src/Action.hs
Mikolaj: If we are convinced you know what you are doing (as for me, I'm only guessing based on the comments for the `persistent` vel `atomically` operation), please just revert the commit:  "Remove the dangerous PersistM instance for Action. When using the instance there is little control over authorization, atomicity, etc. This is best handled explicitly."

src/Action.hs
Mikolaj: Yep. That was the main purpose of this overgrown PR.  To set the user state for each newly created action based on the (fake in-TVar) cookie, instead of clearing it and then auto-logging the admin user.

src/Action.hs
andorp: I imagine the authorization as follows (as a vague idea):

````haskell
class Authorize a where
    authorizeGet :: ActionM r m => Username -> AulaGetter a -> m ()
    authorizeSet :: ActionM r m => Username -> AulaSetter a -> m ()

getDB getter = do
    user <- currentUser
    authorizeGet user getter
    persisistent $ getDB getter
````

Comment (np) [inline src/Action.hs]: I like this idea.

**Comment by np** [comment]
In overall this LGTM. This PR improves the status of "current user hack" by a great deal.

The part where I can't set my mind to agree is the "pass the current user everywhere in Persist". I'm tempted to emphasis the distinction between writing to the state and reading from it using an other type class: `UpdatePersistM m` which refines `PersistM m` and also gives access to the current user.

File: src/Action.hs (hunk 1 above)
Comment (Mikolaj) [inline src/Action.hs]: I was referring to these remarks, not sure how it ties in:

```
class (PersistM r, Monad m) => ActionPersist r m | m -> r where
    -- | Run @Persist@ computation in the action monad.
    -- Authorization of the action should happen here.
    -- FIXME: Rename atomically, and only call on
    -- complex computations.
    persistent :: r a -> m a
```

Comment (andorp) [inline src/Action.hs]: You are right. Probably changing the `Monad` in `Authorize` will do the trick.

````haskell
class Authorize a where
    authorizeGet :: PersistM m => Username -> AulaGetter a -> m ()
    authorizeSet :: PersistM m => Username -> AulaSetter a -> m ()
````

Comment (andorp) [inline src/Action.hs]: I prefer the first better. And write different and explicit authorization when we use the `persistent` combinator.

**Comment by Mikolaj** [comment]
@np: I'm glad you like it. Please feel free to tweak it to your taste as you move forward with accommodating thentos session.
Re passing user explicitly in Persist functions, I tried to avoid that in the first commit, introduced a class HasCurrentUser, but then reverted the commit for the reasons explained in the commit message. It turned out there are only half a dozen, or perhaps a dozen including tests, places where the user has to be passed to Persist functions, so far. Furthermore, it should scale well, if we don't make Action an instance of the PersistM monad, but wrap any Persist operations in ActionM wrappers. But if we really want to use the Persist operations a lot on the Action level (through PersistM Action instance), then the user argument gets in the way, indeed. Please experiment and decide.

**Comment by fisx** [comment]
@np I think I agree with your `UpdatePersistM` idea.  Open an issue or rather do it now?

Anyway I'll check the merge conflicts now and read the commits.

File: package.yaml [hunk 4]
```diff
@@ -85,7 +85,6 @@ tests:
     main: Spec.hs
     source-dirs:
       - tests
-      - src
```
fisx: Why?  I'm not sure, but don't we need this for sensei to pick up changes under `./src/` when re-running the test suite?

src/Action.hs
```diff
@@ -192,14 +219,17 @@ modifyCurrentUser :: (ActionPersist r m, ActionUserHandler m) => (User -> User)
 modifyCurrentUser f =
   currentUser >>= persistent . flip modifyUser f . (^. _Id)
 
+noCurrentUser :: ActionUserHandler m => m Bool
+noCurrentUser = (UserLoggedOut ==) <$> userState
```
fisx: I would call this `isLoggedIn` and invert its value.  Similarly, when used in `Frontend.Core`, I would remove the negation from `noLoggedInUser`.

fisx:
@Mikolaj as for the redirect-to-login-page, I would just raise an exception in `makeFrame`, and we have a function `redirect` for that in `Frontend.Core`.  It means we have to add a lot of constraints all over the place, but otherwise I would expect this to work.  I will do this once I'm done with reading the rest of the code.

Do you have any objections?

package.yaml
Mikolaj: I hope sensei is unaffected, because after `make sensei`, I'm seeing `test -d .stack-work/ && echo "stack exec --" || echo "cabal exec --" sensei -isrc -itests tests/Spec.hs`, note the `-isrc`. The automatic reloading and testing doesn't work for me, but this is unrelated and I don't care, I'm fine with my universal emacs setup, which doesn't overheat my laptop, until I tell it to. :)

In my setup, I'm just building with 'cabal build' (in my emacs window) and with the 'src' it compiles all the modules again for the tests target, hence the change. I guess it's doing the same in other setups, too, compiling the modules twice, instead of using their already compiled (or interpreted?) version from the lib.

src/Action.hs
Mikolaj: Agreed.

Mikolaj:
@fisx: good, exception is hacky, but that's probably how the machinery is supposed to work; I had no clue how to restructure it to accept the special case cleanly

np:
In a subsequent PR, I want to introduce a class `UpdatePersistM`: Issue #201 .

src/Frontend.hs
```diff
@@ -69,7 +70,7 @@ type AulaTop =
 
 aulaTop :: (GenArbitrary r, PersistM r) => (Action r :~> ExceptT ServantErr IO) -> Server AulaTop
 aulaTop (Nat runAction) =
-       enter runActionForceLogin (catchAulaExcept proxy (aulaMain :<|> aulaTesting))
+       enter (Nat runAction) (catchAulaExcept proxy (aulaMain :<|> aulaTesting))
```
fisx: yeay!

src/Frontend.hs
```diff
@@ -262,7 +258,6 @@ aulaTesting =
   :<|> batchCreateUsers
   :<|> (PageShow <$> Action.persistent mkRandomPassword)
 
-
```
fisx: hm, not yeay.  i'm pretty sure there is a layout rule on that.  (-:

Mikolaj: Sorry, I keep tripping that rule. Yes, there is (it mentions haddock sections, but I guess that's what you mean, since haddocks section markup, with stars, from module export lists is not defined nor recongnized inside module implementation, I think).

src/Persistent/Api.hs
```diff
@@ -353,18 +329,18 @@ instance FromProto Delegation where
     fromProto (ProtoDelegation ctx f t) m = Delegation m ctx f t
 
 -- | So far `mkMetaInfo` is only used by `nextMetaInfo`.
-mkMetaInfo :: User -> Timestamp -> AUID a -> MetaInfo a
-mkMetaInfo user now oid = MetaInfo
+mkMetaInfo :: User -> AUID User -> Timestamp -> AUID a -> MetaInfo a
+mkMetaInfo cUser uid now oid = MetaInfo
```
fisx: Why do you pass `uid` as a second argument and have an explicit contract that `cUser ^. _id == uid` must always hold?  I would like to change that back if it's ok.

src/Frontend.hs
andorp: Git hook? :)

src/Persistent/Api.hs
Mikolaj: Yes, you are right, I have that contract and failed to document it. I separated the argument to make sure `addFirstUser` doesn't diverge, while bootstrapping. In this form it's rather obvious it doesn't, but I'm sure there are ways of expressing that, which don't require the extra argument (and the contract).

src/Frontend.hs
Mikolaj: @andorp: heh, yes, but I'd need to teach it to tell "real" "haddock" section marks from just random long lines of minus signs, etc. :)

fisx: (sorry about the long-lines-with-minusses.  they should never be there.  i started that because there haddocks section headings need to be in the explicit export list if there is one, not in the code itself.  so @Mikolaj you were technicall / legally doing the right thing deleting that empty line.)

fisx:
Done reading, looks great!  (LGTM, that is.)  I will get back to the remaining tasks we discussed in a bit.  (Now for some more distractions!)

Mikolaj:
The redirection commit is much cleaner than I imagined possible. Thank you!

package.yaml
fisx: I tested, sensei still works.  Will leave as is, then.

fisx:
@Mikolaj I think I'm done.  Could you take a quick look at my last commits and complain / WFTAM (wait for travis and merge if green)?

PR #193, opened 2016-03-11
Delegation networks

fisx:
Some work on rendering delegation networks.

I decided against using cytoscape after failing for 30 minutes to get past the errors in the tutorial and the examples I found.  It generally looks a lot more heavy-weight than d3, it appears to be less mature, and I have only worked with d3 before and have some example code to steal from.

New features:

- maintain list of delegations in database
- generate arbitrary delegation networks with plausible data, avatar images etc.
- use d3js for rendering

Missing:

- rest api for delegation data
- filter by delegation context
- show avatar (it's just in the json data so far)
- show name and power of each user
- hide / show individual nodes

PR #195, opened 2016-03-12
Fixes for #193

fisx:
(@Mikolaj you've reviewed #193, and we both forgot: wasn't there a problem with `unsafePerformaIO` inlining before?)


Mikolaj:
Oh, I missed the TODO. Re `unsafePerformIO`, I don't think GHC would every consider it cheap and so inline it, but it doesn't hurt to make sure, given that repeated execution would add user "admin" multiple times, etc. Merging.

PR #200, opened 2016-03-14
2016 03 cc fix fonts

### LINKED ISSUE #187: Missing static resources
Two of the static resources are 404:

2016-03-11-084442_750x789_scrot

local-girl:
This fixes #187
This was due to my symbolic link getting paths to font files wrong. It turns out fontcustom can make scss files so there was no need to make a symbolic link. I have however created a ```fontcustom.yaml``` file right in the root to make things easy and updated the docs, we could do it without and with a really long fontcustom command if you like but this is easier, especially while I am building the site.


fisx:
LGTM, thanks!

PR #207, opened 2016-03-14
Use the session handling from Thentos

np:
This is ready now.


fisx:
Cool!

@np, I know you don't like writing tests, but if you run into issues that have not been caught by our test suite, perhaps you can write *empty* tests that just describe what *should be* tested?  With a `pendingWith "no time to implement test"`?  That would be very helpful.

src/Config.hs
```diff
@@ -9,24 +10,32 @@ import Data.Maybe (fromMaybe)
...
 
 data Config = Config
     { _dbPath :: FilePath
     , _listenerInterface :: String
     , _listenerPort :: Int
     , _htmlStatic :: FilePath
+    , _cfgCsrfSecret :: CsrfSecret
     }
-  deriving (Show, Eq)
+  deriving (Show) -- , Eq)
```
andorp: Please remove if `Eq` is not needed anymore.

src/Frontend/Core.hs
```diff
@@ -346,5 +346,8 @@ redirectFormHandler getPage processor = getH :<|> postH
     renderer page v fa = FormPage page . toHtml . fmap (formPage v fa) <$> makeFrame page
 
 
-redirect :: (MonadError ActionExcept m) => ST -> m a
-redirect = Servant.Missing.redirect
+--redirect :: (MonadError ActionExcept m) => ST -> m a
+--redirect = Servant.Missing.redirect
```
andorp: Please remove it.

andorp:
In general this is big and nice change. I have some question:
On error 500 what kind of page do we get? If an empty one please open an enhancement issue to define its styling, and its redirection behavior. In my opinion Those kind of error pages should contain a button or automatic redirection to the login page.

fisx:
> On error 500 what kind of page do we get?

Good point, but should be dealt with elsewhere.  Not sure if we have an open issue, but we have a `FIXME` in `Frontend`, in section `error handling in servant / wai`.  We probably want to have a `Page500` type anaogous to `Page404`.

fisx:
I've pushed the two changes @andorp suggested.  Will take another look now.

src/Action.hs
```diff
@@ -69,13 +77,35 @@ import Test.QuickCheck (arbitrary, generate)
 
 -- | User representation during an action
 -- FIXME: Figure out which information is needed here.
-data UserState
-    = UserLoggedOut
-    | UserLoggedIn { _username :: UserLogin, _sessionCookie :: ST }
+data UserState = UserState
+    { _usSessionToken :: Maybe ThentosSessionToken
```
fisx: What is this used for?  Since user id is stored here, too, I think we don't need the session token for authentication or authorization.  Does thentos require it for the `GetThentosSessionToken` instance?

src/Action.hs
```diff
@@ -164,75 +221,61 @@ fakeCookieTVar = unsafePerformIO (newTVarIO Nothing)
...
-           let uLogin = case muser of
-                 Nothing -> error "mkRunAction: currently logged in user not in DB"
-                 Just user -> user ^. userLogin
-           return $ UserLoggedIn uLogin "session"
-      put uState
+    checkCurrentUser = do
+        isValid <- userState $ to validUserState
+        unless isValid $ do
+            logout
+            throwError500 "Invalid internal user session state"
```
fisx: Isn't this status code <500?  It can be caused by invalid cookies, right?  (On the other hand, once the cookie has integrity protection, it wouldn't be triggered here.)

src/Action.hs
```diff
@@ -164,75 +221,61 @@ fakeCookieTVar = unsafePerformIO (newTVarIO Nothing)
...
 
 
 ----------------------------------------------------------------------
 -- Action Combinators
 
+-- | Returns the current user ID
+currentUserId :: ActionUserHandler m => m (AUID User)
+currentUserId = userState usUserId >>= \case
+    Nothing -> throwError500 "User is logged out"
```
fisx: This should probably be a redirect rather than a 500.  We don't have to fix it here, but in my eyes it would deserve a `FIXME`.

src/Frontend.hs
```diff
@@ -45,14 +48,24 @@ import qualified Persistent.Implementation.STM
 ----------------------------------------------------------------------
 -- driver
 
-runFrontend :: Config.Config -> IO ()
+extendClearanceOnSessionToken :: Applicative m => ThentosSessionToken -> m ()
+extendClearanceOnSessionToken _ = pure () -- FIXME
```
fisx: `extendClearanceOnUserId` instead?  not sure if i'm missing anything.

tests/FrontendSpec.hs
```diff
@@ -61,15 +60,16 @@ type TestApi =
...
-  :<|> throwError (err303 { errHeaders = ("Location", "/target") : errHeaders err303 })
-  :<|> throwError err500
+  :<|> throwServantErr (err303 { errHeaders = ("Location", "/target") : errHeaders err303 })
+  :<|> throwServantErr err500
 
 testAppAulaExcept :: IO Application
 testAppAulaExcept = do
-    action <- mkRunAction <$> Persistent.Implementation.STM.mkRunPersist
+-- FIXME
+    action <- mkRunAction <$> (ActionEnv <$> Persistent.Implementation.STM.mkRunPersist <*> pure Config.test)
```
fisx: - What is the `FIXME` about?
- I just realized that `Config.test` shouldn't be under `./src`, but under `./test`.  I'm ok with pulling it into this module, but sooner or later we will need a module hierarchy under `./test/Test` for things like this, or modules like `./src/Arbitrary.hs`.  What does everybody think?

fisx:
Cool!

I'm done reading, pushed another nit-pick commit (please kill if it's in your way), and have made a few minor comments.  Please assign to me if you're done with them and I'll merge.

Out of curiousity: do you want to make `LIO` the base of `Action` soon?  (I am slightly leaning towards "that would be a good idea".)

src/Config.hs
```diff
@@ -19,7 +19,7 @@ data Config = Config
     , _htmlStatic :: FilePath
     , _cfgCsrfSecret :: CsrfSecret
     }
-  deriving (Show) -- , Eq)
+  deriving (Show)
```
np: Thanks. The reason was a missing Eq instance for CsrfSecret which would require another round trip on thentos.

src/Frontend/Core.hs
```diff
@@ -346,8 +346,5 @@ redirectFormHandler getPage processor = getH :<|> postH
     renderer page v fa = FormPage page . toHtml . fmap (formPage v fa) <$> makeFrame page
 
 
---redirect :: (MonadError ActionExcept m) => ST -> m a
---redirect = Servant.Missing.redirect
-
```
np: Hum here this was really an implicit FIXME of moving this to Servant.Missing and then one can use this redirect again.

src/Action.hs
np: The session token is used to make CSRF tokens session specific. The ThentosSessionToken is no more than a nonce (in crypto terms).

src/Action.hs
np: Yes we should get the cookie protection first.

src/Action.hs
np: I'm actually wondering if returning a maybe would be better. At some places we end up with "if isLogged then get currentUserId ... else ..." which would be better taken care using a case on maybe. On the other end, situations where you are not logged in are more exceptional than common.

src/Frontend.hs
np: Probably this a consequence of generalizing thentos just enough to be used elsewhere. It is not clear that this is the right hook for aula.

tests/FrontendSpec.hs
np: I would be glad to have a place for reusable test related code. However Config.test is so small that I would not even bother.

np: Oh, and I'll remove this `FIXME`.

np:
About `LIO`, I've made the changes required for things to work and having a `MonadLIO` instance was one of them. I have no set opinion on how to deal with authorization yet. Maybe we should first write with comments the checks we want to enforce and then decide if this can be dealt we a simple firewall kind of code or this would benefit from a more integrated approach such as LIO.

np:
I now recall what is not really finished in this PR (apart from #219).
The last test was disabled because internal errors (code 500) seems to be thrown instead of returned to the client. I shall investigate that.

src/Action.hs
fisx: > I'm actually wondering if returning a maybe would be better. At some places we end up with "if isLogged then get currentUserId ... else ..." which would be better taken care using a case on maybe. On the other end, situations where you are not logged in are more exceptional than common.

why not have both?  `currentUserId` and `currentUserIdMaybe`?

src/Action.hs
```diff
@@ -104,35 +134,52 @@ class (PersistM r, Monad m) => ActionPersist r m | m -> r where
...
     -- | Make the user logged in
     login  :: UserLogin -> m ()
-    -- | Read the actual user state
-    userState :: m UserState
+    -- | Read the current user state
+    userState :: Getting a UserState a -> m a
     -- | Make the user log out
     logout :: m ()
 
 -- | FIXME: every login changes all other logins (replaces the previous one)
```
Mikolaj: Is it still the case?

src/Action.hs
```diff
@@ -144,14 +191,6 @@ instance GenArbitrary r => GenArbitrary (Action r) where
 
 -- * concrete monad type; user state
```
Mikolaj: actually, the 'user state' bit of the section header is outdated for some time already

Mikolaj:
Anyway, great job!

src/Action.hs
fisx: > >  -- * concrete monad type; user state
>
> actually, the 'user state' bit of the section header is outdated for some time already

why?  it's still in `Action`, right?  i think i'm missing something.

Mikolaj: The `UserState` type is defined much higher up. And if the comment means the monad's main purpose is to hold user state, then the wording confused me.

fisx: > The `UserState` type is defined much higher up. And if the comment means the monad's main purpose is to hold user state, then the wording confused me.

ok, i know what you mean now.  the fact that there is a `UserState` somewhere in this type doesn't justify it being mentioned so prominently in the title.  will remove.

PR #173, opened 2016-03-07
Locations, spaces, topics of ideas.

fisx:
This raises some interesting questions on how we currently do stuff.  I think it was a good idea to clean this up!

Fixes #165

PR #209, opened 2016-03-15
Locations, spaces, topics of ideas.

fisx:
This obsoletes #173.

Last commit breaks syntax and should be removed.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -49,6 +49,15 @@ hlint:
 	$(HLINT) --version
 	find src exec tests -name '*.hs' | xargs $(HLINT)
 
+test-everything:
+	perl -i -pe 's/ghc-options: -Wall/ghc-options: -Wall -Werror/' package.yaml
+	hpack
+	cabal install --enable-test
+	make hlint
+	make click-dummies-recreate
+	git checkout -- package.yaml
+	hpack
+
 ghci-no-type-errors:
 	$(EXEC) ghci $(FULL_SOURCES) -fdefer-type-errors
 
--- a/exec/RenderHtml.hs
+++ b/exec/RenderHtml.hs
@@ -118,7 +118,6 @@ instance Arbitrary (ToHtmlSpecial ViewIdea) where
                 <*> arb
                 <*> arb
                 <*> arb
-                <*> pure Nothing
                 <*> (Set.fromList <$> vectorOf 5 arb)  -- comments
                 <*> (Set.fromList <$> vectorOf 5 arb)  -- likes
                 <*> arb
--- a/src/Arbitrary.hs
+++ b/src/Arbitrary.hs
@@ -99,7 +99,7 @@ instance Arbitrary ViewIdea where
     arbitrary = ViewIdea <$> arb <*> arb
 
 instance Arbitrary CreateIdea where
-    arbitrary = CreateIdea <$> arb <*> arb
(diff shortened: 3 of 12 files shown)


fisx:
Great lens mini-tutorial, thanks @np!  I cleaned up the commit history and your authorship info got lost (recovered it in the commit message).

Two test cases still fail.  My first guess is that there are now harder consistency requirements due to the idea space lookup via topic id, and the crash is caused by dangling topic ids.  Will look into it.

fisx:
### Comments by @np recovered from earlier commits (rebase is dangerous!).

```
data IdeaLocation =
      IdeaLocationSpace { _ideaLocationSpace :: IdeaSpace }
    | IdeaLocationTopic { _ideaLocationSpace :: IdeaSpace, _ideaLocationTopicId :: AUID Topic }

makeLenses ''IdeaLocation
```

Following the docs of https://hackage.haskell.org/package/lens-4.13/docs/Control-Lens-TH.html#v:makeLenses we get a lens to access the idea space and only a traversal to access the topic id.

```
ideaLocationSpace :: Lens' IdeaLocation IdeaSpace
ideaLocationTopicId :: Traversal' IdeaLocation (AUID Topic)
```

I'm making a commit with that.

My guess is that this type is impossible to fulfill. `ideaLocationTopicId` cannot be:

* a getter, since there might be no `AUID Topic`
* a lens, since it cannot be a getter
* a prism, since we cannot reconstruct the `IdeaSpace` if we start only with an `AUID Topic`

What we have from my previous comment is a `Traversal' IdeaLocation (AUID Topic)`.
Looking at the main diagram https://hackage.haskell.org/package/lens-4.13 a traversal is weaker than a lens or a prism. A traversal can be turned into a setter or a fold. Assume `t` to be a
+`Traversal' s a`:

* `s ^.. t` get all the elements as a list: `[a]`
* `s ^? t` get the first element: `Maybe a`
* `s & t .~ a` set all the elements to `a`.
* `s & t %~ f` modify all the elements with `f`.

Then `ideaTopicId` is the composition of `ideaLocation` (a lens) and `ideaLocationTopicId` (a traversal). However guessing the correct type signature can be difficult. My advice is to omit the type
+signatures at first. Since a traversal is weaker than a lens the resulting composite is a traversal.

```
ideaTopicId :: Traversal' Idea (AUID Topic)
ideaTopicId = ideaLocation . ideaLocationTopicId
```

What you were getting at with "is it somehow ok to yield a maybe here?" is that we know that there is either no topic id or a single topic id. The traversal is weaker it can be any number of them.
+A custom lens can be written as:

```
ideaLocationMaybeTopicId :: Lens' IdeaLocation (Maybe (AUID Topic))
ideaLocationMaybeTopicId f = \case
    IdeaLocationSpace spc     -> mk spc <$> f Nothing
    IdeaLocationTopic spc tid -> mk spc <$> f (Just tid)
  where
    mk spc = \case
        Nothing  -> IdeaLocationSpace spc
        Just tid -> IdeaLocationTopic spc tid
```

The composite lens is then:

```
ideaMaybeTopicId :: Lens' Idea (Maybe (AUID Topic))
ideaMaybeTopicId = ideaLocation . ideaLocationMaybeTopicId
```

np:
LGTM but travis disagree.

Mikolaj:
LGTM.

PR #210, opened 2016-03-15
Do not use buttons in forms for anything but submit!

fisx:


andorp:
... inside a `<form>` tag.

andorp:
LGTM

Issue #201, opened 2016-03-14
Add a class `UpdatePersistM` capturing the current user


np:

fisx:
:+1:

np:
I propose to merge the current user with the prototype in #239

fisx:
Fixed in #239.

PR #211, opened 2016-03-15
WIP UpdatePersistM

np:
Attempt at solving #201.
To be continued after merging #207.

### DIFF
--- a/src/Persistent/Api.hs
+++ b/src/Persistent/Api.hs
@@ -10,6 +10,7 @@
 
 module Persistent.Api
     ( PersistM
+    , UpdatePersistM
     , AMap
     , AulaLens
     , AulaGetter
@@ -21,6 +22,7 @@ module Persistent.Api
     , getDb
     , addDb
     , modifyDb
+    , currentUser
     , findIn
     , findInBy
     , findInById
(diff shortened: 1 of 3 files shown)


src/Persistent/Api.hs
```diff
@@ -126,12 +128,18 @@ emptyAulaData = AulaData nil nil nil nil nil 21 21 30 3 0
 -- FIXME move enough specialized calls to IO in PersistM to remove MonadIO
 class MonadIO m => PersistM m where
     getDb :: AulaGetter a -> m a
+
+class PersistM m => UpdatePersistM m where
     modifyDb :: AulaSetter a -> (a -> a) -> m ()
```
andorp: In my opinion this could be `modifyDb :: User -> AulaSetter a -> (a -> a) -> m ()`

src/Persistent/Api.hs
```diff
@@ -126,12 +128,18 @@ emptyAulaData = AulaData nil nil nil nil nil 21 21 30 3 0
...
+    currentUser :: m User
```
andorp: And we could delete this combinator.

fisx: > And we could delete this combinator.

a reason for doing it *with* the combinator may be that it can be hidden in a reader in the implementation of 'Action'.

andorp: Ok, no strong opinion here. Probably I've just overlooked something.

PR #213, opened 2016-03-15
2016 03 cc create idea detail styles

local-girl:
Made view idea look nicer also styled comments but it needs a form.
I doubt I have in every phase combo here but there's a lot.


fisx:
LGTM.  I will merge this and work on some fixups in a separate PR.

PR #215, opened 2016-03-15
Fix html

fisx:


src/Frontend/Page/Idea.hs
```diff
@@ -68,7 +68,11 @@ instance ToHtml ViewIdea where
     -- The result might be that too much information is displayed.
     toHtml p@(ViewIdea idea phase) = semanticDiv p $ do
         header_ [class_ "detail-header"] $ do
-            a_ [class_ "btn m-back detail-header-back", href_ U.Broken] "Zum Thema"
+            a_ [ class_ "btn m-back detail-header-back"
```
Mikolaj: Unrelated to this change, but why is it 'btn', but `a_` instead of not  `button_`?

Mikolaj:
LGTM.

andorp:
Apart from the @Mikolaj 's question, LGTM.

src/Frontend/Page/Idea.hs
fisx: > Unrelated to this change, but why is it 'btn', but `a_` instead of not  `button_`?

My guess is that the css class 'btn' makes it *look* like a button, whereas the anchor element `a_` is better because it doesn't have certain undesired effects.  For instance, there can only be one `button_` per form, and its mechanics are very specific.

Mikolaj: Fair enough. There is a couple more such occurrences, so that's probably not a typo:
```
src/Frontend/Page/Admin.hs:                    a_ [href_ U.Broken, class_ "btn-cta"] "Passwort zurücksetzen" >> br_ []
src/Frontend/Page/Admin.hs:                    a_ [href_ U.Broken, class_ "btn-cta"] "Nutzer löschen" >> br_ []
```

PR #218, opened 2016-03-16
2016 03 cc fix js menu

local-girl:
This fixes 2 pop menus on a page conflicting.
As a bonus I moved icon SVGs so static-src as they are not needed on production


fisx:
LGTM, thanks!

PR #220, opened 2016-03-16
Misc cleanup.

fisx:


fisx:
There is still one TODO in the code.  Otherwise good to merge.

exec/RenderHtml.hs
```diff
@@ -88,12 +87,9 @@ instance (ToHtml p) => ToHtml' (ToHtmlDefault p) where
     toHtml' (ToHtmlDefault p) = toHtml p
 
 instance (FormPageView p) => ToHtml' (ToHtmlForm p) where
-    toHtml' (ToHtmlForm p) = unwrap2 $ do
-        v <- unwrap1 $ getForm "" (makeForm p)
+    toHtml' (ToHtmlForm p) = toHtml $ do
+        v <- return . runIdentity $ getForm "" (makeForm p)
```
np: why not `let v = getForm ...` ?

src/Frontend/Core.hs
```diff
@@ -113,13 +113,24 @@ class FormPageView p where
...
 -- | Wrap anything that has 'ToHtml' and wrap it in an HTML body with complete page.
 data Frame body = Frame User body | PublicFrame body
   deriving (Functor)
 
--- FIXME: Replace it with enum type that represents the extra headers.
-data Frame' body = Frame' User (Html ()) body | PublicFrame' (Html ()) body
-  deriving (Functor)
+frameAlgebra :: (body -> a) -> Frame body -> a
+frameAlgebra f = \case (Frame _ b)     -> f b
+                       (PublicFrame b) -> f b
```
np: Oh my! This could use the getter of frameBody lens if we were to derive it. See my next commit.

Mikolaj:
Thank you for making the section headings uniform. Was the sole `-----` line from Persistent/Api.hs cleanup up, too?

fisx:
@Mikolaj yes, I decided that we don't want `---------------` in the code at all.  Also, this module needs some section headings and shuffling, but I didn't want to do that with so many people around affected by the dust.

exec/RenderHtml.hs
fisx: you mean with a `runIdentity`, but without `return`?  you're right.  will you change it or should I?

np:
Added 2 more commits. LGTM

fisx:
Thanks!

PR #225, opened 2016-03-16
Use Login path in redirect.

andorp:


Mikolaj:
A good catch. Merging.

PR #226, opened 2016-03-16
Misc cleanup.

### LINKED ISSUE #224: public pages show private frame; login page does not load frame and has no css.

fisx:
Joint work with @andorp.

Fixes #224.


src/Frontend/Core.hs
```diff
@@ -368,14 +366,13 @@ redirectFormHandler getPage processor = getH :<|> postH
...
             Just payload -> processor2 page payload >>= redirect
-            Nothing      -> renderer page v fa)
+            Nothing      -> FormPage v fa <$> makeFrame page)
             `finally` cleanupTempCsvFiles formData
 
     -- (possibly interesting: on ghc-7.10.3, inlining `processor1` in the `postForm` call above
     -- produces a type error.  is this a ghc bug, or a bug in our code?)
     processor1 = makeForm
     processor2 page result = processor result $> absoluteUriPath (redirectOf page)
-    renderer page v fa = FormPage page . formPage v fa <$> makeFrame page
```
andorp: Kill the evil renderer! \o/

src/Frontend.hs
```diff
@@ -141,7 +141,7 @@ aulaMain =
   :<|> pure (Frame frameUserHack PageStaticTermsOfUse) -- FIXME: Generate header with menu when the user is logged in.
 
   :<|> Page.login
-  :<|> Page.logout
+  :<|> redirect "/login"
```
Mikolaj: How applying the fix from #225 here? Also in another place.

np:
Only had a quick look but LGTM.

fisx:
argh, travis!  will push another fix in a moment.

src/Frontend.hs
```diff
@@ -152,7 +151,7 @@ aulaMain =
   :<|> pure (Frame frameUserHack PageStaticTermsOfUse) -- FIXME: Generate header with menu when the user is logged in.
 
   :<|> Page.login
-  :<|> Page.logout
+  :<|> (redirect . absolutePath . relPath $ P.Login)
```
Mikolaj: While we are being pedantic, there is one more redirect in this file with a string literal path...

Mikolaj:
Other than the 2nd redirect, LGTM.

src/Frontend/Core.hs
```diff
@@ -144,7 +144,7 @@ makeFrame :: (ActionPersist r m, ActionUserHandler m, MonadError ActionExcept m,
 makeFrame p = do
   isli <- isLoggedIn
   if | not isli && isPrivatePage p -> redirect "/login"
-     | isPrivatePage p             -> flip Frame p <$> currentUser
+     | isli     || isPrivatePage p -> flip Frame p <$> currentUser
```
np: This is exactly the place where `currentMaybeUser` and a case would be better. (no need to did yet though)

np:
LGTM again

src/Frontend.hs
fisx: > While we are being pedantic, there is one more redirect in this file with a string literal path...

thanks, will fix in #229.

PR #227, opened 2016-03-16
2016 03 cc more admin styling

local-girl:
Styled the edit user page - not exactly sure what the camera thing is supposed to do, might break when you code it up - super easy to fix


np:
LGTM

PR #236, opened 2016-03-17
PersistM no longer implies MonadIO

np:


src/Persistent/Api.hs
```diff
@@ -128,11 +129,20 @@ dbTopics = dbTopicMap . to Map.elems
...
--- FIXME move enough specialized calls to IO in PersistM to remove MonadIO
-class MonadIO m => PersistM m where
+class Monad m => PersistM m where
     getDb :: AulaGetter a -> m a
     modifyDb :: AulaSetter a -> (a -> a) -> m ()
 
+    getCurrentTimestamp :: m Timestamp
+
+    default getCurrentTimestamp :: MonadIO m => m Timestamp
+    getCurrentTimestamp = Timestamp <$> liftIO getCurrentTime
```
Mikolaj: Perfect. This mention of MonadIO doesn't leak any IO power outside, but documents that we mean the ordinary IO stuff.

Mikolaj:
LGTM. That's the way to go with any monad, IMHO. Abstract over the monad's implementation and then explicitly limit and hide access to MonadIO behind the abstraction.

PR #239, opened 2016-03-17
Refactoring: package the current user with the prototype upon addDb

np:
This replaces #211. Some UpdatePersistM class is still a good idea but not to solve the current user tracking.

### DIFF
--- a/src/Frontend/Page/Topic.hs
+++ b/src/Frontend/Page/Topic.hs
@@ -20,7 +20,7 @@ module Frontend.Page.Topic
     , moveIdeasToTopic )
 where
 
-import Action (ActionM, ActionPersist(..), ActionUserHandler, ActionExcept, currentUser)
+import Action (ActionM, ActionPersist(..), ActionUserHandler, ActionExcept, currentUserAddDb)
 import Frontend.Prelude hiding (moveIdeasToLocation)
 
 import qualified Persistent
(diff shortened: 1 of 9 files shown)


src/Frontend/Page/Topic.hs
```diff
@@ -197,10 +197,9 @@ viewTopic _space tab topicId = makeFrame =<< persistent (do
     Just topic <- findTopic topicId
     ViewTopicIdeas tab topic <$> (findIdeasByTopic topic >>= mapM getNumVotersForIdea))
 
-createTopic :: (ActionM r action) => IdeaSpace -> [AUID Idea] -> ServerT (FormHandler CreateTopic) action
+createTopic :: ActionM r m => IdeaSpace -> [AUID Idea] -> ServerT (FormHandler CreateTopic) m
 createTopic space ideas =
-  redirectFormHandler (pure $ CreateTopic space ideas)
-    (\protoTopic -> currentUser >>= persistent . flip addTopic protoTopic)
+  redirectFormHandler (pure $ CreateTopic space ideas) (currentUserAddDb addTopic)
```
Mikolaj: Nice.

Mikolaj:
The travis problem (after I restarted a build stalled on github pull) is just spurious parens by hlint.

src/Frontend/Page/Topic.hs
fisx: agreed!  (-:

fisx:
LGTM.

Not sure about `UserWithProto a`, since it only works in some places and has exactly the same number of characters as `(User, Proto a)`.  Just saying because I can't *not* find anything.  (-:

Thanks!  Will merge if green.

PR #241, opened 2016-03-17
2016 06 cc view idea

local-girl:
Styled viewcontent page - also constrained width of content header and content here, might have weird effects but very easy to revert.

I couldn't quite work around all the switches here over to you @fisx !

Also the tick and cross are a bit bold compared to the design I made a note to get skinny ones (if that even matters)


fisx:
LGTM.  made a few small changes and will merge when green.

PR #243, opened 2016-03-18
Makefile: tweak test-everything.

fisx:
- do not touch package.yaml, aula.cabal.
- run ghc with -O0.
- run test suite, haddock.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -50,12 +50,11 @@ hlint:
 	find src exec tests -name '*.hs' | xargs $(HLINT)
 
 test-everything:
-	perl -i -pe 's/ghc-options: -Wall/ghc-options: -Wall -Werror/' package.yaml
-	hpack
-	cabal install --enable-test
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -50,12 +50,11 @@ hlint:
 	find src exec tests -name '*.hs' | xargs $(HLINT)
 
 test-everything:
-	perl -i -pe 's/ghc-options: -Wall/ghc-options: -Wall -Werror/' package.yaml
-	hpack
-	cabal install --enable-test
+	cabal install --enable-test --ghc-options="-Werror -Wall -O0"
+	cabal test
+	cabal haddock
```
andorp: Do we want to do build haddock with stack too?

fisx: Possibly?  Different PR, though.  I also want to have explicit export lists in all modules, and section headings only in the export lists.  This line here is just to make sure the syntax of the comments does not deteriorate too much.

andorp: Ok

PR #245, opened 2016-03-18
Edit Topic Page

andorp:
Edit Topic page now contains the title and the description apart from the ideas.

### DIFF
--- a/src/Frontend/Page/Topic.hs
+++ b/src/Frontend/Page/Topic.hs
@@ -146,7 +146,7 @@ instance FormPage CreateTopic where
         <*> ("desc"  .: (Markdown <$> DF.text Nothing))
         <*> ("image" .: DF.text nil)
         <*> pure space
-        <*> selectedIdeaParameters ideas
+        <*> makeFormIdeaSelection ideas
 
     formPage v fa p@(CreateTopic _space ideas) =
         semanticDiv p $ do
@@ -156,7 +156,7 @@ instance FormPage CreateTopic where
                 DF.inputTextArea Nothing Nothing "desc" v >> br_ []
                 -- FIXME: There is no image in the click dummy.
                 DF.inputText     "image" v >> br_ []
-                ideaCheckboxes v ideas
+                formPageIdeaSelection v ideas
                 DF.inputSubmit   "Add Topic"
 
 -- Edit topic description and add ideas to topic.
@@ -176,7 +176,7 @@ instance FormPage EditTopic where
         TopicFormPayload
         <$> ("title" .: DF.text (Just (topic ^. topicTitle)))
         <*> ("desc"  .: (Markdown <$> DF.text (Just $ fromMarkdown (topic ^. topicDesc))))
-        <*> selectedIdeaParameters ideas
+        <*> makeFormIdeaSelection ideas
 
     formPage v fa p@(EditTopic _space _topic ideas) = do
         semanticDiv p $ do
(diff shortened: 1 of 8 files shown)


fisx:
The dangling links are very strange.  How did they make it to master?  And why didn't I have them?

fisx:
The click dummy is a bit outdated here.  What we want is something like this:

1. two pages `CreateTopic` and `EditTopic`.
1. `CreateTopic` does everything it already does, plus what page `MoveIdeasToTopic` does.
1. `EditTopic` has the same structure and functionality, except it works on ideas that already exist.
1. `MoveIdeasToTopic` become unnecessary.

What you called `EditTopic` is not a page, and I would like to call it `TopicFormPayload` unless you can find a better name.

@andorp: do you think it makes sense to do all that in this PR?  It doesn't strike me as a big task, and it would be nice to have it out of our way.  Let me know what you think.

fisx:
(Other than that, this is probably good work.  Sorry about the poor task specification!)

andorp:
I am going to apply the proposed changes. Thanks.

fisx:
I've checked the dangling-links commit again, and those links don't dangle in my working copy.  @andorp are your submodules up to date?

fisx:
re. dangling links: see also #255.

andorp:
My bad, updating submodules...

fisx:
I opened the related #256 (can be fixed elsewhere).

src/Frontend/Page/Topic.hs
```diff
@@ -158,44 +146,64 @@ instance FormPage CreateTopic where
...
 ideaToFormField idea = "idea-" <> cs (show $ idea ^. _Id)
 
+ideaCheckboxes :: (Monad m) => View (HtmlT m ()) -> [Idea] -> HtmlT m ()
+ideaCheckboxes v ideas =
+    ul_ . for_ ideas $ \idea ->
+        li_ $ do
+            DF.inputCheckbox (ideaToFormField idea) v
+            idea ^. ideaTitle . html
+
+selectedIdeaParameters :: forall m v . (Monad m, Monoid v)
```
fisx: Can you document what this does?  Find a better name?  (Discuss off-line?)

fisx: `makeFormIdeaSelection`?

src/Frontend/Page/Topic.hs
```diff
@@ -158,44 +146,64 @@ instance FormPage CreateTopic where
...
+                DF.inputTextArea Nothing Nothing "desc" v >> br_ []
+                ideaCheckboxes v ideas
                 DF.inputSubmit "Speichern"
                 button_ "Abbrechen" -- FIXME
 
 ideaToFormField :: Idea -> ST
 ideaToFormField idea = "idea-" <> cs (show $ idea ^. _Id)
 
+ideaCheckboxes :: (Monad m) => View (HtmlT m ()) -> [Idea] -> HtmlT m ()
+ideaCheckboxes v ideas =
```
fisx: `formPageIdeaSelection`?

src/Frontend/Page/Topic.hs
fisx: I think this may be a good application for nested forms.  But we don't have to investigate this idea here, just saying.

tests/Frontend/CoreSpec.hs
```diff
@@ -277,7 +282,7 @@ instance ArbFormPageResult PageHomeWithLoginPrompt where
 instance ArbFormPageResult CreateTopic where
     arbFormPageResult (CreateTopic space ideas) =
             set protoTopicIdeaSpace space
-          . set protoTopicIdeas ideas
+          . set protoTopicIdeas (map (^. _Id) ideas)
```
fisx: `(view _Id <$> ideas)`?  I have no strong opinion, but we seem to prefer `view` over operator sections.

fisx:
Made a few comments, otherwise LGTM.

fisx:
Thanks!

PR #247, opened 2016-03-18
Page CreateIdea category selection

fisx:
Obsoletes #242.


fisx:
This fixes a few issues with `custom.js`.  The issue it does not fix is this: the click handlers for the buttons triggers successfully the state change for the radios, but once the handler returns, somebody else comes along and resets the state of the radios.

Neither @fisx nor @andorp have any idea who that might be so far.  I think we need someone who knows js...

static/js/custom.js
```diff
@@ -70,4 +69,8 @@ function removeClass(el, cl) {
 
 function addClass(el, cl) {
     if(el) el.className = el.className + " " + cl;
-}
\ No newline at end of file
+}
+
+function endsWith(str, suffix) {
```
andorp: Please remove if it not used.

fisx:
@andorp thanks, removed dead code.

@local-girl could you take a look?  thanks!

fisx:
Fixed and tested!  I commented out one test because I have other more urgent tasks (last commit).  If the reviewer has an idea how to fix it, I'd be interested.  Otherwise we can fix it later.

andorp:
LGTM

PR #252, opened 2016-03-20
Fix: do not hide popup menu behind other contents.

### LINKED ISSUE #249: The popup menu in the frame (upper right) does not open.
Somewhat related, I propose the following changes:

1. merge #247 
1. move all script tags into html header and wrap the entire contents of `custom.js` in an onload handler.
1. do not use the `p in pops` pattern, but use `for` loops over array length instead, like elsewhere in #247.
1. end `custom.js` with a newline to make `git diff` shut up about that.

none of this is likely to fix the issue, though.

fisx:
Fixes #249


fisx:
@local-girl I did that cargo-cult style, but it seems to work.  Could you take a look and fix if necessary?

Thanks!

PR #253, opened 2016-03-20
Handle login errors

### LINKED ISSUE #196: When user does not exist, logging in as that user should display something and redirect somewhere
Currently there is internal server error.

fisx:
Fixes #196.

### DIFF
--- a/src/Action/Implementation.hs
+++ b/src/Action/Implementation.hs
@@ -67,7 +67,7 @@ instance PersistM r => ActionUserHandler (Action r) where
         muser <- persistent $ findUserByLogin uLogin
         case muser of
             Nothing ->
-                redirect . absoluteUriPath . relPath $ U.Login
+                redirect . absoluteUriPath . relPath . U.Login $ Just False
             Just user -> do
                 usUserId .= Just (user ^. _Id)
                 sessionToken <- freshSessionToken
--- a/src/Action/Implementation.hs
+++ b/src/Action/Implementation.hs
@@ -20,14 +20,18 @@ import Control.Monad.Trans.Except (ExceptT(..), runExceptT, withExceptT)
 import Prelude
 import Servant
 import Servant.Missing
+import Thentos.Action (freshSessionToken)
+import Thentos.Prelude (DCLabel, MonadLIO(..), MonadRandom(..), evalLIO, LIOState(..), dcBottom)
 
 import qualified Data.ByteString.Lazy as LBS
 
 import Action
+import Data.UriPath
 import Persistent
 import Types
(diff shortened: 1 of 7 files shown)


fisx:
[irc]

```
21:57 < andorp> I usually don't like `Maybe Bool` like types :)
21:57 < fisx> i expected that feedback.  (:
21:57 < fisx> sigh.
21:57 < fisx> it has to be `Maybe` because of how `QueryParam` works.
21:58 < andorp> but it introduces mental overhead :)
21:58 < fisx> the `Bool` could indeed be an aula type, but that would make for a lot more code.
21:58 < andorp> 3 different case, with an implicit interpreter in your mind :)
21:58 < fisx> we could do `Maybe ()` instead.
21:58 < fisx> i thought that was stupid.
21:59 < andorp> That is even worse :)
21:59 < fisx> why?
21:59 < fisx> it's isomorphic to `Bool`.
21:59 < andorp> Nothing = False, Just () = True
21:59 < fisx> right.
21:59 < andorp> False => Invalid Password, True => ???
21:59 < fisx> and then there is `data LoginFailure = LoginFailure`.
22:00 < andorp> let's discuss that tomorrow
22:00 < fisx> then `.. :> QueryParam "login_status" LoginFailure :> ...
22:00 < andorp> I dislike the idea passing this as a query parameter
22:01 < fisx> do you have a better idea?
22:01 < andorp> as an attacker my first attack would be to set it True
22:01 < andorp> Kinda
22:01 < andorp> but I don't want to think through at the moment :)
```

Mikolaj:
If the type had 3 aptly named cases, it would be much more easy for me to review. Other than that, LGTM (in particular, the described attack just makes the app fail to display a warning, nothing more).

np:
Could we have a list of messages in the session state to display to the user? A failed login would push such a message and trigger a redirect. Something closer to the current attempt would be to have a different route such as /login/retry.

fisx:
@Mikolaj the type has to match `Maybe a` because the query param may be missing, and that's how servant wants to handle this.  (Of course, we could define our own combinator, but I don't think that's the place where we should do that.)

I just talked to @andorp, here is the summary:

1. query params are meant to shape the data contained in an http response, not for fancy protocol data.
2. even though it seems safe, it's weird to give the adversary write access to system messages.
3. can we use form vaildation for this?

@np i like the idea of adding a message queue to the cookie.  we have code for this in thentos that we can use / recycle.  but if we could do it with form validation in a more implicit way, it may be even better.

fisx:
merging this and deferringg remaining tasks to #258.

PR #259, opened 2016-03-21
Add user details to login page (needed for demo system).

### LINKED ISSUE #203: Add demo server instructions to login page
we could mention helpful details like:

- if you login as admin with any password, you can play.
- if you login as any other user, the server will crash a little.

this list could be extended with other users that we add to allow for testing different roles.

@rittermo, this is against the click dummy, but it might help.  what do you think?

fisx:
Fixes #203


fisx:
2016-03-21-120516_1680x1050_scrot

Mikolaj:
LGTM (didn't check the German though)

PR #273, opened 2016-03-22
Demo data improvements.

andorp:


src/Persistent/Api.hs
```diff
@@ -273,6 +274,14 @@ addLikeToIdea cUser iid = do
     metainfo <- nextMetaInfo cUser
     modifyIdea iid (ideaLikes %~ Set.insert (IdeaLike metainfo))
 
+-- FIXME: Save comments properly.
+addCommentToIdea :: User -> AUID Idea -> Document -> PersistM m => m Comment
```
fisx: We could wrap the idea id and the document into a proto type, but for now let's keep it simple.

fisx:
LGTM.  Thanks!

Issue #256, opened 2016-03-21
RedirectOf for topics and ideas.


fisx:
- after creating or editing a topic, you should be redirected to view topic, not topic list.
- after creating or editing an idea, you should be redirected to view idea, not the location of the idea.

(as discussed with @rittermo last thursrday.)

andorp:
To be able to handle this functionality we have to change the `FormPage` class.

Before:
````haskell
-- | Render Form based Views
class Page p => FormPage p where
    type FormPageResult p :: *
    -- | The form action used in form generation
    formAction :: p -> UriPath
    -- | Calculates a redirect address from the given page
    redirectOf :: p -> UriPath
    -- | Generates a Html view from the given page
    makeForm :: (Monad m) => p -> DF.Form (Html ()) m (FormPageResult p)
    -- | Generates a Html snippet from the given view, form action, and the @p@ page
    formPage :: (Monad m) => View (HtmlT m ()) -> ST -> p -> HtmlT m ()
````

After:
````haskell
-- | Render Form based Views
class Page p => FormPage p where
    -- | Information which parsed from the form
    type FormPageResult p :: *
   -- | Information created while processing the form data
    type FormPageInnerResult p :: *
    -- | The form action used in form generation
    formAction :: p -> UriPath
    -- | Calculates a redirect address from the given page
    redirectOf :: p -> FormPageResult p -> UriPath
    -- | Generates a Html view from the given page
    makeForm :: (Monad m) => p -> DF.Form (Html ()) m (FormPageResult p)
    -- | Generates a Html snippet from the given view, form action, and the @p@ page
    formPage :: (Monad m) => View (HtmlT m ()) -> ST -> p -> HtmlT m ()
````

Also it introduces changes in `redirectFormHandler` propagating the information to the new `redirectOf` function.

PR #282, opened 2016-03-22
Redirect based on form result.

andorp:
Fixes: #256


src/Frontend/Page/Admin.hs
```diff
@@ -222,7 +222,7 @@ instance FormPage PageAdminSettingsDurations where
     formAction _ = relPath $ U.Admin U.AdminDuration
 
     -- FIXME: Do we redirect to the same page???
-    redirectOf _ = relPath $ U.Admin U.AdminDuration
+    redirectOf _ _ = relPath $ U.Admin U.AdminDuration
```
fisx: When matching `()`, I prefer the more specific pattern `()` over `_`.  Makes it more readable and gives you more guidance when the type changes and you want to react to it.

This has some implications re. strictness, but I don't think that matters here.

Just a comment, also I'm curious what you all think. let's leave it as is for now.

fisx:
LGTM.

PR #287, opened 2016-03-23
Styling edit idea page

local-girl:
As discussed this could easily be merged with the create idea form, over to you!


fisx:
LGTM.  Added some FIXMEs, but those can be dealt with in a separate PR.

Thanks!

PR #288, opened 2016-03-23
Style edit topic page

local-girl:
Again, this form is the same as as create topic

### DIFF
--- a/src/Frontend/Page/Topic.hs
+++ b/src/Frontend/Page/Topic.hs
@@ -167,21 +167,21 @@ instance FormPage CreateTopic where
 
 createOrEditTopic :: Monad m => View (HtmlT m ()) -> [Idea] -> HtmlT m ()
 createOrEditTopic v ideas = do
-                        label_ $ do
-                            span_ [class_ "label-text"] "Wie soll der Titel des Themas lauten?"
-                            inputText_ [class_ "m-small", placeholder_ "z.B. Computerraum"]
-                                "title" v
-                        label_ $ do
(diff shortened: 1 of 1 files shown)


fisx:
LGTM.  I shuffled the code around a bit.  Ready to merge IMO.

src/Frontend/Page/Topic.hs
```diff
@@ -162,21 +162,26 @@ instance FormPage CreateTopic where
...
+createOrEditTopic :: Monad m => View (HtmlT m ()) -> [Idea] -> HtmlT m ()
+createOrEditTopic v ideas = do
+    label_ $ do
+        span_ [class_ "label-text"] "Wie soll der Titel des Themas lauten?"
+        inputText_ [class_ "m-small", placeholder_ "z.B. Computerraum"]
+            "title" v
+    label_ $ do
+        span_ [class_ "label-text"] "Beschreiben Sie das Thema"
+    -- FIXME I want a placeholder here too
+        DF.inputTextArea Nothing Nothing "desc" v
```
andorp: On Edit topic the exiting text of the topic should be rendered as the content in the  `inputTextArea`.

fisx: > On Edit topic the exiting text of the topic should be rendered as the content in the  `inputTextArea`.

yes, but in the 'create' mode, it doesn't.  and we don't want to have it as the default value either.  there is a 'placeholder_' attribute, but it's not supported by DF.inputTextArea, so we need to write another wrapper.

andorp: @fisx show me offline, that works as expected. Sorry for the noise.

PR #295, opened 2016-03-23
Lookup the user upon login as part of form validation

np:
* FormPage.makeForm: the monad now has the ActionM constraint
* ActionUserHandler.login: now takes a User instead of UserLogin
* New module Action.Dummy: a tiny monad stack which errors out most of the time
* The FormPagePayload of the login page is now User, the DB lookup of the
  login happens as part of validation
* As we cannot reconstruct the password the corresponding test has been disabled

### DIFF
--- a/aula.cabal
+++ b/aula.cabal
@@ -70,6 +70,7 @@ library
     , vector ==0.10.12.3
   exposed-modules:
       Action
+      Action.Dummy
       Action.Implementation
       Arbitrary
       Config
--- a/src/Action/Dummy.hs
+++ b/src/Action/Dummy.hs
@@ -1,0 +1,53 @@
+{-# LANGUAGE GeneralizedNewtypeDeriving #-}
+{-# LANGUAGE MultiParamTypeClasses      #-}
+{-# LANGUAGE OverloadedStrings     #-}
+module Action.Dummy
+    ( DummyT(DummyT,unDummyT), Dummy
+    , runDummyT, runDummy
+    , notImplemented
+    ) where
+
+import Control.Monad.Except (runExceptT)
+
+import Action
+import Frontend.Prelude
+
+newtype DummyT m a = DummyT { unDummyT :: ExceptT ActionExcept m a }
+    deriving (Functor, Applicative, Monad, MonadError ActionExcept)
+
+type Dummy = DummyT Identity
+
+runDummyT :: DummyT m a -> m (Either ActionExcept a)
+runDummyT = runExceptT . unDummyT
+
+runDummy :: Dummy a -> Either ActionExcept a
+runDummy = runIdentity . runDummyT
+
+notImplemented :: Monad m => String -> String -> DummyT m a
+notImplemented meth _class = throwError500 $ unlines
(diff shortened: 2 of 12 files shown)


fisx:
There is actually a way to reconstruct the password of a user: When students are created, their password is set to 'initial'.  This means that it is stored in plaintext, can be extracted into Csv downloads by admins, and forces the student to change it to something scrypted the first time they log in.

We can use the initial password type here to test login.

aula.cabal
```diff
@@ -70,6 +70,7 @@ library
     , vector ==0.10.12.3
   exposed-modules:
       Action
+      Action.Dummy
```
fisx: Under `Persistent`, we have the module structure `Action.Implementation.*`, where `*` is any implementation.  I guess the equivalent here would be:

- `Action.Implementation` -> `Action.Implementation.Basic`
- `Action.Dummy` -> `Action.Implementation.Dummy`

src/Action/Dummy.hs
```diff
@@ -0,0 +1,53 @@
...
+type Dummy = DummyT Identity
+
+runDummyT :: DummyT m a -> m (Either ActionExcept a)
+runDummyT = runExceptT . unDummyT
+
+runDummy :: Dummy a -> Either ActionExcept a
+runDummy = runIdentity . runDummyT
+
+notImplemented :: Monad m => String -> String -> DummyT m a
+notImplemented meth _class = throwError500 $ unlines
```
fisx: Leading `_` means "ghc: don't wanr about unused bind".  We've done this before, and lens does it, but I'd prefer if we didn't.  `cl`?

fisx:
I am beginning to really like the DF stuff that is happening here.

Happy with this PR, with a few minor comments.  If you have processed them all and have no objections, you can merge this.

Thanks!

src/Frontend/Core.hs
```diff
@@ -123,7 +123,7 @@ class Page p => FormPage p where
     -- | Calculates a redirect address from the given page
     redirectOf :: p -> FormPageResult p -> UriPath
     -- | Generates a Html view from the given page
-    makeForm :: (Monad m) => p -> DF.Form (Html ()) m (FormPagePayload p)
+    makeForm :: (Monad m, ActionM r m) => p -> DF.Form (Html ()) m (FormPagePayload p)
```
andorp: The `Monad m` can be removed from the constraints.

np: Nice catch.

src/Action/Dummy.hs
np: Granted.

aula.cabal
np: I'm not opposed to it.

np:
@fisx I thought about that trick to get the password back but I'm not sure it would work. Here we are testing with arbitrary data as input, thus the password does not need to match the one of the database.

fisx:
> @fisx I thought about that trick to get the password back but I'm not sure it would work. Here we are testing with arbitrary data as input, thus the password does not need to match the one of the database.

You're right.  Let's leave this disabled for now.

PR #296, opened 2016-03-23
Introduce rest api for delegation network queries.

fisx:
- Add Backend module and export fish sample networks.
- Allow GenArbitrary to run `Gen` actions.
- Make handler types and fish sample network generation more polymorphic.

This is all a bit silly: the new end-point logs in admin implicitly; the returned delegation networks are generated on top of the existing data; testing doesn't really test anything.  But it is self-contained and a good basis to continue from.

### DIFF
--- a/src/Arbitrary.hs
+++ b/src/Arbitrary.hs
@@ -390,7 +390,7 @@ instance Arbitrary Timestamp where
     arbitrary = Timestamp <$> arb
 
 instance GenArbitrary r => GenArbitrary (Action r) where
-    genArbitrary = liftIO $ generate arbitrary
+    genGen = liftIO . generate
 
 
 -- * arbitrary readable text
@@ -739,7 +739,7 @@ fishAvatars =
     , "tetraodontiformes/thumnails/tetraodon_nigroviridis.gif"
     ]
 
-mkFishUser :: (GenArbitrary m, ActionM r m) => Maybe SchoolClass -> URL -> m User
+mkFishUser :: (GenArbitrary r, ActionM r m) => Maybe SchoolClass -> URL -> m User
 mkFishUser mSchoolClass (("http://zierfischverzeichnis.de/klassen/pisces/" <>) -> avatar) = do
     let first_last = cs . takeBaseName . cs $ avatar
         (fnam, lnam) = case ST.findIndex (== '_') first_last of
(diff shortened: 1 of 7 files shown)


andorp:
LGTM

andorp:
@np Feel free to merge it, after review.

src/Arbitrary.hs
```diff
@@ -385,7 +385,7 @@ instance Arbitrary Timestamp where
     arbitrary = Timestamp <$> arb
 
 instance GenArbitrary r => GenArbitrary (Action r) where
-    genArbitrary = liftIO $ generate arbitrary
+    genGen = liftIO . generate
```
np: at this stage we should just call it `generate` and hide the original one.

np: Actually with DefaultSignatures this is quite decent. Here is what we should do for most those IO actions:

```
class Foo m where
  foo :: A -> B -> m C
  default foo :: MonadIO m => A -> B -> m C
  foo a b = liftIO $ Orig.foo a b
```

src/Frontend.hs
```diff
@@ -60,11 +61,15 @@ runFrontend cfg = do
     persist <- Persistent.Implementation.STM.mkRunPersist
     let runAction :: Action Persistent.Implementation.STM.Persist :~> ExceptT ServantErr IO
         runAction = mkRunAction (ActionEnv persist cfg)
-        aulaTopProxy = Proxy :: Proxy AulaTop
-        stateProxy   = Proxy :: Proxy UserState
+
+        aulaTopProxy           = Proxy :: Proxy AulaTop
+        stateProxy             = Proxy :: Proxy UserState
         aulaMainOrTestingProxy = Proxy :: Proxy (AulaMain :<|> "testing" :> AulaTesting)
```
np: Why not put the api next to main and testing? On the one hand one might want to use a different stack but so far this is the same so there is no need to spawn the whole thing twice (session handling and so on).

np:
LGTM, I'll merge if I need this part.

src/Arbitrary.hs
fisx: I may miss a point here, but I think I disagree.  I am in favour of calling things different if they are technically different.  `generate` requires `IO`, and we want to get away from `IO` here.

I'm not fond of the `genGen` name, but I would like to keep a name that's not `generate`, and avoid the default method trick.

src/Frontend.hs
fisx: yes, that would be better.  i'll do that.

PR #298, opened 2016-03-24
Only ideas deemed non-feasible need an explanation for their result

np:


fisx:
We actually want to allow for justifying *any* verdict, both feasible and not feasible.  But I think we will develop a saner model of the phases and how they interact soon, and then these types will probably change slightly again.  So...

LGTM.

Issue #208, opened 2016-03-14
404 content-type enconding


fisx:
...  under `/bröken`, the heart in the footer is not displayed correctly.

fisx:
related: #121, #146

fisx:
not sure if that's related: when not logged in, we should always redirect to logged in. currently for non-existign pages we get 404 instead.

Issue #146, opened 2016-03-01
404 under /static behaves oddly.


fisx:
Download of `/static/does_not_exist.csv` in chrome shows the user a source dump of the framed 404 page.  Probably because the file ending suggests that it's not html?

To fix this, we probably need to customize wai's `serveDirectory` some more.

fisx:
Possibly related: #208.

Issue #121, opened 2016-02-25
Render 500 Server Error Page for exceptions.


andorp:
Rendered 500 page should show information in the frame.

fisx:
(This is less urgent IMO, because we already get information on 500 errors on stdout.)

fisx:
Also, losely related: 404 shouldn't be rendered in a `Middleware`, but by a catch-all end point.  This could either be a `Raw`, or a custom combinator and `HasServer` instance.

fisx:
also losely related: if you are not logged in, you get 404 on pages that do not exist, and 303/login on pages that do.  it should be the latter for both.

fisx:
there was talk about a catchall handler over at servant.  not sure what the status there is, or whether it helps.

PR #299, opened 2016-03-24
WIP 404

np:
This is unfinished and here such that we don't loose track of it. The main question becomes should be the persistent layer have exceptions or should we mark explicitly the exceptions with (Maybe, Either) on a per-function basis.

### DIFF
--- a/src/Action/Implementation.hs
+++ b/src/Action/Implementation.hs
@@ -1,3 +1,4 @@
+{-# LANGUAGE FlexibleContexts           #-}
 {-# LANGUAGE FlexibleInstances          #-}
 {-# LANGUAGE GeneralizedNewtypeDeriving #-}
 {-# LANGUAGE TypeOperators              #-}
@@ -50,7 +51,7 @@ instance PersistM r => ActionM r (Action r)
 instance ActionLog (Action r) where
     logEvent = liftIO . print
 
-instance PersistM r => ActionPersist r (Action r) where
+instance (MonadError ServantErr r, PersistM r) => ActionPersist r (Action r) where
     persistent r = view persistNat >>= \(Nat rp) -> liftIO $ rp r
 
 instance MonadLIO DCLabel (Action r) where
--- a/src/Frontend/Page/Admin.hs
+++ b/src/Frontend/Page/Admin.hs
@@ -480,7 +480,7 @@ adminSettingsGaPUserEdit uid = redirectFormHandler editUserPage editUser
   where
     editUserPage = persistent $
         PageAdminSettingsGaPUsersEdit
-        <$> ((\(Just u) -> u) <$> findUser uid) -- FIXME: Error handling (404?)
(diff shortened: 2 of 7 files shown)


fisx:
Here is my position on this.  Please everybody see if you have any concerns, otherwise this will become official law.  (@kosmikus: you and me had this discussion with thentos; did i miss anything that we concluded there?)

1. we are going to use acid-state for production in the first year, and something postgres-based as soon as scaling or availability become issues.
1. my goal is to keep all three backends (STM, acid-state, postgres-whatever) productive concurrently.  if the cost for this is getting too high, we should give that up, but it would be nice to have for testing.  also, we may get better performance on small production deployments with acid-state than with postgres.
1. we will have `class (Monad m, MonadError SomeErr m) => PersistM m`.  This has effects on acid-state performance, but I'm not too worried about those.  See http://www.petterbergman.se/acid_error.html for the idea.  The alternative (`Maybe`, `Either`) would lead to convoluted transaction code, and I would like to avoid that.
1. `SomeErr` may remain `ServantErr` for a while, but I think I want to change it to a custom error type like in thentos, only with more fine-grained "sub-typing" (i.e., nesting of `CustomError a` constructors).

np:
I fully agree on having `MonadError` part of `PersistM`. On the `STM` implementation this should be pretty easy to add an `ExceptT` layer, still some care is need to make sure it does what we want.

1&2. let's discuss that on the call
3. agreed
4. agreed. The "prismatic errors" I've used before makes this sub-typing even better IMO.

andorp:
I agree with the points above. I have a strong opinion on software architectures, which is based on modules and boundaries of modules. In this case I would create and ADT with constructors like:

````haskell
data PersistEx
  = NotFound (forall a . AUID a)
  | Exception SomeException 
````

Of course there is a question, if the natural transformation from `Persist ~> Action` should react on the exception transforming to `Perist ~> ExceptT Action` or just silently handle some special cases and re-throw others, in which case the natural transformation would stay `Persist ~> Action` I don't have a strong opinion here.

The question here we need to answer: Do we want to implement error case handling in the client code that uses the Action/Persist functionality? Eg: Rendering different page snippets if one of the entities were not found, or do we handle errors only in the action handling, showing error pages to the users?

@np had a good point in the morning call. There is a need for investigation on this point.

np:
@andorp I think they should all be re-thrown (from Persist to Action), from what I've seen in almost all the cases we actually don't to catch it and just let the main handler produce an error page.

andorp:
I would like to take this over. Please assign it to me, if you are OK with it.

np:
We need to wait for the big acid-state PR to be merged, except if you want to rebase or redo it on top #333 ?

fisx:
related: #208, #146, #121.

fisx:
this discussion is still valid, but the code is either in master from elsewhere or not applicable any more.  i'm going to close this.

PR #300, opened 2016-03-24
Towards support for posting comments and replies

np:


src/DemoData.hs
```diff
@@ -156,6 +187,9 @@ universe rnd = void $ do
 
 -- * Helpers
 
+-- This 'Monad m => m a' is strange, why not simply 'a' instead?
+-- Second, is this actually safe to use the same generator over and over or
+-- does it need to be threaded?
```
andorp: there was a version, where I used split, but the code was noisy, when we use this module to generate better test data, not just demo data, we can use split and safer approach.

src/DemoData.hs
```diff
@@ -156,6 +187,9 @@ universe rnd = void $ do
 
 -- * Helpers
 
+-- This 'Monad m => m a' is strange, why not simply 'a' instead?
```
andorp: `Monad a` is used in favor of do notation and imperative style. No hard opinion here.

src/DemoData.hs
np: Ok this is good enough for demo data but instead of split we should use a monad. I guess that Gen was getting in the way of persist. If one of them was a monad transformer we could easily combine the two.

andorp: There is one `GenT` on hackage: https://hackage.haskell.org/package/QuickCheck-GenT

fisx:
Great work, thanks!

LGTM, somebody please review my comments and merge.

fisx:
(Forgot one general remark: If a comment is a question pointing to a gap in the specification, it should be marked either as FIXME or as TODO.)

src/Frontend/Page/Idea.hs
```diff
@@ -62,6 +64,13 @@ data EditIdea = EditIdea Idea
 instance Page EditIdea where
   isPrivatePage _ = True
 
+-- | X. Comment idea
+data CommentIdea = CommentIdea Idea
+  deriving (Eq, Show, Read)
+
+instance Page CommentIdea where
+  isPrivatePage _ = True
```
andorp: `isPrivatePage _ = True` is the default value. This line can be removed.

andorp:
It is a pretty big change. I could not point out big things I dislike.

andorp:
LGTM

src/DemoData.hs
np: `GenT` looks good.

src/Frontend/Page/Comment.hs
```diff
@@ -50,7 +50,28 @@ instance ToHtml CommentWidget where
                         i_ [class_ "icon-flag"] nil
                         "melden"
         div_ [class_ "comment-replies"] $ do
-            for_ (comment ^. commentReplies) $ \reply -> do
-                div_ [class_ "comment-reply"] $ do
```
np: My hope was that the class comment-reply would be enough for styling.

src/Persistent/Api.hs
```diff
@@ -171,7 +176,7 @@ type AddDb m a = UserWithProto a -> PersistM m => m a
 -- global map of the comments, votes, likes and still call @addDb@ only once.
 addDb :: (HasMetaInfo a, FromProto a) => AulaTraversal (AMap a) -> AddDb m a
 addDb l (cUser, pa) = do
-    when enforceDebuggingAssertions $ do
+    assertPersistM $ do
```
np: Nice.

src/Persistent/Api.hs
```diff
@@ -308,8 +313,7 @@ instance FromProto IdeaLike where
     fromProto () = IdeaLike
 
 -- FIXME: Same user can like the same idea more than once.
--- Assumption (check when enforceDebuggingAssertions):
--- the given @AUID Idea@ MUST be in the DB.
+-- FIXME: Assumption: the given @AUID Idea@ MUST be in the DB.
```
np: Why the FIXMEs?

src/Frontend.hs
```diff
@@ -204,7 +204,7 @@ aulaSpace space =
        Page.viewIdeas  space
   :<|> Page.viewIdea
   :<|> Page.editIdea
-  :<|> Page.commentIdea locSpace
+  :<|> Page.commentIdea
```
np: I don't really like the fact we don't use/check parts of the url (here the space).

np:
Your changes LGTM @fisx

src/Frontend/Page/Comment.hs
fisx: good point.  but if that's the case, we can always revert to one widget type after styling.

fisx: > My hope was that the class comment-reply would be enough for styling.

good point.  if that's the case though, we can just revert to one widget type after styling.

src/Persistent/Api.hs
fisx: because they should be assertions?  perhaps that was a misunderstanding, and we don't want to / can't assert this?

src/Frontend.hs
fisx: good point.  i think though that the way to resolve this is not to artificially use every part of the uri path, but to prune the path until it only contains useful parts.  see also: #302.

PR #306, opened 2016-03-28
Misc

fisx:
See commit messages.

The most interesting part may be the config files.  The new functionality should be sufficient for the first year.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,8 +1,8 @@
 SHELL=/bin/bash
 EXEC=`test -d .stack-work/ && echo "stack exec --" || echo "cabal exec --"`
 HLINT=$(EXEC) hlint
-SOME_SOURCES=-isrc -itests -idist/build/autogen
-FULL_SOURCES=$(SOME_SOURCES) -i$(THENTOS_ROOT_PATH)/src/ -i$(THENTOS_ROOT_PATH)/../thentos-tests/src/ -i$(THENTOS_ROOT_PATH)/../thentos-tests/tests/
+AULA_SOURCES=-isrc -itests -idist/build/autogen
+FULL_SOURCES=$(AULA_SOURCES) -i$(THENTOS_ROOT_PATH)/src/ -i$(THENTOS_ROOT_PATH)/../thentos-tests/src/ -i$(THENTOS_ROOT_PATH)/../thentos-tests/tests/
 AULA_IMAGE=quay.io/liqd/aula
 
 .phony:
@@ -25,7 +25,7 @@ unregister-full:
 
 # only aware of aula sources
 sensei: .phony aula.unregister
-	$(EXEC) sensei $(SOME_SOURCES) tests/Spec.hs $(SENSEI_ARGS)
+	$(EXEC) sensei $(AULA_SOURCES) tests/Spec.hs $(SENSEI_ARGS)
 
 # aware of aula and thentos sources
 sensei-full: .phony unregister-full
--- a/Makefile
+++ b/Makefile
@@ -51,11 +51,10 @@ hlint:
 	find src exec tests -name '*.hs' | xargs $(HLINT)
 
 test-everything:
+	cabal install --enable-test --ghc-options="-Werror -Wall -O0"
+	cabal test
 	make hlint
 	make click-dummies-recreate
-	cabal test
-	cabal install --enable-test --ghc-options="-Werror -Wall -O0"
-	cabal haddock
 
 ghci-no-type-errors:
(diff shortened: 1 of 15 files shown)


src/Config.hs
```diff
@@ -34,13 +70,44 @@ devel = Config
...
+        ]
 
+    msg2 :: String -> IO ()
+    msg2 emsg = f
+        [ "could not read config file:"
+        , emsg
+        , "to fix this, write the following lines to $AULA_ROOT_PATH/aula.yaml:"
+        ]
+
+    f :: [String] -> IO ()
```
andorp: Please rename `f` for a more meaningful name, like `warnings`.

Makefile
```diff
@@ -1,7 +1,8 @@
 SHELL=/bin/bash
 EXEC=`test -d .stack-work/ && echo "stack exec --" || echo "cabal exec --"`
 HLINT=$(EXEC) hlint
-FULL_SOURCES=-isrc -itests -i$(THENTOS_ROOT_PATH)/src/ -i$(THENTOS_ROOT_PATH)/../thentos-tests/src/ -i$(THENTOS_ROOT_PATH)/../thentos-tests/tests/
+SOME_SOURCES=-isrc -itests -idist/build/autogen
```
andorp: Please rename this to `AULA_SOURCES`.

src/Types.hs
```diff
@@ -359,8 +359,8 @@ instance SOP.Generic ProtoDelegation
 
 data DelegationContext =
     DelCtxIdeaSpace { _delCtxIdeaSpace :: IdeaSpace  }
-  | DelCtxTopic     { _delCtxTopicId   :: AUID Topic }
-  | DelCtxIdea      { _delCtxIdeaId    :: AUID Idea  }
+  | DelCtxTopicId   { _delCtxTopicId   :: AUID Topic }
```
andorp: Nitpick we should rename `DelCtx...` to `DlgCtx...`. First I thought this is about deletion of some context.

andorp:
LGTM.

andorp:
@np Please take a look at the login part.

src/Types.hs
fisx: sounds good.  changed!  (-:

src/Action.hs
```diff
@@ -119,7 +120,7 @@ instance ThrowError500 ActionExcept where
 
 class ActionError m => ActionUserHandler m where
     -- | Make the user logged in
-    login  :: User -> m ()
+    login  :: AUID User -> m ()
```
np: The point of taking a user what to somewhat enforce that we are passing a user we have in the DB. I agree that even the previous version did not enforce that completely but still.

src/Action.hs
```diff
@@ -133,6 +134,14 @@ class MonadError ActionExcept m => ActionError m
 
 -- * Action Combinators
 
+loginByUser :: ActionUserHandler m => User -> m ()
+loginByUser = login . view _Id
```
np: The issue I have with such functions is that this easy to end-up looking a user by id in the DB just to call this function which then discards all but the Id.

src/Config.hs
```diff
@@ -1,16 +1,49 @@
...
 
+
+instance ToJSON CsrfSecret where
+  toJSON (CsrfSecret s) = String $ cs s
+
+instance FromJSON CsrfSecret where
+  parseJSON o = CsrfSecret . (cs :: String -> SBS) <$> parseJSON o
+
+-- FIXME: if we decide that json instances for csrf secrets are a good idea, they should move to
+-- thentos.
```
np: I'm ok with them.

tests/AulaTests.hs
```diff
@@ -30,10 +30,11 @@ import Frontend.Prelude as X hiding (get, put)
...
 
 testConfig :: IO Config
-testConfig = (devel & generateDemoData .~ False &) . (listenerPort .~) <$> pop
-  where
-    pop :: IO Int
-    pop = modifyMVar testConfigPortSource $ \(h:t) -> pure (t, h)
+testConfig = do
+    cfg <- getConfig DontWarnMissing
+    pop <- modifyMVar testConfigPortSource $ \(h:t) -> pure (t, h)
+    return (cfg & generateDemoData .~ False & listenerPort .~ pop)
```
np: I think these `&` tend to look much better when aligned:

```
pure $ cfg & generateDemoData .~ False
                 & listenerPort .~ pop
```

This might also work if you like it:

```
cfg & generateDemoData .~ False
    & listenerPort .~ pop
    & pure
```

np:
LGTM

PR #311, opened 2016-03-29
Basic support for voting

np:


src/Frontend/Path.hs
```diff
@@ -78,17 +81,7 @@ main (Login mb)       root = root </> fromString ("login" <> if fromMaybe True m
...
-  | ListTopics
-  | ListTopicIdeas (AUID Topic)
-  | ViewTopicIdea (AUID Topic) (AUID Idea)
-  | EditTopicIdea (AUID Topic) (AUID Idea)
-  | CommentTopicIdea (AUID Topic) (AUID Idea)
-  | CreateTopicIdea (AUID Topic)
+    ListTopics
   | ViewTopicIdeasVoting (AUID Topic)
   | ViewTopicIdeasWinning (AUID Topic)
   | ViewTopicDelegations (AUID Topic)
```
fisx: Nice!  (-:

(I suspect that other `ViewTopicIDeas*` will also move to the idea paths.  That would leave two path constructors under `Space`.)

fisx:
LGTM.  I've played with idea-level votes / likes and will merge if travis has no objections.

np:
Thanks for the play code I was currently working on making the buttons work ran into troubles:

* Phases: I was looking for an idea in the right phase to be voted on. Is there a way as an admin to force a particular phase?
* Context: To vote on comments one needs to keep track of its location space and idea. This should be straight forward to include this in the widget type.

fisx:
@np: @andorp is working on moving topics to other phases.  i think the easiest way to get them is through the demo data generator.  the admin is not supposed to be able to freely move phases around.  i'm not sure if it's worth it, but i guess we could implement a rest api for doing that, create a new role 'god' or 'janitor', and give it access to that api.

PR #316, opened 2016-03-30
Persist vs. Except

fisx:
This introduces exceptions for PersistM in a minimal way.

The next step would be to change the types `ActionExcept` and `PersistExcept` as indicated in the resp. FIXMEs.

### DIFF
--- a/src/Action/Implementation.hs
+++ b/src/Action/Implementation.hs
@@ -13,7 +13,7 @@ where
 
 import Control.Exception (SomeException(SomeException), catch)
 import Control.Lens
-import Control.Monad.Except (MonadError)
+import Control.Monad.Except (MonadError, throwError)
 import Control.Monad.IO.Class
 import Control.Monad.RWS.Lazy
 import Control.Monad.Trans.Except (ExceptT(..), runExceptT, withExceptT)
@@ -50,7 +50,10 @@ instance ActionLog (Action r) where
     logEvent = liftIO . print
 
 instance PersistM r => ActionPersist r (Action r) where
-    persistent r = view persistNat >>= \(Nat rp) -> liftIO $ rp r
+    persistent r = do
+        Nat rp <- view persistNat
+        v  <- liftIO . runExceptT . rp $ r
+        either (throwError . ActionExcept) pure v  -- TODO: is this strict enough?  how can we test this?
 
 instance MonadLIO DCLabel (Action r) where
     liftLIO = liftIO . (`evalLIO` LIOState dcBottom dcBottom)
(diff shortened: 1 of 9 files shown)


fisx:
does not compile yet, but is ready for feedback.  the interesting commit is 359b3d8.

src/Frontend.hs
```diff
@@ -68,11 +68,12 @@ runFrontend cfg = do
     app <- serveFAction (Proxy :: Proxy AulaActions) stateProxy extendClearanceOnSessionToken
             runAction aulaActions
 
-    unNat persist genInitialTestDb -- FIXME: Remove Bootstrapping DB
+    Right _ <- runExceptT $ unNat persist genInitialTestDb -- FIXME: Remove Bootstrapping DB
```
andorp: This could be defined in a function, saving lot of boilerplate.

````haskell
runPersist :: (Persist :~> ExceptT e IO) -> Persist a -> IO a
runPersist persist m = either (error . show) id <$> runExceptT (unNat persist) m
````

This helper could be used in the Test Suite and here, as this snippet is going to be removed soon.

**Comment by andorp** [comment]
In the future, we should introduce prismatic exceptions or wrapped exceptions, but for the time being, I am happy with this change.

**Comment by andorp** [comment]
LGTM

**Comment by Mikolaj** [comment]
oh, cool, much better now; my problem was shallow, syntactic, so @andorp's helper really helps :)

**Comment by Mikolaj** [comment]
and generally LGTM

File: src/Action/Implementation.hs [hunk 2]
```diff
@@ -50,7 +50,10 @@ instance ActionLog (Action r) where
     logEvent = liftIO . print
 
 instance PersistM r => ActionPersist r (Action r) where
-    persistent r = view persistNat >>= \(Nat rp) -> liftIO $ rp r
+    persistent r = do
+        Nat rp <- view persistNat
+        v  <- liftIO . runExceptT . rp $ r
+        either (throwError . ActionExcept) pure v  -- TODO: is this strict enough?  how can we test this?
```
np: this is enough, not sure how to test.

src/Persistent/Implementation/STM.hs
```diff
@@ -24,8 +27,8 @@ import Persistent.Api
 -- FIXME: Remove
 import Test.QuickCheck (generate)
 
-newtype Persist a = Persist (ReaderT (TVar AulaData) IO a)
-  deriving (Functor, Applicative, Monad)
+newtype Persist a = Persist (ExceptT ServantErr (ReaderT (TVar AulaData) IO) a)
```
np: Semantically the order of `ExceptT` and `ReaderT` does not matter. However it might more convenient the other way around.

src/Frontend.hs
np: Agreed.

np:
LGTM

PR #317, opened 2016-03-30
2016 03 30 mk acid state persistent

Mikolaj:


src/Persistent/Implementation.hs
```diff
@@ -0,0 +1,17 @@
+{-# LANGUAGE CPP #-}
```
fisx: I would like to avoid CPP.  Two reasons:

- complicates and destabilizes parsing, e.g. with hlint.
- makes harder to write code like e.g. "run one backend of this type and one of that; call the same actions on both; compare the result".

The benefit of CPP would be that the production code would be less bulky, but I don't want to optimize for this yet.

I think it's quite simple to replace the CPP-switch with a config-based switch.  Am I missing something?  Any other opinions?

src/Persistent/Implementation/AcidState.hs
```diff
@@ -0,0 +1,65 @@
...
+newtype Persist a = Persist (ReaderT (AcidState AulaData) IO a)
+  deriving (Functor, Applicative, Monad)
+
+persistIO :: IO a -> Persist a
+persistIO = Persist . liftIO
+
+instance GenArbitrary Persist where
+    genGen = persistIO . generate
+
+-- TODO: call @closeAcidState db@ somewhere
```
fisx: We need to catch SIGINT, which is a ghc exception.  I remember we've done it in thentos before moving away from acid-state.  if you look at june 2015 there, you might find it.

src/Persistent/Implementation.hs
fisx: Oh, you don't only need to replace `mkRunPersist`, but also `Persist`.  Hum...  type families?  We may want to keep the CPP solution for now if that gets too complicated.

fisx:
(test suite still failing because i've shuffeled the opens and closes around too much.  investigating...)

src/Persistent/Implementation.hs
Mikolaj: Yes, the PR is big enough and long-running. so I will keep CPP for now, with cabal flags. But both Persist implementations are being compiled every time, the CPP is just to pick which one to use without tons of needless generalizations everywhere.

np: I guess this shouldn't be too hard to make the implementation dynamically chosen we could even hide the Persist type completely I think.

np: To be honest the use of CPP is really harmless here.

src/Persistent/Implementation/AcidState.hs
```diff
@@ -0,0 +1,80 @@
...
+
+persistIO :: IO a -> Persist a
+persistIO = Persist . liftIO
+
+instance GenArbitrary Persist where
+    genGen = persistIO . generate
+
+mkRunPersistGeneric :: (AulaData -> IO (AcidState AulaData))
+                    -> (AcidState AulaData -> IO ())
+                    -> IO (Persist :~> ExceptT PersistExcept IO, IO ())
```
np: What about having a data type for our persistence implementations, something such as:

```
data RunPersist where
   RunPersist :: { _runPersistNat :: persist :~> ExceptT PersistExcept IO
                         , _closePersist :: IO ()
                         }
```

The `persist` is effectively an existential type.

Mikolaj: Yes, if you could pull it off, would be great. I guess then half of each of the Persistent,Implementation.* modules would become shared. I'd suggest another PR, though, because it's a refactoring.

fisx: If anybody gets around to do that I'm happy with it.  But yes, preferably in another PR.

(`Nat` always wraps existential quanitification around its types, right?  Just making sure I understand.)

src/Persistent/Implementation.hs
fisx: agreed.  let's keep it as is.

src/Persistent/Implementation/AcidState.hs
andorp: I like the `data RunPersist` idea.

src/Persistent/Implementation/AcidState.hs
```diff
@@ -0,0 +1,80 @@
...
+askDbM :: Query AulaData AulaData
+askDbM = ask
+
+getDbM :: Update AulaData AulaData
+getDbM = get
+
+putDbM :: AulaData -> Update AulaData ()
+putDbM = put
+
+makeAcidic ''AulaData ['askDbM, 'getDbM, 'putDbM]
```
np: As @fisx mentioned the `putDbM` is too bulky. Each tiny change to the DB will synchronously dump the whole thing to the changelog.

This makes a really non-intrusive first step though.

src/Persistent/Implementation/AcidState.hs
```diff
@@ -0,0 +1,80 @@
...
+
+instance PersistM Persist where
+    getDb l = Persist . ExceptT . ReaderT $ fmap (Right . view l) . flip query AskDbM
+    modifyDb l f = Persist . ExceptT . ReaderT $ \state -> fmap Right $ do
+      aulaData <- update state GetDbM
```
np: What happens if one uses `query state AskDbM` instead of `update state GetDbM` here?

src/Persistent/Implementation/AcidState.hs
Mikolaj: Agreed.

src/Persistent/Implementation/AcidState.hs
Mikolaj: Wrong monad error. Then you can try lifting stuff, but it gets complex and probably amounts to the same in the end.

Mikolaj: TBH, I haven't investigated how it differs re log size, etc.

src/Persistent/Implementation/AcidState.hs
np: @fisx in `Nat` this is really a `forall` quantification, I say really because Haskell implements existential quantification through `forall` quantification of data-constructors. In `Nat` however this is the field which has a `forall`

PR #319, opened 2016-03-30
2016 03 cc comment styling

localgrr:


fisx:
LGTM, thanks!

I'm not happy with the distance between the comment replies, and I've added a FIXME for that, but this is still an improvement, so I'll merge it once travis is green.

PR #320, opened 2016-03-30
Phase transitions

fisx:


src/Action.hs
```diff
@@ -186,6 +189,28 @@ isLoggedIn :: ActionUserHandler m => m Bool
...
 
 
+-- * Phase Transitions
+
+topicInRefinementTimedOut :: (ActionPersist r m, ActionUserHandler m) => AUID Topic -> m ()
+topicInRefinementTimedOut tid =
+    join . persistent $ do
+        Just topic <- findTopic tid -- FIXME: Not found
+        case phaseTrans (topic ^. topicPhase) RefinementPhaseTimeOut of
+            Nothing -> throwError $ persistError "Invalid phase transition"
```
andorp: This is going to be thrown in Action monad, will  be done in next PR.

PR #322, opened 2016-03-30
Some work on html

fisx:


src/Frontend/Page/Idea.hs
```diff
@@ -97,12 +107,12 @@ instance ToHtml ViewIdea where
                 idea ^. createdByLogin . fromUserLogin . html
                 " / "
                 when (phase `elem` [Nothing, Just PhaseRefinement]) $ do
-                    totalLikes ^. showed . html <> " Likes"  -- FIXME: singular
-                    " / "
+                    numberWithUnit totalLikes "Like" "Likes"
```
andorp: Does the `"Likes"` is meaningful in German?

andorp:
LGTM

src/Frontend/Page/Idea.hs
fisx: yes, i think we discussed that and decided it is the best "translation".

np:
LGTM

PR #325, opened 2016-03-31
Support reply and vote buttons

np:
* Styling of comment voting buttons has been a bit broken since now there
  is a form element and a button element.
* Handlers for comment replies.
* Pass along the necessary data to vote on comments and replies.


andorp:
LGTM

PR #327, opened 2016-03-31
Dynamic selection of the persistence implementation

np:


fisx:
btw would it make sense to have three backend options, namely AcidStateMem, AcidStateDisk, STM, and drop the `*InMemory` variants of `mkPersist` etc.?  i think it would.

src/Persistent/Implementation.hs
```diff
@@ -1,16 +1,22 @@
...
+
+withPersist' :: Config -> (forall r. (PersistM r, GenArbitrary r) => RunPersistNat IO r -> IO a) -> IO a
+withPersist' cfg = withPersist (mkRunPersist cfg)
+
+mkRunPersist :: Config -> IO RunPersist
+mkRunPersist cfg =
+    case cfg ^. persistenceImpl of
+        STM             -> mkRunPersistSTM
+        AcidStateInMem  -> mkRunPersistInMemory
+        AcidStateOnDisk -> mkRunPersistOnDisk cfg
```
np: @fisx isn't that what you were looking for?

fisx: yes, that looks nice!

np:
I'd like to have this merged regardless of #328

Mikolaj:
A good stuff.

A note: with this setup we have to remember to never run tests with `AcidStateOnDisk` (unless we start all individual tests by wiping out the DB, which I did do in some of my old reverted commits and they did pass with `AcidStateOnDisk`), but only with any of the other two. In the previous idiom, I hardcoded that tests are always run with mkRunPersistenInMemory, so there was no possibility of a mistake. Also, we don't want want to run tests on disk so as not to corrupt a persistent DB (production, staging, demo or any other) via tests. Anyway, just a warning to keep in mind.

tests/Persistent/ApiSpec.hs
```diff
@@ -148,55 +148,63 @@ changeAUID (AUID i) = AUID (succ i)
...
-    addDbSpec "addIdea"  getIdeas addIdea
-    getDbSpec "getWildIdeas"      getWildIdeas
-    getDbSpec "getIdeasWithTopic" getIdeasWithTopic
+    context "Using STM"             $ persistApiSpec STM
+    context "Using AcidStateInMem"  $ persistApiSpec AcidStateInMem
+
+    -- I guess we need to change the config to target a different directory.
+    -- And wipe out the data every time.
+    -- context "Using AcidStateOnDisk" $ persistApiSpec AcidStateOnDisk
+
```
Mikolaj: yep, see my top-level comment on the PR

PR #328, opened 2016-04-01
WIP Flesh out @np's idea about DbField for acid-state

Mikolaj:
After this experiment I think the @np's idea is feasible, but requires some non-trivial decisions, quite a bit of coding and a lens-fu I don't have.

In the first commit, all the needed components are partially coded, but need to be extended, there are some place-holders, some things don't compile and the compilation actually raises an exception at some point.

Please continue.

### DIFF
--- a/src/Persistent/Api.hs
+++ b/src/Persistent/Api.hs
@@ -98,7 +98,8 @@ import Data.Elocrypt (mkPassword)
 import Data.Foldable (find, for_)
 import Data.List (nub)
 import Data.Maybe (fromMaybe)
-import Data.SafeCopy (base, deriveSafeCopy)
+import Data.Serialize (Serialize,get,put)
+import Data.SafeCopy (SafeCopy, base, deriveSafeCopy)
 import Data.Set (Set)
 import Data.String.Conversions (ST, cs, (<>))
 import Data.Time.Clock (getCurrentTime)
@@ -159,6 +160,13 @@ data DbField a where
 
     -- AMap specific
     DbAt :: DbField (AMap a) -> AUID a -> DbField a
+
+-- TODO
+-- deriveSafeCopy 0 'base ''DbField
+instance Serialize (DbField a) where
+    put = error "Serialize (DbField a) TODO"
+    get = error "Serialize (DbField a) TODO"
+instance SafeCopy (DbField a) where
 
 dbFieldTraversal :: DbField a -> AulaTraversal a
 dbFieldTraversal = \case
--- a/src/Persistent/Api.hs
+++ b/src/Persistent/Api.hs
@@ -163,6 +163,28 @@ data DbField a where
     -- @Maybe@ specific
     DbJust :: DbField (Maybe a) -> DbField a
 
+instance Show (DbField a) where
+    show (DbSpaceSet) = "(DbSpaceSet)"
+    show (DbIdeas) = "(DbIdeas)"
+    show (DbUsers) = "(DbUsers)"
+    show (DbTopics) = "(DbTopics)"
+    show (DbDelegations) = "(DbDelegations)"
+    show (DbElaborationDuration) = "(DbElaborationDuration)"
+    show (DbVoteDuration) = "(DbVoteDuration)"
+    show (DbSchoolQuorum) = "(DbSchoolQuorum)"
+    show (DbClassQuorum) = "(DbClassQuorum)"
+    show (DbLastId) = "(DbLastId)"
+
+    show (DbIdeaLikes _) = "(DbIdeaLikes)"
+    show (DbIdeaVotes _) = "(DbIdeaVotes)"
+    show (DbIdeaComments _) = "(DbIdeaComments)"
+    show (DbCommentReplies _) = "(DbCommentReplies)"
+    show (DbCommentVotes _) = "(DbCommentVotes)"
+
+    show (DbAt _ _) = "(DbAt)"
+    show (DbJust _) = "(DbJust)"
+
+
 
 dbFieldTraversal :: DbField a -> AulaTraversal a
 dbFieldTraversal = \case
--- a/src/Persistent/Api.hs
+++ b/src/Persistent/Api.hs
@@ -264,7 +264,7 @@ addDb l (cUser, pa) = do
             fail $ "Persistent.Api.addDb expects the location (lens, traversal) "
                 <> "to target exactly 1 field not " <> show len
     a <- fromProto pa <$> nextMetaInfo cUser
-    modifyAMap l (a ^. _Id) (const a)
+    modifyDb (DbAt l (a ^. _Id)) (const $ Just a)
     return a
 
 addDbValue :: (HasMetaInfo a, FromProto a) => DbField a -> AddDb m a
--- a/src/Persistent/Api.hs
+++ b/src/Persistent/Api.hs
@@ -164,30 +164,6 @@ data DbField a where
     -- @Maybe@ specific
     DbJust :: DbField (Maybe a) -> DbField a
 
-instance Show (DbField a) where
-    show (DbId) = "(DbId)"
-    show (DbSpaceSet) = "(DbSpaceSet)"
-    show (DbIdeas) = "(DbIdeas)"
-    show (DbUsers) = "(DbUsers)"
-    show (DbTopics) = "(DbTopics)"
-    show (DbDelegations) = "(DbDelegations)"
-    show (DbElaborationDuration) = "(DbElaborationDuration)"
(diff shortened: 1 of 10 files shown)


Mikolaj:
The travis error report has more details about the exception raised during compilation and it seems, there is a problem with the polymorphism inside `makeAcidic`. In the worst case we may need to define for `makeAcidic` as many functions as there are constructors of the `DbField` type, but perhaps there is a better way, e.g., adding a phantom type to the state type, and/or defining the instances by hand, not via `makeAcidic`.

src/Persistent/Api.hs
```diff
@@ -325,7 +366,7 @@ instance FromProto IdeaLike where
 -- | FIXME: Same user can like the same idea more than once (Issue #308).
 -- FIXME: Assumption: the given @AUID Idea@ MUST be in the DB.
 addLikeToIdea :: AUID Idea -> AddDb m IdeaLike
-addLikeToIdea iid = addDb (dbIdeaMap . at iid . _Just . ideaLikes)
+addLikeToIdea iid = undefined -- TODO: addDb (dbIdeaMap . at iid . _Just . ideaLikes)
```
np: Here you need to give a name to `dbIdeaMap . at iid . _Just . ideaLikes`, this makes one constructor `DbIdeaLike :: AUID Idea -> DbField IdeaLikes` and the code goes to `dbFieldSetter`.

src/Persistent/Api.hs
```diff
@@ -198,16 +242,16 @@ type AddDb m a = UserWithProto a -> PersistM m => m a
...
 -- to index the record at different locations. For instance we could keep an additional
 -- global map of the comments, votes, likes and still call @addDb@ only once.
-addDb :: (HasMetaInfo a, FromProto a) => AulaTraversal (AMap a) -> AddDb m a
+addDb :: (HasMetaInfo a, FromProto a) => (AUID a -> DbField a) -> AddDb m a
 addDb l (cUser, pa) = do
+    a <- fromProto pa <$> nextMetaInfo cUser
     assertPersistM $ do
         db <- getDb id
-        let len = lengthOf l db
+        let len = undefined + 17 -- TODO: lengthOf (dbFieldSetter (l $ a ^. _Id) *and then go one level up*) db
```
np: Here you got the limitation that `dbFieldSetter` is returning well ... a setter. At first we can disable this assertion.

src/Persistent/Api.hs
```diff
@@ -126,11 +129,52 @@ makeLenses ''AulaData
...
+    DbTopic :: AUID Topic -> DbField Topic
+    DbDelegationMap :: DbField Delegations
+    DbDelegation :: AUID Delegation -> DbField Delegation
+    DbElaborationDuration :: DbField DurationDays
+    DbVoteDuration :: DbField DurationDays
+    DbSchoolQuorum :: DbField Percent
+    DbClassQuorum :: DbField Percent
+    DbLastId :: DbField Integer
+    DbCompleteAulaData :: DbField AulaData
+-- TODO: add more like that, pity it's not composable; or is it?
```
np: It could be but I would favor simplicity first. But since you ask... This would require to index by not only the target but the source:

```
data DbField a b where
   DbIdeaMap :: DbField AulaData Ideas
   ...
   At :: AUID a -> DbField (AMap a) (Maybe a)
   (:.:) :: DbField b c -> DbField a b -> DbField a c
   JustF :: DbField (Maybe a) a
```

src/Persistent/Api.hs
```diff
@@ -126,11 +129,52 @@ makeLenses ''AulaData
...
+    DbIdeaComment :: AUID Idea -> AUID Comment -> DbField Comment
+
+dbFieldSetter :: DbField a -> AulaSetter a
```
np: Does this type-checks with `AulaTraversal a`. If so this is better.

Mikolaj: I haven't tried. I just took what was in the type of `modifyDb` previously.

src/Persistent/Api.hs
Mikolaj: Actually, I meant more like `DbIdeaComment` below (my comments always refer to the item below, unless they start with a caret ^ or are indented wrt to the previous line). But yes, what you described is less constructors than the 10 more extra constructors like `DbIdeaComment` needed right now and more in the future.

src/Persistent/Api.hs
Mikolaj: Right now we have `DbIdeaComment` like that and we need a dozen more, I guess.

src/Persistent/Api.hs
np: `dbIdeaComment` would be a composition `:.:` of `DbIdeaMap`, `At`, `JustF` but these constructors might be not as convenient as their function counterpart so simpler is better here.
(Additionally what gets logged to the changelog will somewhat leak these details)

src/Persistent/Api.hs
np: Not a dozen.

src/Persistent/Api.hs
Mikolaj: Agreed.

src/Persistent/Implementation/AcidState.hs
```diff
@@ -66,8 +66,8 @@ instance MonadIO Persist where
 askDbM :: Query AulaData AulaData
 askDbM = ask
 
-getDbM :: DbField a -> Update AulaData a
-getDbM l = use (dbFieldTraversal l)
+getDbM :: Update AulaData AulaData
+getDbM = get
```
Mikolaj: Yes, you are right, I went overboard with that; indeed, getDbM doesn't need DbField.

fisx:
I rebased and killed all the original commits.  Hopefully no important comments have been lost...

fisx:
(The old commits have already been cleaned out.  Wow, that was fast..)

fisx:
[irc]

```
12:08 < fisx_> i have one more issue with the current situation at the tip of 
               #328 (it may not be very deep): if we call `persistent` to 
               combine two acid-state transactions, we do not get a composite 
               acid-state transaction.  so persistent doesn't give us acid.  
               instead, we'd have to introduce a new DbField constructor, i 
               think.  or a new type somewhere else.  i think if we just 
               remove `persistent` from the Action api and force the app logic 
               implementor to write a new DbField constructor (or something) 
               every time they would have called `persistent` before, we'll be 
               fine.
12:08 < fisx_> opinions?
```

fisx:
This still does not work:

```haskell
class (MonadError PersistExcept m, Monad m) => PersistM m where
    modifyDb :: DbTraversal a -> (a -> a) -> m ()
```

We have make the `Traversal` part serializable, but the operation on the serialized part is still a function.  This works because there was stil an error here that I missed in #317:

```haskell
    modifyDb l f = Persist . ExceptT . ReaderT $ \state -> fmap Right $ do
        aulaData <- update state GetDbM
        update state (PutDbM . (l %~ f) $ aulaData)
```

This is not acidic: it reads an `AulaData` value with an `ask`, lets somebody else change the value, then overwrites those changes with a `put`.

I see two ways to fix this:

1. make the function argument of `modifyDb` serializable in the same way we did with the `Traversal`.
2. do what acid-state is designed for: write a function for every transaction of type `Query` or `Update`, pass it to `makeAcidic`, and to all the pure operations on `AulaData` behind that list of functions.

In my opinion, the purpose of this PR is to figure out these things.  We must not trade consistency for being able to write data to disk.

I will keep investigating.

fisx:
Here's another issue:

```haskell
getCurrentTimestampIO :: AUpdate Timestamp
mkRandomPasswordIO :: AUpdate UserPass
```

If we call these in the transaction, the transaction goes to the changelog, and we decide to replay the changelog later, passwords and timestamps in the DB will change.

I have started a new attempt based on #327, that solves all these problems.  The Api will change quite a bit, but it looks like most of the beautiful lens magic can be rescued.

Mikolaj:
@fisx: unfortunately, you are right about the lack of ACID of @modifyDb@. Concurrent calls of @modifyDb@ are not isolated from each other, so we get a race condition. We'd need a lock on the DB between each @GetDbM@ and @PutDbM@ to recover ACID. (And manual locks are silly, what we need to do is to modify the DB *inside* the critical section that acid-state internally has for each update.) And yes, both options 1. and 2. would solve the problem.

Mikolaj:
Edit: After the @fisx analysis, I see we need makeAcidic for as many functions as there are constructors of the DbField type and the f functions used in modifyDb. However, if DbField is granular enough, the f functions are almost always const, so they can be serialized indeed.

fisx:
closing in favour of #340.

PR #330, opened 2016-04-01
Refactor the routing table [closes #302]

### LINKED ISSUE #302: Refactor Ideas: share code between wild ideas and ideas-in-topic better
We should at least share the sub-routing tables for CRUD between wild ideas and domestic ideas (aka ideas-in-topic).  But perhaps we can even remove the distinction from the routing tables altogether: An idea knows where it belongs, and we don't really need the URI path for that piece of information, so the route for an idea could remain the same before and after domestication.  This would have the extra benefit of slightly more stable URIs.

np:
Fixes #302

### DIFF
--- a/src/Frontend.hs
+++ b/src/Frontend.hs
@@ -13,6 +13,7 @@ module Frontend
 where
 
 import Control.Monad.Trans.Except
+import GHC.TypeLits (Symbol)
 import Lucid hiding (href_)
 import Network.HTTP.Types
 import Network.Wai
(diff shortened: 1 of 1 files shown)


src/Frontend.hs
```diff
@@ -138,19 +139,49 @@ aulaActions =
   :<|> Backend.api
   :<|> aulaTesting
 
+data Reply
+
+type family Singular    a :: Symbol
+type family Plural      a :: Symbol
```
fisx: Plural doesn't appear to be used anywhere.  Is that ok?

fisx:
LGTM.  One comment (question), but I'll merge anyway.

PR #332, opened 2016-04-02
Jury Phase

andorp:


src/Frontend/Testing.hs
```diff
@@ -0,0 +1,19 @@
...
+
+import Action
+import Persistent
+import Types
+
+-- | Make a topic timeout if the timeout is applicable for the topic.
+-- FIXME: Make timeout the jury phase too.
+makeTopicTimeout :: (ActionPersist r m, ActionUserHandler m) => AUID Topic -> m ()
+makeTopicTimeout tid = do
+    Just topic <- persistent $ findTopic tid -- FIXME: 404
```
fisx: this 404 will be thrown by `PersistM`, and we don't have to worry about it here.

src/Frontend/Testing.hs
```diff
@@ -0,0 +1,19 @@
...
+    case topic ^. topicPhase of
+        PhaseRefinement _ -> topicInRefinementTimedOut tid
+        _                 -> return ()
```
fisx: ...  consequently, we can shorten this:

```haskell
topicInRefinementTimedOut tid <$$> view topicPhase <$> persistent (findTopic tid)
```

src/Frontend/Testing.hs
```diff
@@ -0,0 +1,19 @@
+{-# OPTIONS_GHC -Werror -Wall #-}
+
+module Frontend.Testing
```
fisx: I would have appended the contents of this module to the end of `Frontend`.  If you want to have this module, could you move the entire `{A,a}ulaTesting` sub-api here?

src/Frontend/Testing.hs
```diff
@@ -0,0 +1,19 @@
...
+import Control.Lens ((^.))
+
+import Action
+import Persistent
+import Types
+
+-- | Make a topic timeout if the timeout is applicable.
+makeTopicTimeout :: (ActionPersist r m, ActionUserHandler m) => AUID Topic -> m ()
+makeTopicTimeout tid = do
+    Just topic <- persistent $ findTopic tid -- FIXME: 404
```
fisx: ah, some previous comments do not apply any more.  but the FIXME can be removed: it's testing code, and it is not supposed to be fixed here anyway.

src/Frontend/Testing.hs
andorp: I am agree, please adjust this module. :)

src/Persistent/Idiom.hs
```diff
@@ -49,28 +49,51 @@ quorum idea = case idea ^. ideaLocation . ideaLocationSpace of
...
+    DurationDays (days :: Int) <- getDb dbVoteDuration
+    phaseEnd days
+
+
+-- | Returns the Just topic of an idea if the idea is assocaited with a topic, Nothing
 -- if the idea is a wild idea, or throws an error if the topic is missing.
-ideaPhase :: PersistM m => Idea -> m (Maybe Phase)
-ideaPhase idea = case idea ^. ideaLocation of
+ideaTopic :: PersistM m => Idea -> m (Maybe Topic)
+ideaTopic idea = case idea ^. ideaLocation of
```
fisx: we really need to clean up Persistent.Api, and now Persistent.Idiom.  there are a lot of functions similar to this one in Api already.  (to be done later)

fisx:
LGTM.  can be merged if travis is green.

PR #333, opened 2016-04-03
acid-state: a less abstract approach

fisx:
#328 has unresolved issues.  This is the alternative.

### DIFF
--- a/src/Action.hs
+++ b/src/Action.hs
@@ -224,8 +224,7 @@ validUserState us = us == userLoggedOut || validLoggedIn us
 
 -- * Phase Transitions
 
-topicPhaseChange
-    :: (ActionPersist m, ActionUserHandler m) => Topic -> PhaseChange -> m ()
+topicPhaseChange :: Topic -> PhaseChange -> AUpdate ()
 topicPhaseChange topic change = do
     case phaseTrans (topic ^. topicPhase) change of
         Nothing -> throwError500 "Invalid phase transition"
@@ -233,12 +232,12 @@ topicPhaseChange topic change = do
             aupdate $ SetTopicPhase (topic ^. _Id) phase'
             mapM_ (phaseAction topic) actions
 
-topicTimeout :: (ActionPersist m, ActionUserHandler m) => PhaseChange -> AUID Topic -> m ()
+topicTimeout :: PhaseChange -> AUID Topic -> AUpdate ()
 topicTimeout phaseChange tid = do
     Just topic <- aquery $ findTopic tid -- FIXME: Not found
     topicPhaseChange topic phaseChange
 
-phaseAction :: (ActionPersist m, ActionUserHandler m) => Topic -> PhaseAction -> m ()
+phaseAction :: Topic -> PhaseAction -> AUpdate ()
 phaseAction _ JuryPhasePrincipalEmail =
     traceShow "phaseAction JuryPhasePrincipalEmail" $ pure ()
 phaseAction _ ResultPhaseModeratorEmail =
@@ -255,6 +254,7 @@ createIdea = currentUserAddDb AddIdea
 
 createTopic :: Create Topic
 createTopic = currentUserAddDb AddTopic
+
 
 -- * Vote Handling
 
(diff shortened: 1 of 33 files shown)


fisx:
Random thoughts:

- this branch is based on 2016-03-31-np-dyn-persist (which i want to merge)
- I think this will work, but it's by no means done.
- I have spend most of the time on this with trying to salvage the `RunPersist` type, but ultimately have failed.  I don't think I want to keep trying, even if it's possible.  That would be a follow-up project.
- we will probably have to write our own `makeAcidic` to account for the change to `AQuery`, `AUpdate` from `Query`, `Update`.

fisx:
(rebased.)

src/Persistent/Pure.hs
```diff
@@ -220,94 +233,94 @@ type AddDb m a = UserWithProto a -> PersistM m => m a
...
+    assertAulaDataM $ do
+        len <- lengthOf l <$> askDb id
         when (len /= 1) $ do
             fail $ "Persistent.Api.addDb expects the location (lens, traversal) "
                 <> "to target exactly 1 field not " <> show len
-    a <- fromProto pa <$> nextMetaInfo cUser
-    modifyDb l $ at (a ^. _Id) .~ Just a
+    -- FIXME: reduce code: call 'addDbValue' instead of the following three lines.  (this is another lens puzzle)
+    a :: a <- fromProto pa <$> nextMetaInfo cUser
+    modifyDb_ l $ at (a ^. _Id) .~ Just a
```
np: Actually `addDbValue` needs to be changed for that. Let's keep it like that for now.

src/Persistent/Pure.hs
```diff
@@ -512,5 +522,7 @@ mkMetaInfo cUser now oid = MetaInfo
     , _metaChangedAt       = now
     }
 
-nextMetaInfo :: PersistM m => User -> m (MetaInfo a)
-nextMetaInfo cUser = mkMetaInfo cUser <$> getCurrentTimestamp <*> nextId
+nextMetaInfo :: User -> AUpdate (MetaInfo a)
+nextMetaInfo cUser = mkMetaInfo cUser now <$> nextId
+  where
+    now = undefined  -- TODO: #307
```
np: The `AUpdate` type could have another layer, namely a `Reader Timestamp`.

fisx: Yes, I was hoping for something like that.  But it could be done later in this PR.

src/Action.hs
```diff
@@ -125,12 +125,9 @@ class Monad m => ActionLog m where
 -- @r@ is determined by @m@, because @m@ is intended to be the program's
 -- action monad, so @r@ is just the persistent implementation chosen
 -- to be used in the action monad.
-class (PersistM r, Monad m, MonadError ActionExcept m) => ActionPersist r m | m -> r where
-    -- | Run @Persist@ computation in the action monad.
-    -- Authorization of the action should happen here.
-    -- FIXME: Rename atomically, and only call on
-    -- complex computations.
-    persistent :: r a -> m a
+class (Monad r, Monad m, MonadError ActionExcept m) => ActionPersist r m | m -> r where
```
np: This was the case before but the `Monad m` constraint is not needed here.

src/Action.hs
```diff
@@ -129,6 +130,8 @@ class (Monad r, Monad m, MonadError ActionExcept m) => ActionPersist r m | m ->
     -- | Run "Persistent" computation in the 'Action' monad.
     persistent :: r a -> m a  -- TODO: rename to atomic
 
+class (ActionPersist AQuery m, ActionPersist AUpdate m) => ActionPersistM m
```
np: Using `type ActionPersistM m = (ActionPersist AQuery m, ActionPersist AUpdate m)` is better IMO, in particular this avoids the need for empty instances.

src/Frontend/Page/User.hs
```diff
@@ -96,7 +96,7 @@ userSettings = redirectFormHandler (PageUserSettings <$> currentUser) changeUser
   where
     -- FIXME: Set the password
     changeUser (UserSettingData email _oldPass _newPass1 _newPass2) = do
-        modifyCurrentUser (maybe id (\ e -> userEmail .~ Just e) email)
+        maybe nil (modifyCurrentUser (ModifyUserSetEmail e)) email
```
np: I would prefer to move this `changeUser` function to Persistent.Pure, call it `setUser` and make it an update instead of changing modifyUser.

fisx: I agree that the api could use some grooming.  I'll add this thought as a FIXME.

src/Action.hs
```diff
@@ -171,26 +182,21 @@ currentUserId = userState usUserId >>= \case
...
+currentUserAddDb_ :: (HasAUpdate ev a, ActionPersist m, ActionUserHandler m) =>
+                     (UserWithProto a -> ev) -> Proto a -> m ()
 currentUserAddDb_ addA protoA = void $ currentUserAddDb addA protoA
 
 -- | Returns the current user
 currentUser :: (ActionPersist m, ActionUserHandler m) => m User
 currentUser = do
-    muser <- aquery . findUser =<< currentUserId
+    uid <- currentUserId
+    muser <- aquery (findUser uid)
```
fisx: so you want to catch login errors in the action.  that probably makes sense.

fisx: wait, that's not login errors, it's one level down.  but @andorp explains to me that you are still right: current user should exist in the database, and if it doesn't, it's not persist's fault.

fisx:
@np not sure we really want to get rid of the backwards compat CPP switches, but I'd like it.  If you disagree please revert ed18e1a.

fisx:
closing in favour of #340.

PR #335, opened 2016-04-04
Filter categories by ideas.

fisx:
- filter idea lists by categories
- make idea categories optional.
- fix "edit idea" layout.


src/Frontend/Page/Category.hs
```diff
@@ -0,0 +1,108 @@
...
+{-# LANGUAGE FlexibleContexts           #-}
+{-# LANGUAGE GeneralizedNewtypeDeriving #-}
+{-# LANGUAGE LambdaCase                 #-}
+{-# LANGUAGE OverloadedStrings          #-}
+{-# LANGUAGE ScopedTypeVariables        #-}
+{-# LANGUAGE TypeFamilies               #-}
+
+{-# OPTIONS_GHC -Wall -Werror #-}
+
+module Frontend.Page.Category
```
andorp: It is not a Page module, but a snippet. These should be moved into other module structure.

andorp:
LGTM, next round

PR #336, opened 2016-04-05
Result Phase

andorp:


src/Types.hs
```diff
@@ -160,24 +161,39 @@ data IdeaVoteValue = Yes | No | Neutral
...
+    { _ideaVoteResultMeta   :: MetaInfo IdeaVoteResult
+    , _ideaVoteResultValue  :: IdeaVoteResultValue
+    }
+  deriving (Eq, Ord, Show, Read, Generic)
+
+instance SOP.Generic IdeaVoteResult
+
+data IdeaVoteResultValue
+    = Winning     { _ideaResultCreatorStatement  :: Maybe Document }
+    | EnoughVotes
```
fisx: Should we add a `Bool` here?  Then we can give all ideas that finish the voting phase a result, instead of marking losing ideas with `Nothing`.  I think I'll do that.

fisx: Making this change doesn't produce any type errors.  Good then, I guess.  (-:

src/Persistent/Idiom.hs
```diff
@@ -91,7 +91,7 @@ checkInPhaseJury topic =
 checkAllIdeasMarked :: PersistM m => Topic -> m Bool
 checkAllIdeasMarked topic = all isMarkedIdea <$> findIdeasByTopic topic
```
fisx: Rename this 'Marked' to 'Judged', and the winning-'Marked' as 'MarkedWinning'?  Either way I would like to have two names for these two things.

src/Persistent/Api.hs
```diff
@@ -402,12 +402,12 @@ addCommentVoteToIdeaCommentReply iid cid rid =
...
-instance FromProto IdeaResult where
-    fromProto = flip IdeaResult
+instance FromProto IdeaJuryResult where
+    fromProto = flip IdeaJuryResult
 
-addIdeaResult :: AUID Idea -> AddDb m IdeaResult
-addIdeaResult iid =
-    addDbValue (dbIdeaMap . at iid . _Just . ideaResult . _Just)
+addIdeaJuryResult :: AUID Idea -> AddDb m IdeaJuryResult
+addIdeaJuryResult iid =
```
fisx: *set* instead of *add*?  feel free to ignore for now.

src/Action.hs
```diff
@@ -275,6 +279,20 @@ markIdea iid rv = do
   where
     emptyComputation = return (return ())
 
+-- | Mark idea as winner or not enough votes if the idea is in the Result phase,
+-- if not trhows an exception.
```
fisx: typo.

fisx:
Thanks!

PR #340, opened 2016-04-06
Epic acid-state refactoring.

fisx:
Unsquashed version of this commit: https://github.com/liqd/aula/pull/333


src/Action.hs
```diff
@@ -233,75 +261,74 @@ phaseAction _ ResultPhaseModeratorEmail =
...
-    join . persistent $ do
-        allMarked <- checkAllIdeasMarked topic
-        if allMarked
-            then topicPhaseChange topic =<< AllIdeasAreMarked <$> phaseEndVote
-            else emptyComputation
-  where
-    emptyComputation = return (return ())
+    idea  <- amquery $ findIdea iid
+    topic <- amquery $ ideaTopic idea
+    -- FIXME: should this be one transaction?  or the two above as well?
```
andorp: Every complex database access should run in one transaction.

src/Arbitrary.hs
```diff
@@ -755,18 +754,18 @@ fishAvatars =
...
             Nothing -> error $ "mkFishUser: could not parse avatar url: " <> show avatar
             Just i -> ( UserFirstName $ ST.take i first_last
                       , UserLastName  $ ST.drop (i+1) first_last
                       )
-    role <- Student <$> maybe (persistent genArbitrary) pure mSchoolClass
+    role <- Student <$> maybe genArbitrary pure mSchoolClass
     let pu = ProtoUser Nothing fnam lnam role Nothing Nothing
     -- FIXME: change avatar in the database, not just in the user returned from this function!
-    (userAvatar .~ Just avatar) <$> currentUserAddDb addUser pu
+    (userAvatar ?~ avatar) <$> currentUserAddDb (AddUser (UserPassInitial "streng geheim!")) pu
```
andorp: Is the `"streng geheim!"` intended?

src/DemoData.hs
```diff
@@ -120,58 +124,62 @@ data CommentInContext = CommentInContext
...
-mkUniverse :: forall m . PersistM m => IO (m ())
-mkUniverse = do
-    r <- newQCGen
-    return (universe r)
+mkUniverse :: forall m . ActionM m => IO (m ())
+mkUniverse = universe <$> newQCGen
 
-universe :: QCGen -> forall m . PersistM m => m ()
-universe rnd = void $ do
+universe :: QCGen -> forall m . ActionM m => m ()
```
andorp: This type change will generate a lot of transactions. This should be run in only one. That's why the demo data generation is takes too much time.

src/Persistent/Implementation.hs
```diff
@@ -2,22 +2,29 @@
...
-module Persistent.Implementation (mkRunPersist, withPersist)
+module Persistent.Implementation (mkRunPersist, withPersist, withPersist')
 where
 
-import Control.Lens
 import Config
+import Control.Exception (finally)
+import Control.Lens
+import Data.Monoid
 import Persistent.Api
```
andorp: As far as I see, `Persistent.Api` is removed.

andorp: Ignore this. It is still there, just reduced greatly.

src/Persistent/TemplateHaskell.hs
```diff
@@ -0,0 +1,345 @@
...
+import Data.Acid.Advanced (IsAcidic(acidEvents), Event(..), Method, MethodResult, MethodState)
+
+import Persistent.Pure (AulaData, AUpdate, AddDb, EnvWithProto, PersistExcept, aUpdateEvent)
+
+import Data.List ((\\), nub)
+-- import Data.Maybe (mapMaybe)
+import Data.SafeCopy
+import Data.Typeable
+import Data.Char
+-- import Control.Applicative
```
andorp: Please clean unused imports.

src/Arbitrary.hs
fisx: It's a silly joke.  I assumed it was harmless because it's obvious that any meaning is conincidental and unimportant.

What would you like me to use instead?

src/Persistent/Implementation.hs
fisx: Anyway travis is good at catching those.  I never read the import sections any more.  (-:

src/Arbitrary.hs
andorp: I have no translation or interpretation of that text. I am ok with it. Just asking if it wasn't a leftover.

andorp: It can stay.

src/Persistent/Pure.hs
```diff
@@ -0,0 +1,663 @@
...
+{-# LANGUAGE ViewPatterns                #-}
+
+{-# OPTIONS_GHC -Wall -Werror -fno-warn-orphans #-}
+
+-- | This module exports acid-state transactions.  It introduces a state type 'AulaData' and 'Query'
+-- and 'Update' operations on that state.  (To be more specific, two new types for queries and
+-- pudates are defined that specific to 'AulaData' and introduce exceptions.)
+--
+-- Serializability happens outside of this module.
+--
```
Mikolaj: pudates/updates Serializability/Serialization   :)

src/Action.hs
```diff
@@ -91,47 +102,61 @@ makeLenses ''UserState
...
--- to be used in the action monad.
-class (PersistM r, Monad m, MonadError ActionExcept m) => ActionPersist r m | m -> r where
-    -- | Run @Persist@ computation in the action monad.
-    -- Authorization of the action should happen here.
-    -- FIXME: Rename atomically, and only call on
-    -- complex computations.
-    persistent :: r a -> m a
+-- | A monad that can run acid-state.
+class (MonadError ActionExcept m) => ActionPersist m where
+    aqueryDb :: m AulaData
```
andorp: I feel like the `a` prefix in this class is a leftover from the beginning of the refactor. In my opinion `queryDb`, `query`, `mquery`, `equery`.

src/Action.hs
```diff
@@ -91,47 +102,61 @@ makeLenses ''UserState
...
+    aupdate  :: HasAUpdate ev a => ev -> m a
+
+    aquery :: AQuery a -> m a
+    aquery q = runReader q <$> aqueryDb
```
andorp: `aequery` should be melt into `aquery`. When we run updates in db, exceptions should be caught and thrown in wrapped form.

andorp: I did not see the extra `E` in the type name. I am OK with it. My bad :)

Mikolaj:
LGTM.

fisx:
@np: we're done testing.  we found a lot of small issues in the UI, but nothing that belongs into this PR.

Please merge if you have nothing else to add.

PR #341, opened 2016-04-07
Test commenting on ideas.

andorp:


tests/AulaTests/Stories/Interpreter/Action.hs
```diff
@@ -69,52 +82,119 @@ runClient (Free (SelectIdeaSpace s k)) = do
...
 assert msg False = error $ "assertion failed: " <> show msg
     -- FIXME: give source code location of the call.
+
+shouldBe :: (Monad m, Eq a, Show a) => a -> a -> m ()
+shouldBe actual expected = assert (actual, expected) (actual == expected)
+    -- FIXME: give source code location of the call.
+
+-- ** Notations for test step sections
+
+precondition :: Monad m => m a -> m a
```
andorp: Follows the BDD convention, `given`, `when`, `then` combined with the contract based approach `precondition`, `step`, `postcondition`. At the moment these combinators has no effects at all. The other approach to this separation are add comments all over the place `-- precondtion` etc, which I would like to avoid.

src/Types.hs
```diff
@@ -322,6 +322,11 @@ phaseName = \case
     PhaseVoting     _ -> "Abstimmungsphase"
     PhaseResult       -> "Ergebnisphase"
 
+followingPhase :: Phase -> Phase -> Bool
```
fisx: Rename to `followsPhase`?

fisx: (Or `phaseFollows`?  I think these two are almost exactly equally good.)

andorp: In that case the parameters should be swapped, but I am ok with that version too.

andorp: My first idea was `follows` but that is to general I think.

fisx:
LGTM, thanks!

PR #342, opened 2016-04-07
Email notifications

### LINKED ISSUE #212: sending email
I think we can use thentos for this, too.

Emails need to be send to the idea space moderator for every page that reaches a quorum.  But it would also be ok to resolve this issue for now by adding some test page that sends a one-line email on button push.

np:
Fixes #212.


src/Types.hs
```diff
@@ -77,6 +79,12 @@ type CSI' s a = CSI s s a a
 csi :: CSI s t a b => Iso s t a b
 csi = iso cs cs
 
+-- FIXME: [NP] I'm not quite sure of what this means but at
+-- least this is useful in combination with DF.optionalText
+-- (see Frontend.Page.User) where
```
fisx: you mean you don't understand it yourself?  in that case i would prefer if you could keep it out of the code, and do something more accessible instead.

src/Types.hs
```diff
@@ -788,6 +808,37 @@ instance HasMetaInfo IdeaVote where metaInfo = ideaVoteMeta
...
+{- Examples:
+    e :: EmailAddress
+    s :: ST
+    s = emailAddress # e
+
+    s :: ST
+    s = "foo@example.com"
+    e :: Maybe EmailAddress
+    e = s ^? emailAddress
+-}
```
fisx: those are useful!  thanks!  (-:

src/Action/Smtp.hs
```diff
@@ -0,0 +1,123 @@
...
+makeLenses ''EmailMessage
+
+data SendMailFlag = IgnoreMissingEmails
+    deriving (Eq, Ord, Show, Read, Generic)
+
+class (MonadSendMailError e m, MonadReaderConfig r m) => HasSendMail e r m where
+    sendMailToAddress :: Address -> EmailMessage -> m ()
+
+    default sendMailToAddress :: MonadIO m => Address -> EmailMessage -> m ()
+    sendMailToAddress = sendMailToAddressIO
```
fisx: I factored this trick out for persistence a few days ago, and I would like to not have it here either.

I know that you are proposing to use the concrete `Action` type only in one more place, but it requires more discipline to not just add `MonadIO` somewhere in the API if we can.

I don't understand why the alternative of hiding the relationship between `IO` and `Action` in the implementation is so bad?

fisx:
LGTM.

The only thing I'm worried about is the `prelens` spell (well, and a few others), and the `MonadIO` default instance.  Could you take care of those?  Or should I?

thanks!

src/Action/Smtp.hs
```diff
@@ -0,0 +1,123 @@
...
+checkSendMail :: Config -> IO ()
+checkSendMail cfg = do
+    let address = Address Nothing "user@example.com"
+        msg     = EmailMessage "Test Mail" "This is a test" Nothing
+    r <- runExceptT (runReaderT (sendMailToAddress address msg :: ReaderT Config (ExceptT SendMailError IO) ()) cfg)
+    case r of
+        Left _ -> throwIO $ ErrorCall "sendmail seems to not work.\
+                                    \ Maybe the sendmail path is misconfigured?"
+        Right () -> pure ()
+-}
```
np: Oops, I meant to keep that example/test code alive but I couldn't get rid of the type error quick enough...

fisx:
(Also, a few TODOs.)

src/Action/Smtp.hs
np: * The `Action` type is already in `MonadIO` so I'm using it. Again in the implementation the monad stack clearly has `IO` at the bottom `MonadIO` is here to help with that.
* If we were to remove the `MonadIO` from `Action` which I do not recommend but I'm not strongly opposed to that either, then we could have this one liner definition copied over `Action.Implementation`.
* I read this default implementation first as a hint and second as a shortcut if you have `MonadIO` anyway.

np:
Thanks for the errata. I'd like to keep the prelens as-is for now I waiting for some feedback from #haskell-lens a better name and docs should follow that. About `MonadIO` such a change IMO should be about removing `MonadIO` of `Action`.

fisx:
> Thanks for the errata. I'd like to keep the prelens as-is for now I waiting for some feedback from #haskell-lens a better name and docs should follow that. About `MonadIO` such a change IMO should be about removing `MonadIO` of `Action`.

ok, agreed.

fisx:
After a bit more nit-picking around, these are left to be sorted out IMO:

1. testing.  (i triggered an email sending event, but nothing happened.)
1. prelens
1. dead code in `Action.Smtp` in the comment at the bottom.  please revive or remove.

back to you!  (-:

fisx:
too fast...

fisx:
`/usr/bin/sendmail` does not exist....

fisx:
with the right path, i can receive emails.  \o/

fisx:
Ok, two left:

1. `prelens`
1. dead code in `Action.Smtp` in the comment at the bottom.

fisx:
Thanks!

If travis has no objections, I'd like to merge now.

src/Action.hs
```diff
@@ -280,44 +280,39 @@ sendMailToRole role msg = do
...
-                  ]
-            , _msgHtml = Nothing -- Not supported yet
-            }
-  where
-    topicTemplate = "the topic titled \"" <> t ^. topicTitle <> "\" at URL " <> absPath (listTopicIdeas t)
-    absPath p = "FIXME: http.../" <> absoluteUriPath (relPath p)  -- TODO!
+phaseAction :: (MonadReaderConfig r m, ActionPersist m, ActionSendMail m)
+            => Topic -> PhaseAction -> m ()
+phaseAction topic phaseAction = do
+    cfg <- asks (view getConfig)
```
np: `cfg <- view getConfig` is enough since `view` already works with `MonadReader`. Also if I recall `MonadReaderConfig` exposes `viewConfig`, thus `cfg <- viewConfig`.

exec/Aula.hs
```diff
@@ -8,12 +8,14 @@ import Text.Show.Pretty
...
 import Frontend
 import Config
+import Action.Smtp
 
 
 main :: IO ()
 main = do
     setCurrentDirectoryToAulaRoot
     cfg <- readConfig CrashMissing
+    checkSendMail cfg
```
np: Hum do we want to try send a test email to user@example.com at each run of the server?

np:
LGTM

PR #343, opened 2016-04-07
Combinators to update the DB with meta-data

np:


andorp:
I would like to achieve an abstraction where we can manipulate our dataset in one transaction, in the more compilcated cases like this:

````haskell
markIdeaInJuryPhase :: ActionM m => AUID Idea -> IdeaJuryResultValue -> m ()
markIdeaInJuryPhase iid rv = do
    idea  <- mquery $ findIdea iid
    topic <- mquery $ ideaTopic idea
    -- FIXME: should this be one transaction?  or the two above as well?
    equery $ checkInPhase (PhaseJury ==) idea topic
    currentUserAddDb_ (AddIdeaJuryResult iid) rv
    checkCloseJuryPhase topic
````

np:
Maybe this is a situation where we should do it first and refactor later. What I mean is that we can take an extra parameter `EnvWith ()` or wrap one parameter, `EnvWith IdeaJuryResultValue` in your example. However in your example `checkCloseJuryPhase` should first move to `AUpdate`, in turn `topicPhaseChange` should also be an `AUpdate`, in turn `phaseAction` would be an `AUpdate` as well then here we hit the road: `sendMailToRole`. The solution I see here is log what notification to make. Then the `update` function would trigger the notifications.

andorp:
I like to use the compute a computation, when we collect all the information that we need to assemble the final computation, and run it. This solution is a generalization of `STM` computations returning `IO` computations. Which is the same transactional behavior we need in genereal.

np:
One additional benefit of using a writer monad for notifications is that one can work on these notifications in the background and return early to the user.

andorp:
In one transaction the updated values could be reread from the persistent, if we update everything at the end of the computation we are going to have inconsistent data.

np:
@andorp sorry can you reformulate that last comment.

andorp:
Maybe we are not talking about the same thing. Let me detail my thoughts. In my understanding the writer monad you mentioned would record the necessary action when we run a transaction. Like `ReadX`, `ReadY`, `SaveZ` etc. If we have an operation `ModifyX` and we record that to for the later execution, the following execution trace reproduce an inconsistent data: `ReadX`, `ModifX`, `ReadX` where the second read would read the value from the persistent because it is not modified so far, only a modification action would run at the very end of the computation.

If my assumptions are wrong, let's discuss your idea in a collaborative editor. :)

PR #347, opened 2016-04-08
Random stuff

fisx:


fisx:
I declare this trivial (which is stretching the concept a little, admittedly).  Everybody please raise any concerns you have even after I've merged.

PR #349, opened 2016-04-08
Jury ideas

andorp:

PR #351, opened 2016-04-11
Jury phase pages

fisx:
Copy of #349.

- Ignore aula.yaml and persistented data.
- Run test suite with in memory acid-state implementation.
- Add some paths tests.
- Add simple user context for idea list related pages.
- Add RenderContext.
- Use composite query in juryIdea.
- Use 'ideaCapabilities' in ListIdeaItem rendering.

### DIFF
--- a/src/Frontend/Path.hs
+++ b/src/Frontend/Path.hs
@@ -80,9 +80,11 @@ isPostOnly = \case
     -- FIXME[#312] Logout -> True
     _ -> False
 
--- FIXME: Remove
+-- FIXME: fix & emove
 isBroken :: Main -> Bool
 isBroken Broken = True
+isBroken (IdeaPath _ (JudgeIdea _ _)) = True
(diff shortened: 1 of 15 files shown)


src/Frontend/Path.hs
```diff
@@ -77,9 +80,11 @@ isPostOnly = \case
     -- FIXME[#312] Logout -> True
     _ -> False
 
--- FIXME: Remove
+-- FIXME: fix & emove
```
andorp: `remove` instead of `emove`

src/Types.hs
```diff
@@ -79,6 +79,9 @@ type CSI' s a = CSI s s a a
 csi :: CSI s t a b => Iso s t a b
 csi = iso cs cs
 
+lowerCase :: String -> String
```
andorp: I think `lowerCase` is not used anywhere.

fisx:
thanks!  (-:

since you're here, i let you push the merge button if you're happy (:

andorp:
LGTM

PR #352, opened 2016-04-11
Fixes #308: At most one vote per user

### LINKED ISSUE #308: No more than one like/vote per user on each idea
I've recently changed votes and likes from a `Set` to an `AMap` which keeps `addDb` as the main entry point to add records to the DB.

However I want to argue here that not having IDs for votes and likes could be even nicer. Recall that we do not want more than one like/vote per user&idea. So the votes could be indexed by user id. When one needs to point to a particular vote we can use `(ID Idea, ID User)`.

np:


src/Types.hs
```diff
@@ -555,7 +571,10 @@ data MetaInfo a = MetaInfo
     }
   deriving (Eq, Ord, Show, Read, Generic)
 
-instance SOP.Generic a => SOP.Generic (MetaInfo a)
+instance SOP.Generic auid => SOP.Generic (GMetaInfo a auid)
+
+type MetaInfo a = GMetaInfo a (AUID a)
+type UMetaInfo a = MetaInfo a -- TODO GMetaInfo (AUID User)
```
fisx: call this `RawMetaInfo`?

src/Types.hs
```diff
@@ -517,21 +518,36 @@ defaultSettings = Settings
...
   deriving (Eq, Ord, Show, Read, Generic, FromHttpApiData, Enum, Real, Num, Integral)
 
+instance SOP.Generic (RawAUID a)
+
+type family   AUID a where
+              AUID User             = RawAUID User
+              AUID Idea             = RawAUID Idea
+              AUID Topic            = RawAUID Topic
+              AUID Delegation       = RawAUID Delegation
+              AUID Comment          = RawAUID Comment
```
np:
```
AUID Comment = (RawAUID Idea, NonEmpty (RawAUID Comment))
```

src/Types.hs
fisx: actually, remove.  (:

np:
This fixes #308

static/js/custom.js
```diff
@@ -87,3 +87,8 @@ function removeClass(el, cl) {
 function addClass(el, cl) {
     if(el) el.className = el.className + " " + cl;
 }
+
+function incrCommentVote(e) {
+    var node = e.parentNode.parentNode.childNodes[0];
```
fisx: Nice and simple!  (-:

We can use this as a first approximation, but we will want to have a request that returns the two new values from the server, and a js function that writes them into the right places.  perhaps we can even do that without json/ajax, by violating the form paradigm some more?  (-:

andorp:
I have concerns about the `AUID User` type now have 4 different interpretations. I've got the point that this way it is easy to lookup, if a user did something with an idea related information. I have the gut-feeling that this change a workaround for the `Idea` type in not granulated enough. In my opinion if we want to achieve something similar in the `AMap`s I would introduce a new `Map` types. E.g.

````haskell
type UMap a = Map (AUID User) a

data Idea = Idea
    { _ideaMeta       :: MetaInfo Idea
    , _ideaTitle      :: ST
    , _ideaDesc       :: Document
    , _ideaCategory   :: Maybe Category
    , _ideaLocation   :: IdeaLocation
    , _ideaComments   :: Comments
    , _ideaLikes      :: UMap IdeaLike
    , _ideaVotes      :: UMap IdeaVote
    , _ideaJuryResult :: Maybe IdeaJuryResult  -- invariant: isJust => phase of containing topic > JuryPhsae
    , _ideaVoteResult :: Maybe IdeaVoteResult  -- invariant: isJust => phase of containing topic > VotingPhase
    }

data Comment = Comment
    { _commentMeta    :: MetaInfo Comment
    , _commentText    :: Document
    , _commentVotes   :: UMap CommentVote
    , _commentReplies :: Comments
    }
````

fisx:
But that would make a lot of functions such as e.g. `addDb` unapplicable.

I would like to merge this now and discuss alternatives later, I hope that's ok?

np:
@andorp the `UMap` type you wrote I what I started with but this was indeed breaking `addDb`.
More importantly this would make the identifier for a vote a lot less useful. Indeed what is the purpose of an id, I can see two main points:

- identity: two records might be structurally equal still we want them to be different
- index: an fast access to the data given just the id

Your solution provides identity but not index as we should use the author id to reach the vote.
The index property is a bit broken so far since we need to reach the comment, and to reach the comment we need the idea id. This could be solved by having a global table for each record kind but this would make search request a lot slower if we don't add extra maps. The solution we are going to is to tune `IdOf` to have enough information to achieve the `index` property.

PR #353, opened 2016-04-11
Refactor RenderHtml.

fisx:
With this, we can extract page values from the running system and make them sample pages.  This solves the problem that our arbitrary instances were getting harder to shape in a way that we needed for exposing the parts of the application where the Html code needs work.


src/Frontend/Page/Idea.hs
```diff
@@ -362,10 +362,14 @@ instance FormPage JudgeIdea where
 -- on the bright side, it makes shorter uri paths possible.)
 viewIdea :: (ActionPersist m, MonadError ActionExcept m, ActionUserHandler m)
     => AUID Idea -> m (Frame ViewIdea)
-viewIdea ideaId = makeFrame =<< (do
+viewIdea ideaId = viewIdea' ideaId >>= makeFrame
+
+viewIdea' :: (ActionPersist m, MonadError ActionExcept m, ActionUserHandler m)
```
andorp: What do you tihink about renaming `viewIdea'` to `viewIdeaPage` or `viewIdeaData`?

fisx: Good idea, will do!

fisx:
I've introduced the `please review` label.  If nobody gives me a LGTM, I'll leave the label on and merge it "tentatively".

exec/RenderHtml.hs
```diff
@@ -378,7 +166,7 @@ refreshSamples = do
 runTidyIfAvailable :: FilePath -> IO ()
 runTidyIfAvailable fn' = withTidy >>= (`when` doTidy)
   where
-    fn'' = dropExtensions fn' <.> ".html-tidy.html"
+    fn'' = dropExtension fn' <.> ".tidy"
```
Mikolaj: I see that only my bash completion is fooled by .tidy, other programs, including `file` and `emacs` happily recognize this as a HTML file, despite it not ending in  .html.

fisx: > I see that only my bash completion is fooled by .tidy, other programs, including `file` and `emacs` happily recognize this as a HTML file, despite it not ending in  .html.

which is not even the intended behaviour, since it contains plaintext error messages from tidy, not html.  oh well.  nobody uses this feature anyway.  :)

Mikolaj:
So there is no example SAMPLE data available now, right? But the instructions for creating them are clear enough, I just haven't gotten around to testing them. I assume you did. LGMT.

exec/RenderHtml.hs
Mikolaj: Oh, so that's why I never saw anything in the `.html-tidy.html` files! :)

PR #354, opened 2016-04-12
misc design

fisx:


fisx:
@localgrr don't try to make sense of it!  (-:  I'd like to explain this to you when you have time.

andorp:
* Why the aula.cabal is removed?
 * Why `1.hs` .. `5.hs` are included in the PR?

fisx:
@andorp good questions!  (-:

not sure about aula.cabal, but i'll fix it.  the sample files are there because as of #353, we do not get them from `Arbitrary` any more, but from `Frontend.Testing.RenderHtmlSource`.  the process is that we add those to the PR to indicate which parts of the app need design work, and them remove them before merge.

I like the idea so far, but we can talk about whether you have a better idea if you want.

fisx:
@localgrr please do not update .gitignore any more.  you can either ignore aula.cabal with your own brain, or change .gitignore, but not commit it.  (i'll fix it, no need to do anything.)

fisx:
LGTM.

I'll make a clean PR from this and merge.

Thanks!

localgrr:
Sorry :(

Issue #346, opened 2016-04-08
Introduce Frontend.Snippet module.


andorp:
The following modules are misplaced:

- `Frontend.Page.Comment`
- `Frontend.Page.Category`

These modules should have placed into a different module structure, like `Frontend.Snippet.*`

andorp:
#335

andorp:
The `ListItemIdea` from `Frontend.Core` should be part of some snipplet module.

fisx:
@andorp this can be closed, right?  I'm closing it, please re-open if not.

andorp:
Implemented in #383

PR #355, opened 2016-04-12
Misc fixes.

fisx:


src/Frontend/Page/Idea.hs
```diff
@@ -270,13 +270,12 @@ instance FormPage EditIdea where
...
-    -- FIXME: category choice should look like in CreateIdea.
-    formPage = createOrEditPage True
+    formPage v form p@(EditIdea idea) = createOrEditPage True (idea ^. ideaLocation) v form p
 
 createOrEditPage :: (Monad m, Typeable page, Page page) =>
-    Bool -> View (HtmlT m ()) -> (HtmlT m () -> HtmlT m ()) -> page -> HtmlT m ()
-createOrEditPage showDeleteButton v form p = semanticDiv p $ do
+    Bool -> IdeaLocation ->
+    View (HtmlT m ()) -> (HtmlT m () -> HtmlT m ()) -> page -> HtmlT m ()
+createOrEditPage showDeleteButton cancelUrl v form p = semanticDiv p $ do
```
andorp: Is the `showDeleteButton` could be computed from `renderContext` via extending the `ideaCapabilities` description?

fisx: yes, good idea.  will do!

fisx: Actually, i changed my mind.  `showDeleteButton` is set by one of the two callers of this, which are the two `ToHtml` instances of `EditIdea`, `CreateIdea` resp.

So this is not really about the session context, but about which of two similar forms we happen to render.

Does that make sense?

andorp: Yes. On the other hand I am agree with Connor McBride, we should introduce different types for these parameters, even if at the beginning it is isomorphic to `Bool`. But I can live with the current solution.

src/Frontend/Page/Topic.hs
```diff
@@ -23,11 +23,12 @@ where
 import Action (ActionM, ActionPersist(..), ActionUserHandler, ActionExcept, renderContext)
 import Control.Exception (assert)
 import Frontend.Page.Category
+import Frontend.Page.Overview
```
andorp: One page module should not include other one.We have an issue #346 connected to this change. But for the time being it is ok.

andorp:
Two minor comments, otherwise LGTM

PR #357, opened 2016-04-12
d3 new force graph, some basic page layout

localgrr:
This is a very basic layout until I get my head around exactly whats required, but we are now using the very fancy force graph I found. If we can try and hammer our data in to it all the better.


fisx:
@localgrr: I can't find a reference to the code you use in this PR.  Could you please look it up, post it here, ping rittermo, and then we should resolve this before we talk about the contents.

my delegation page showed some changes in the menus, but still the old code.  but that can be resolved when we have the license.

thanks!

fisx:
this PR contains code copyrighted here: http://bl.ocks.org/GerHobbelt/3104394.  license negotiations pending.

fisx:
update on license negotations:

email from ger:

```
> Datum: 16. April 2016 um 21:22:04 MESZ
> All d3 work of mine is free - by default it has the same license AS the d3 library itself.
```

d3js license is here: https://github.com/mbostock/d3/blob/master/LICENSE

fisx:
obsoleted by #712, #710, see also: #714.

Issue #83, opened 2016-02-22
Redirect 303 to page that redirects to "Your something has created" page that redirects too.


andorp:
From https://github.com/liqd/aula/pull/56#issuecomment-185281700 :

> This should probably go to some US page, but let me just write it down here: On saving a newly created topic, we will probably have to show a page saying "your topic has been created", which redirects on button click or with a delay of 3 seconds.

We have to implement this requirement to avoid the [Double Submit problem](http://www.theserverside.com/news/1365146/Redirect-After-Post) which occurs when there is no redirect 303 after POST requests. We need to create (lets call) ACK pages, but in the following way:

For example:

 1. GET request on "/ideas/create"
 2. User adds content, submits POST
 3. POST request on "/ideas/create"
 4. Redirect 303 to "/ideas/create-ack"
 5. User's browser renders the ACK page which redirects to "/ideas" page after 3 seconds (redirect in HTML header)

andorp:
The Mantra
    
    PRG pattern can be rephrased like this:
    
        Never show pages in response to POST
        Always load pages using GET
        Navigate from POST to GET using REDIRECT

np:
Could redirect directly to "/ideas" but with a small banner message saying "Your idea has been created".

fisx:
I guess the idea is that if you show the post page even for three seconds, the user can try to reload it, and will thus repost.

This is more of a usability issue: with @andorp's way, the user will never have to deal with annoying browser popups saying that reload is not recommended in this situation.

andorp:
The problem is that we have an exact requirement on this. At least we had last week.

andorp:
@fisx: It not just a usability issue. If we allow such behaviour, users will do repost, and we had to come up some logic which has a heuristic on data similarity as a change request from the client. __We do not want to go that direction__ believe me.

andorp:
The redirection page from the old click dummy: https://marvelapp.com/ehhb43#10293398
The new does not have it, but I don't know the actual requirement set.

np:
I didn't meant to show the post page but to redirect. However instead of redirecting twice, I suggest to redirect only once.

andorp:
@np. I got it, and this is the way it works now. But as far as I know that contradicts for the actual requirement set we have.

fisx:
> @fisx: It not just a usability issue. If we allow such behaviour, users will do repost, and we had to come up some logic which has a heuristic on data similarity as a change request from the client. __We do not want to go that direction__ believe me.

[...]

PR #358, opened 2016-04-12
Misc fixes.

fisx:


fisx:
(There is a lot of stuff in here that i want to release now, but I'll leave the `please review` label on.)

src/Frontend/Page/Overview.hs
```diff
@@ -174,26 +174,24 @@ instance ToHtml Tabs where
...
   deriving (Eq, Show, Read, Generic)
 
 data ListItemIdea = ListItemIdea
-      { _listItemIdeaContext    :: ListItemIdeaContext
-      , _listItemIdeaPhase      :: Maybe Phase
-      , _listItemIdeaNumVoters  :: Int
-      , _listItemIdea           :: Idea
-      , _listItemRenderContext  :: RenderContext
+      { _listItemRenderContext  :: RenderContext
+      , _listItemIdeaContext    :: ListItemIdeaContext
```
Mikolaj: Perhaps Position? There is already ` data IdeaLocation`. Also, we don't want to mix it up with Space.

src/Persistent/Pure.hs
```diff
@@ -245,6 +246,9 @@ runPersistExcept :: PersistExcept -> ServantErr
 runPersistExcept (PersistError500 msg)            = err500 { errBody = cs msg }
 runPersistExcept (PersistError404 msg)            = err404 { errBody = cs msg }
 runPersistExcept (PersistErrorNotImplemented msg) = err500 { errBody = cs msg }
+runPersistExcept (UserLoginInUse li) =
+    err500 { errBody = "user login in use: " <> cs (show li) }
+    -- FIXME: what's a good status code for 'login in use'?
```
Mikolaj: 200? USers are know to reload pages and go back, etc. Otherwise, perhaps 403:  https://www.w3.org/Protocols/rfc2616/rfc2616-sec10.html#sec10.4.4

src/Action.hs
```diff
@@ -353,18 +353,19 @@ voteIdeaCommentReply ideaId commentId replyId =
...
     idea  <- mquery $ findIdea iid
     topic <- mquery $ ideaTopic idea
-    -- FIXME: should this be one transaction?  or the two above as well?
     equery $ checkInPhase (PhaseJury ==) idea topic
     currentUserAddDb_ (AddIdeaJuryResult iid) rv
     checkCloseJuryPhase topic
 
 checkCloseJuryPhase :: ActionM m => Topic -> m ()
 checkCloseJuryPhase topic = do
+    -- FIXME: should this be one transaction?
```
Mikolaj: I think so, and the same above. I think an alternative is to check (in the operations above that modify the DB, internally, necessarily within a single transaction with the update) that the current values are as expected, and if not abort with an error like "the ideal is not in the expected phase".

Mikolaj:
LGTM, though I don't understand some of the details.

src/Frontend/Page/Overview.hs
fisx: > Perhaps Position? There is already ` data IdeaLocation`. Also, we don't want to mix it up with Space.

you're right, and both `location` and `position` are still very general and thus not very meaningful.  `WhatList`?

i'm really fond of the `WhoWhen` type, but i'm not sure if anybody can share this enthusiasm.  :)

dada!

src/Persistent/Pure.hs
fisx: On Tue, Apr 12, 2016 at 06:02:27PM -0700, Mikolaj Konarski wrote:
> Date: Tue, 12 Apr 2016 18:02:27 -0700
> From: Mikolaj Konarski <notifications@github.com>
> To: liqd/aula <aula@noreply.github.com>
> Cc: fisx <mf@zerobuzz.net>
> Subject: Re: [liqd/aula] Misc fixes. (#358)
>
> > @@ -245,6 +246,9 @@ runPersistExcept :: PersistExcept -> ServantErr
> >  runPersistExcept (PersistError500 msg)            = err500 { errBody = cs msg }
> >  runPersistExcept (PersistError404 msg)            = err404 { errBody = cs msg }
> >  runPersistExcept (PersistErrorNotImplemented msg) = err500 { errBody = cs msg }
> > +runPersistExcept (UserLoginInUse li) =
> > +    err500 { errBody = "user login in use: " <> cs (show li) }
> > +    -- FIXME: what's a good status code for 'login in use'?
>
> 200? USers are know to reload pages and go back, etc. Otherwise, perhaps 403:  https://www.w3.org/Protocols/rfc2616/rfc2616-sec10.html#sec10.4.4

we should make sure that page reload on forms does not happen, and i think we do (see #83).  so 403 is probably the right one.  will fix, thanks!

static/js/custom.js
```diff
@@ -88,7 +87,10 @@ function addClass(el, cl) {
...
 
-function incrCommentVote(e) {
-    var node = e.parentNode.parentNode.childNodes[0];
-    node.data = Number(node.data) + 1;
+function handleLikeOrVote(e) {
+    // FIXME: this is a race condition: if we wait for 0 ms, the page
+    // will usually be reloaded before the POST request updating the
+    // score can be processed.
+    // FIXME: we are required to avoid reload!
+    setTimeout(function() { document.location.reload(true); }, 50);
```
np: @fisx why did you removed the previous version?

fisx: > @fisx why did you removed the previous version?

sorry (:

my solution wasn't as successful as i hoped, but it has the nice effect on chrome that it looks almost like it works, even when clicking erratically on both many times.  your solution was not supporting the case where the user changed their mind.

anyway we'll need something more sophisticated.  probably involving ajax.

PR #360, opened 2016-04-13
Misc.

fisx:


src/Frontend/Page/Overview.hs
```diff
@@ -175,6 +175,7 @@ instance ToHtml Tabs where
 -- * idea lists
 
 data ListItemIdeaContext  -- TODO: rename to 'ListLocation'?  'Context' is used for too many things.
+                          -- 'Position'?  i think my favorite so far would be `WhatList`.  ~~mf
```
andorp: This type is only for pass the informatin where the ListItemIdea values are generated, we have only one function which converts the `ListItemIdea` value to `Html`, but the rendering depend on the caller, so it is context dependentent, that's why it is called `ListItemIdeaContext`. If we interpret location as a location where the idea is listed that's ok, but it has nothing to do with the position of the idea in the list of ideas.

andorp: Other solutions:
 * Use 3 different functions for the very similar task
 * Use 3 different types for the very similar task

And abstract out the common HTML generation parts.

src/Persistent/Pure.hs
```diff
@@ -425,6 +427,23 @@ findUser = findInById dbUserMap
 getUsers :: Query [User]
 getUsers = view dbUsers
 
+getUsersInClass :: SchoolClass -> Query [User]
+getUsersInClass clss = filter (isClassInRole clss . view userRole) <$> view dbUsers
```
andorp: There is a similar query:
https://github.com/liqd/aula/blob/master/src/Persistent/Idiom.hs#L34

src/Frontend/Page/Admin.hs
```diff
@@ -44,6 +45,8 @@ data PageAdminSettingsQuorum =
 
 instance Page PageAdminSettingsQuorum
 
+-- FIXME: the following names are a little ridiculous.  s/PageAdminSettingsGaPUsersView/GroupPermUserView/?9
```
andorp: We could drop the `PageZZZ` preffixes from the other page types as well. Wdyt?

src/Arbitrary.hs
```diff
@@ -345,6 +345,7 @@ instance Arbitrary UserSettingData where
 instance Arbitrary RenderContext where
     arbitrary = RenderContext <$> arbitrary
 
+
```
andorp: #361

src/Frontend/Page/Admin.hs
fisx: > We could drop the `PageZZZ` preffixes from the other page types as well. Wdyt?

ok, fine with me.  i'm actually in favour of adding them to all page types consistently. but i don't think we have to vote on that, since i already know @np's opinion on this.  :)

src/Persistent/Pure.hs
fisx: you can move this one there if you want, but you'll have to do it yourself.  (-:

src/Frontend/Page/Overview.hs
fisx: nah, i like the type, i just want to find a more awesome name.   oh, just noticed it's a TODO.

hum.

I'll rename it to `WhatListPage` and see if anybody will try to hurt me.

andorp: For me `WhatListPage` is a question. :) This is the first time in my life, I see a question as a type name. Which is kind of make sense, if we see this as an information about the context, which is represented as an extra paramter, which is isomorph to a pair. Also we push this info from a call, like the inversion of control from the client code to the function call, where inversion could mean the dual of the function call, which basically a dual of a function which is a pair. Does it hurt? :D

Summary: I kind of like it.

src/Persistent/Pure.hs
andorp: Maybe next week. :)

andorp: Probably a code or abstraction duplication is in the air. Please check and if that is the case, unify the two functions.

fisx:
erm...  any idea how to solve the hlint complaint?  I tried a few things in HLint.hs, but without success.

src/Types.hs
```diff
@@ -1010,3 +1010,27 @@ countIdeaVotes v = countEq v ideaVoteValue
...
+  | EventLogUserDelegates         User Delegation User
+  | EventLogTopicMovesFromTo      Phase Phase EventTriggeredBy
+  | EventLogIdeaMovesToTopic      User Idea Topic
+  | EventLogIdeaWins              User Idea
+  deriving (Eq, Ord, Show, Read)
+
+data EventTriggeredBy = EventTriggeredBy User | EventTriggeredByTimeout | EventTriggeredByAllIdeasMarked
+  deriving (Eq, Ord, Show, Read)
+
+data Thing = TopicThing Topic | IdeaThing Idea | CommentThing Comment
```
andorp: I don't get the name `Thing` :) Could you rename it to a more descriptive name, or add a comment?

andorp:
No idea, maybe it is an issue in HLint?

src/Types.hs
np: `Record` ?

src/Frontend/Testing.hs
```diff
@@ -73,10 +74,12 @@ type RenderHtmlSource' =
        "idea"  :> Capture "iid" (AUID Idea)  :> Get '[PlainText] String
   :<|> "topic" :> Capture "tid" (AUID Topic) :> Get '[PlainText] String
```
np: I know it's unrelated to your PR but it's Misc...

These captures can be `Idea ::>` and `Topic ::>`

src/Frontend/Page/Admin.hs
```diff
@@ -592,7 +609,13 @@ instance Csv.FromRecord CsvUserRecord where
             | otherwise        = Just . UserLogin $ v !! i
 
 instance Csv.ToRecord CsvUserRecord where
-    toRecord (CsvUserRecord fn ln em li pw) = Csv.toRecord (show fn, show ln, show em, show li, show pw)
+    toRecord (CsvUserRecord fn ln em li pw) = Csv.toRecord
+        (          _fromUserFirstName            fn
```
np: `fn ^. _UserFirstName` should do.

src/Frontend/Page/Admin.hs
```diff
@@ -592,7 +609,13 @@ instance Csv.FromRecord CsvUserRecord where
...
+        ,          _fromUserLastName             ln
+        , maybe "" (show . internalEmailAddress) em
```
np: `emailAddress # em`

src/Frontend/Page/Admin.hs
```diff
@@ -592,7 +609,13 @@ instance Csv.FromRecord CsvUserRecord where
...
+        , maybe "" _fromUserLogin                li
+        , maybe "" id                            pw
```
np: `pw ^. _Just`

This works because when `^.` is applied to a `Prism` it goes down to a `Fold` and get this type: `(^.) :: Monoid a => s -> Fold s a -> a`. In our case `(^.) :: Maybe ST -> Fold (Maybe ST) ST -> ST` which means that the monoid instance for ST is doing the work of defaulting to the empty string.

PR #362, opened 2016-04-13
Delete and report comments

np:


src/Frontend/Path.hs
```diff
@@ -14,10 +14,11 @@ module Frontend.Path
     , UserPs(..)
     , AdminPs(..)
     , IdeaMode(..)
+    , CommentMode(..)
     , viewIdea, editIdea, commentIdea, createIdea, listIdeas, listTopicIdeas
     , likeIdea, voteIdea, judgeIdea
     , voteCommentIdea, voteCommentIdeaReply, voteCommentWithContext
-    , replyCommentIdea, commentOrReplyIdea, isPostOnly, isBroken
+    , replyCommentIdea, commentOrReplyIdea, isPostOnly, isBroken, onComment
```
andorp: It seems that the `onComment` is not used outside of the module, also seems it is just the abstract common part of some others. It shouldn't be in the export list, or am I missing something?

src/Persistent/Pure.hs
```diff
@@ -476,6 +480,26 @@ findIdeasByTopic = findAllInBy dbIdeas ideaLocation . topicIdeaLocation
...
 
+findComment :: AUID Idea -> Maybe (AUID Comment) -> AUID Comment -> MQuery Comment
+findComment iid mparentid cid = preview $ dbIdeaMap . at iid . _Just . ideaComments . l
+  where
+    l = case mparentid of
+            Nothing       -> at cid . _Just
+            Just parentid -> at parentid . _Just . commentReplies . at cid . _Just
+
+-- This function should become useless once the comment ids are complete
+findComment' :: AUID Idea -> Maybe (AUID Comment) -> AUID Comment -> EQuery (Idea, Maybe Comment, Comment)
```
andorp: If this is a temporary hack, please rename it to `findCommentHack`.

src/Frontend/Path.hs
np: Your missing the next commits, I preferred using onComment for delete and report instead of export 2 (or 4) more functions.

andorp: Good point.

src/Frontend/Page/Comment.hs
```diff
@@ -18,37 +20,62 @@
...
+            when (CanComment `elem` w ^. cwIdeaCaps && CanReplyComment `elem` comCaps) .
+                button_ [class_ "btn comment-footer-button", onclick_ $ U.replyCommentIdea idea parent] $ do
+                    i_ [class_ "icon-reply"] nil
+                    "antworten"
+            postButton_ [class_ "btn comment-footer-button"]
+                        (U.onComment idea mparent comment U.ReportComment) $ do
                 i_ [class_ "icon-flag"] nil
                 "melden"
+            when (CanDeleteComment `elem` comCaps) .
+                postButton_ [class_ "btn comment-footer-button", Lucid.onclick_ "handleDeleteComment(this)"]
```
andorp: I am not familiar with the trick here. Does the button click run the `JavaScript` and also send `POST` request to the server?

static/js/custom.js
```diff
@@ -94,3 +94,7 @@ function handleLikeOrVote(e) {
     // FIXME: we are required to avoid reload!
     setTimeout(function() { document.location.reload(true); }, 50);
 }
+
+function handleDeleteComment(e) {
```
andorp: Related question: #363

tests/Frontend/CoreSpec.hs
```diff
@@ -54,7 +54,7 @@ spec = do
         , H (arb :: Gen PageStaticImprint)
         , H (arb :: Gen PageStaticTermsOfUse)
         , H (arb :: Gen PageAdminSettingsGaPClassesEdit)
-        , H (CommentWidget <$> arb <*> arb)
+        , H (arb :: Gen CommentWidget)
```
andorp: Thanks!

andorp:
@np: There is no handler for some path defined. Please take a look at travis.

np:
Should be ready now. Keep commenting on!

tests/Frontend/PathSpec.hs
```diff
@@ -33,6 +35,19 @@ import Network.Wai
 import Test.Hspec.Wai (get, post, shouldRespondWith)
 import qualified Test.Hspec.Wai.QuickCheck as Wai (property)
 
+isBrokenPath :: Main -> Bool
```
andorp: Can you add some explanation why the random values here broke the test cases? Which assumption is not hold?

andorp: Excluding non-working test cases is a bad practice. I can accept your decision if you provide more information. Based on that, we could make better test data generators.

andorp:
LGTM

andorp:
@fisx Please review and merge if you are ok with this PR.

tests/Frontend/PathSpec.hs
np: For the report handler, I fetched the comment from the DB, since the path has been randomly generated so the different parts are not tied together. This should be fixed once we have a complete comment id.

src/Frontend/Page/Comment.hs
np: It does.

src/Action.hs
```diff
@@ -347,6 +353,55 @@ voteIdeaCommentReply :: AUID Idea -> AUID Comment -> AUID Comment -> Create_ Com
...
+deleteIdeaCommentReply :: AUID Idea -> AUID Comment -> AUID Comment -> ActionPersist m => m ()
+deleteIdeaCommentReply ideaId commentId replyId =
+    update $ DeleteCommentReply ideaId commentId replyId
+
+-- FIXME:
+-- One might want to check that the parameters corresponds to an existing comment.
+-- More generally do we do anything to prevent abuse of the report system?
+-- One thing could be log the event or count the reports made by one user.
+-- Since no record of the report are kept in base not only multiple users can
+-- report the same comment but the same user can report multiple times.
```
fisx: good points, we'll definitely want to fix that once we are feature-complete.

src/Frontend/Core.hs
```diff
@@ -307,11 +308,11 @@ instance Show a => ToHtml (PageShow a) where
     toHtmlRaw = toHtml
     toHtml = pre_ . code_ . toHtml . ppShow . _unPageShow
 
-data CommentVotesWidget = VotesWidget CommentContext Comment
+data CommentVotesWidget = VotesWidget [IdeaCapability] CommentContext Comment
```
fisx: So far we've created the idea caps lists locally in the ToHtml instances, now we are beginning to pass them around.  Perhaps we should consistently keep them in the RenderContext?  It doesn't alway apply there, though.  Should RenderContext become polymorphic in the page type then?

Probably best to ponder this some more before we change anything.

fisx:
LGTM.

I had a few nit-picks, please take a look and merge.  Thanks!

src/Frontend/Core.hs
np: I was not aware of the RenderContext.

np:
Passing down the rendering context is better indeed.

src/Frontend/Page/Comment.hs
```diff
@@ -29,11 +29,10 @@ import LifeCycle
 import qualified Frontend.Path as U
 
 data CommentWidget = CommentWidget
-    { _cwIdeaCaps :: [IdeaCapability]
-    , _cwUserId   :: AUID User
-    , _cwRole     :: Role
-    , _cwContext  :: CommentContext
-    , _cwComment  :: Comment
+    { _cwIdeaCaps      :: [IdeaCapability]
```
andorp: There is a type synonym for `[IdeaCapability]` that is `IdeaCapabilities` please use that, as the same style is applied on the persistent layer.

fisx: > There is a type synonym for `[IdeaCapability]` that is `IdeaCapabilities` please use that, as the same style is applied on the persistent layer.

no there isn't, i think i might have removed that.  i don't like introducing more names for something os simple.

andorp:
There are paths without handlers.

src/Frontend/Page/Comment.hs
andorp: Accepted.

tests/Frontend/PathSpec.hs
andorp: As far as I know, there is no DB read involved in the `Paths and handlers` test suite, it uses the `mockAulaMain` uses only arbitrary instances to generate some mock data for the handler. I am opening an issue about it.

andorp: Issue in: https://github.com/liqd/aula/issues/365

fisx: > > +isBrokenPath :: Main -> Bool
>
> As far as I know, there is no DB read involved in the `Paths and handlers` test suite, it uses the `mockAulaMain` uses only arbitrary instances to generate some mock data for the handler. I am opening an issue about it.

i really like this test, almost as much as i dislike our generic Arbitrary instances.  i think the problem in this case was that arbitrary generates urls to comments of arbitrary nesting depth (not maximum 2), but similar problems keep kreeping up.

you are right, though: we're not using the database state here, and i'm not sure how we would take an AulaData value generated based on stories and generated arbitrary valid paths from that.  nor that that would be the same test.  i'll give this another go.

testing is exhausting!  :/

fisx:
well, that wasn't exactly fun, but interesting.  i think i fixed PathSpec!  (-:

@andorp @np could you please review the trick in e7e5784.  if there are no more objections, i'd like to merge

thanks!

fisx:
(had to rebase to get rid of some temporary commits, sorry!)

andorp:
The problem comes when we introduce assumptions silently, and that was the case with the test suite. We overgenerated test cases, which were not valid any more, we have two choices here, limit the test data generation, or limit our abstraction. I also would reformulate the IdeaPath contruction. But for the time being it is ok.

src/Frontend/Page/Comment.hs
np: I'm in favor of those simple names but I won't fight either.

np:
Ack.

PR #367, opened 2016-04-15
Comments identifiers now contains context: idea and parents.

np:


src/Frontend/Path.hs
```diff
@@ -174,35 +176,39 @@ listTopicIdeas :: Topic -> Main
...
-    i = idea ^. _Id
-    base =
-        case mc of
-            Nothing -> root </> "idea" </> uriPart i </> "comment" </> uriPart c
-            Just p  -> root </> "idea" </> uriPart i </> "comment" </> uriPart (p ^. _Id) </> "reply" </> uriPart c
+    -- FIXME: Deep replies are not supported yet.
+    -- Meanwhile urls are automatically shortened to fit the current API.
+    -- In particular voting/deleting/reporting can only apply up to depth 2
+    -- and replying up to depth 1.
+    base n =
```
andorp: I like the explanation in the comment. Thanks for the clarification.

src/Types.hs
```diff
@@ -273,6 +282,25 @@ data Comment = Comment
 
 instance SOP.Generic Comment
 
+-- This is the complete information to recover a comment in AulaData
+data CommentId = CommentId
+    { _cidIdeaLocation  :: IdeaLocation
+    , _cidIdea          :: AUID Idea
+    , _cidParents       :: [AUID Comment]
```
andorp: Please add an explanation, how we should interpret the list of the parents.

src/Types.hs
```diff
@@ -551,20 +579,20 @@ type instance IdOf User             = AUID User
...
--- become: (IdOf Comment, AUID User)
+type instance IdOf Comment          = CommentId
+type instance IdOf CommentVote      = CommentVoteId
+type instance IdOf IdeaVote         = IdeaVoteLikeId
+type instance IdOf IdeaLike         = IdeaVoteLikeId
 type instance IdOf IdeaVoteResult   = AUID IdeaVoteResult
 type instance IdOf IdeaJuryResult   = AUID IdeaJuryResult
 
-type AMap a = Map (IdOf a) a
+type family   AUIDOf a
```
andorp: As `CommentId` and `IdeaVoteLikeId`, `CommentVoteId` are composite keys for some information in the database. `AUIDOf` should be called `AKey` or something similar. I am not sure what the `A` preffix stands for in the `AUID`.

np: The A stands for Aula. The status of these keys/ids has to be discussed since these types both evolved too far and far enough such that we start to see how to clean and improve. The distinction I want to make is between what serves as the key for final map in which the record is located and the parent keys needed to traverse the whole structure from the top.

I think that the first kind of key can be set to always be an `AUID` and the type of parent keys can be driven with a type family similar to what we have now.

andorp: I like the idea of having this type family. In my opinion the name `AKey` is more descriptive, than `AUIDOf` becuase the ID is the most simple version of a primary key, so if we take steps towards complex keys, we should call them as keys, even if in some cases that means an AUID.

src/Persistent/Pure.hs
```diff
@@ -499,39 +496,28 @@ findIdeasByTopic = findAllInBy dbIdeas ideaLocation . topicIdeaLocation
...
-            pure (idea, Nothing, comment)
-        Just parentid -> do
-            parent <- maybe404 (idea ^. ideaComments . at parentid)
-            comment <- maybe404 $ parent ^. commentReplies . at cid
-            pure (idea, Just parent, comment)
+findComment' :: AUID Idea -> [AUID Comment] -> AUID Comment -> MQuery Comment
+findComment' iid parents = preview . dbComment' iid parents
+
+findComment :: CommentId -> MQuery Comment
+findComment cid = findComment' (cid ^. cidIdea) (cid ^. cidParents) (cid ^. cidAUID)
```
fisx: nice (-:

src/Frontend/Path.hs
```diff
@@ -174,35 +176,39 @@ listTopicIdeas :: Topic -> Main
...
+        ReportComment -> base 2 </> "report"
+        VoteComment v -> base 2 </> "vote" </> uriPart v
+        ViewComment   -> root  </> "idea" </> uriPart i </> "view" </#> commentAnchor commentId
   where
-    i = idea ^. _Id
-    base =
-        case mc of
-            Nothing -> root </> "idea" </> uriPart i </> "comment" </> uriPart c
-            Just p  -> root </> "idea" </> uriPart i </> "comment" </> uriPart (p ^. _Id) </> "reply" </> uriPart c
+    -- FIXME: Deep replies are not supported yet.
```
fisx: If "deep replies" means "comment hierarchies of depth >2": Leave the sentence, but drop the FIXME?  We are not supposed to support that.

fisx:
LGTM.

@np, if you have made the changes we discussed (and considered the trivial comment I made), you can merge.

Thanks!

tests/AulaTests/Stories/Interpreter/Action.hs
```diff
@@ -164,10 +164,10 @@ runClient (Free (CommentOnComment t cp c k)) = do
 -- * helpers
 
 findIdeaByTitle :: (ActionM m) => IdeaTitle -> ActionClient m (Maybe Idea)
-findIdeaByTitle t = fmap (find ((t ==) . view ideaTitle)) . lift $ query getIdeas
+findIdeaByTitle t = lift $ query (findIdeaBy ideaTitle t)
```
fisx: yes, this is nicer.

src/Types.hs
```diff
@@ -1071,6 +1084,8 @@ countIdeaVotes v = countEq v ideaVoteValue
 countCommentVotes :: UpDown -> CommentVotes -> Int
 countCommentVotes v = countEq v commentVoteValue
 
+-- Given a list of parents and a collection (AMap) of comments
+-- returns the collection of comments after following the parenting list.
 traverseParents :: [AUID Comment] -> Traversal' Comments Comments
```
fisx: I am skeptical of plural type synonyms such as `Comments`, but since it hasn't been introduced here, we don't have to remove it here.  Just saying.

PR #371, opened 2016-04-18
Add FormPageHandler type.

andorp:
The `FormPageHandler` helps testing, in this way the implementation for Actions won't diverge in the Action interpreter for the test DSL, in the case of form pages.

### DIFF
--- a/src/Types.hs
+++ b/src/Types.hs
@@ -99,8 +99,8 @@ data Either3 a b c = Left3 a | Middle3 b | Right3 c
 
 instance (SOP.Generic a, SOP.Generic b, SOP.Generic c) => SOP.Generic (Either3 a b c)
 
-app2 :: (c -> d) -> (a -> b -> c) -> (a -> b -> d)
-app2 f g = \x y -> f $ g x y
+app2 :: (c -> d) -> (a -> b -> c) -> a -> b -> d
+app2 f g x y = f $ g x y
 
 -- * prototypes for types
 
--- a/src/Types.hs
+++ b/src/Types.hs
@@ -99,6 +99,8 @@ data Either3 a b c = Left3 a | Middle3 b | Right3 c
 
 instance (SOP.Generic a, SOP.Generic b, SOP.Generic c) => SOP.Generic (Either3 a b c)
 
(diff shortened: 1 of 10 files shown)


src/Types.hs
```diff
@@ -99,6 +99,8 @@ data Either3 a b c = Left3 a | Middle3 b | Right3 c
 
 instance (SOP.Generic a, SOP.Generic b, SOP.Generic c) => SOP.Generic (Either3 a b c)
 
+app2 :: (c -> d) -> (a -> b -> c) -> (a -> b -> d)
+app2 f g = \x y -> f $ g x y
```
fisx: What about `form0`, `form1`, `form2`, ... instead?

src/Types.hs
```diff
@@ -243,6 +243,10 @@ data IdeaJuryResultValue
 
 type instance Proto IdeaJuryResult = IdeaJuryResultValue
 
+ideaJuryResultValueToType :: IdeaJuryResultValue -> IdeaJuryResultType
+ideaJuryResultValueToType NotFeasible{} = IdeaNotFeasible
+ideaJuryResultValueToType Feasible{}    = IdeaFeasible
```
fisx: `Value` should be capitalized.

andorp: It is capitalized.

fisx: if you only need this function in the story interpreter, would it make sense to define it there?

fisx:
LGTM.  Some nit-picks.  Please check if there is anything you want to act on.

src/Types.hs
andorp: We have the base case as `form`, and only tha case of `app2 form (f x y)`, I think intoroducing form2 wouldn't be better. The read the `app2` is the following: "apply an arbitrary function to the result of ...", instead of introducing a `form2` with a comment on top of it, that this one waits for a function with two arguments functions, or even further we could introduce a typeclass to form this abstraction, but this only occurs in the `Frontend` module, I wouldn't complicate things.

fisx: ok, fair enough.

src/Types.hs
fisx: never mind.  this is fine.

fisx:
Thanks!  (-:

PR #372, opened 2016-04-18
User settings form redirects to itself.

fisx:
not very exciting, but we were asked to fix that this week.  so, there!  (-:


andorp:
LGTM

PR #373, opened 2016-04-18
Improving isPrivatePage and makeFrame

np:


src/Frontend/Core.hs
```diff
@@ -191,7 +191,7 @@ class Page p => FormPage p where
 
 -- | Defines some properties for pages
 class Page p where
-    isPrivatePage :: p -> Bool
+    isPrivatePage :: proxy p -> Bool
```
andorp: Maybe I miss the point, but why not `Proxy p` ?

np: This is strictly more general than `Proxy` since `isPrivatePage (Proxy :: Proxy a)` will still work. So is `isPrivatePage [a]` (where `a` is a value here). So is `isPrivatePage mp` where `mp` is of type `m p`.

andorp: Ok Agreed, nice trick!

fisx:
[irc]

```
18:42 < npou> 373 can be merged if you want otherwise I might add another
              commit to move most makeFrame calls to Frontend.hs
18:45 < andorp_> We are waiting for your commit probably
```

Mikolaj:
So far so good.

src/Frontend/Core.hs
```diff
@@ -207,20 +207,21 @@ instance Page ST where
...
     extraPageHeaders = extraPageHeaders . view frameBody
 
 makeFrame :: (ActionPersist m, ActionUserHandler m, MonadError ActionExcept m, Page p)
           => p -> m (Frame p)
 makeFrame p = do
   isli <- isLoggedIn
-  if | not isli && isPrivatePage p -> redirect . absoluteUriPath $ relPath P.Login
-     | isli     || isPrivatePage p -> flip Frame p <$> currentUser
-     | otherwise                   -> return $ PublicFrame p
+  let isPrivate = isPrivatePage [p] -- Here '[]' is used as the 'proxy'.
```
fisx: heh.  pretty!  (-:

fisx:
LGTM, thanks!

PR #375, opened 2016-04-19
Refine js click handlers.

fisx:
- reload on clicks and jump to the right place on the page.
- provide for some minimum type safety.
- fix UI issue.

### DIFF
--- a/src/Frontend/Path.hs
+++ b/src/Frontend/Path.hs
@@ -188,7 +188,7 @@ ideaMode (OnComment ck m) root = commentMode ck m root
 ideaMode CreateIdea        root = root </> "idea" </> "create"
 
 anchor :: IsString s => AUID a -> s
-anchor (AUID c) = fromString $ "comment-" <> show c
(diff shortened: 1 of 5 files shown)


fisx:
This is ready for review and merging (I think).

src/Frontend/Path.hs
```diff
@@ -187,8 +187,8 @@ ideaMode (CommentIdea i)   root = root </> "idea" </> uriPart i </> "comment"
 ideaMode (OnComment ck m) root = commentMode ck m root
 ideaMode CreateIdea        root = root </> "idea" </> "create"
 
-commentAnchor :: IsString s => AUID Comment -> s
-commentAnchor (AUID c) = fromString $ "comment-" <> show c
+anchor :: IsString s => AUID a -> s
+anchor (AUID c) = fromString $ "comment-" <> show c
```
np: What about "auid-" instead of "comment-" since this is no longer specific to comments.

np:
1 small remark. LGTM

PR #376, opened 2016-04-19
More test steps.

andorp:


tests/AulaTests/Stories/Tests.hs
```diff
@@ -18,14 +18,22 @@ import AulaTests.Stories.DSL
...
+    createIdea idea1a "desc" CatRules
+    editIdea idea1a idea1 "desc1" CatRules
     likeIdea idea1
-    commentIdea idea1 "This is a comment"
-    commentOnComment idea1 "This is a comment" "This is a comment1"
-    createTopic idea1 topic1 "desc"
+    commentIdea idea1 comment1
+    commentOnComment idea1 comment1 comment2
+    voteOnComment idea1 comment1 Up
+    voteOnCommentComment idea1 comment1 comment2 Up
```
fisx: in other parts of our code base, we call that "comment reply".  would it make sense to adjust to that here, too?

andorp: Yep, I'll do it.

tests/AulaTests/Stories/DSL.hs
```diff
@@ -58,6 +59,9 @@ selectIdeaSpace n = liftF $ SelectIdeaSpace n ()
 createIdea :: IdeaTitle -> IdeaDescription -> Category -> Behavior ()
 createIdea title desc cat = liftF $ CreateIdea title desc cat ()
 
+editIdea :: IdeaTitle -> IdeaTitle -> IdeaDescription -> Category -> Behavior ()
+editIdea oldTitle newTitle desc cat = liftF $ EditIdea oldTitle newTitle desc cat ()
```
fisx: what if there are two ideas with the same title?

andorp: It depends on the interpretation. The actual implementation picks the first one for the `Action` interpreter. The test suite has to designed the way, to avoid that collusion and test against them on a different level.

fisx: Ok, that sounds like it was intentional.  Good enough.  (-:

fisx:
LGTM.  If you are done with the renaming, this can be merged.

Thanks!

tests/AulaTests/Stories/Interpreter/Action.hs
```diff
@@ -239,7 +239,7 @@ findCommentCommentByText c t = find ((t ==) . fromMarkdown . _commentText) . Map
 
 findIdeaAndComment :: (ActionM m) => IdeaTitle -> CommentText -> ActionClient m (Maybe (Idea, Maybe Comment))
 findIdeaAndComment it cp =
-    fmap (fmap (\idea -> (idea, findCommentByText idea cp))) (findIdeaByTitle it)
+    fmap (fmap (id &&& flip findCommentByText cp)) (findIdeaByTitle it)
```
fisx: Yes, I saw that and was so happy you'd missed it...  ;-)

PR #374, opened 2016-04-18
improvements to the RenderHtml process

fisx:
This requires liqd/thentos#506 to land first, and for aula to bump its thentos dependency.

PR #378, opened 2016-04-19
improvements to the RenderHtml process

fisx:
Replaces #374.


fisx:
As mentioned in #374, I will merge this without consent from travis.  Testing locally instead.

PR #377, opened 2016-04-19
User profile improvements

np:

PR #381, opened 2016-04-19
Jury buttons on view idea page

andorp:
Fixes: https://github.com/liqd/aula/issues/346

 * Introduces the `Frontend.Page.Snippet` module.
 * Move snippet like functionality into the introduced module.
 * Buttons for jury ideas is reused in list of ideas and on view idea page.


andorp:
It leaves untouched the `Frontend.Page.Cagetory` module. I am not sure if we should merge that module to the snippets.

src/Frontend/Page/Idea.hs
```diff
@@ -154,6 +153,8 @@ instance ToHtml ViewIdea where
             div_ [class_ "sub-heading"] $ do
                 toHtml $ IdeaVoteLikeBars caps p
 
+            Snippet.juryIdea True idea caps
```
fisx: Please rename.  `feasibilityVerdict`?  `feasibilityWidget`?

fisx:
Rename `Frontend.Page.Snippet` to `Frontend.PageFragment`?

fisx:
I know #346 has been open a while and I should have given feedback earlier, but I'm not sure I see the point.  What's wrong with keeping a mix of html building blocks under page that are sometimes an entire page and sometimes only a part?

I agree that it could be easier to find things (except there is `git grep`), but I'm not sure making the distinction between fragments and full pages helps with that.  Shouldn't we draw lines between application concepts?

fisx:
The fact that you didn't merge the `Category` module with `Snippets` also indicates that this module structure is not very convenient.

What about having application-based modules under `Frontend.Page.*`, but naming some of them `*Fragment` and some `*`?

fisx:
more thoughts:

- come to think of it, i still prefer not making the distinction between snippets and pages at all.
- is the idea list still showing which ideas have been marked, and how?
- this may be material enough for two independent PRs.

Mikolaj:
I'm strongly of the Algebraic Specifications school, which says modules should be built around types (to make the types more and more abstract, ideally, by providing more and more operations and exposing less and less implementation). With bigger programs, types are often in separate modules, but still modules should group operations on a given type, or a group of interrelated types. So, as long as we don't have a type that distinguishes a completely constructed and sealed page from a partial page,  I don't see a reason to make Snippet a separate module. OTOH, I see the point of keeping operations related to Comments and related to idea lists separate, because they are related to separate types.

fisx:
Thanks @Mikolaj, that makes a lot of sense.  @andorp and I just fought this out a little more, and the consensus we reached is pretty close to what you are saying:

- introduce `Frontend.Fragment.*`.
- move `Frontend.Page.{Comment,Category}` to fragments.
- move resp. fragments from `Frontend.Page.Overview` to `Frontend.Fragment.IdeaList`.
- create new fragment module `Feasibility` for the changes that gave this PR its name.

Have I missed anything?  I'm ok with leaving it at this, we can always do more cleanup work later.

We will deal with #377 now and then do this again, since it's probably easier than rebasing.

PR #382, opened 2016-04-19
Rename from to un in types

andorp:
Fixes: https://github.com/liqd/aula/issues/315

### DIFF
--- a/src/Arbitrary.hs
+++ b/src/Arbitrary.hs
@@ -747,7 +747,7 @@ instance Aeson.ToJSON D3DN where
             ]
 
         renderNode n = object
-            [ "name"   .= (n ^. userLogin . fromUserLogin)
+            [ "name"   .= (n ^. userLogin . unUserLogin)
             , "avatar" .= (n ^. userAvatar)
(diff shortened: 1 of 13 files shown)


andorp:
I am not sure if the second commit introduce more readability, feel free to revert.

fisx:
Not sure if there is some morphological rule according to which it sometimes has to be `from` and sometimes `un`, but I like the consistency.

There is also `fromPhase`, `fromProto`.  On the other hand, we won't change `fromList`, `fromMaybe`, `fromEnum`, `fromString`, ...

If nobody has any objections, I'll merge this as is and let #315 close itself.

Thanks!

andorp:
The `fromProto` is differs from the ADT selectors. We are making some other value of a different type with some additional functionality. I would leave it as it is.

andorp:
The `fromPhase` is in a different context, where it indicates a phase transition, in that context the `from` is the valid naming.

fisx:
Oops, I wasn't really thinking.  You're right, merging.

PR #384, opened 2016-04-20
Default emails in config.

fisx:


src/DemoData.hs
```diff
@@ -152,6 +153,20 @@ genCommentVote comments_in_context students = do
 updateAvatar :: User -> URL -> forall m . ActionM m => m ()
 updateAvatar user url = update $ SetUserAvatar (user ^. _Id) url
 
+addUserWithEmailFromConfig :: Proto User -> forall m . ActionM m => m User
+addUserWithEmailFromConfig puser =
+    setEmailFromConfig puser >>= currentUserAddDb AddUser
```
andorp:
````haskell
addUserWithEmailFromConfig puser = setEmailFromConfig puser >>= currentUserAddDb AddUser
````
can be written as:
````haskell
addUserWithEmailFromConfig = setEmailFromConfig >=> currentUserAddDb AddUser
````

File: src/Frontend/Page/Login.hs [hunk 2]
```diff
@@ -11,11 +11,12 @@ where
 import Text.Digestive
 
 import Action (ActionM, query)
-import Persistent
 import qualified Action
+import Persistent
 import Frontend.Prelude
 
 import qualified Frontend.Path as U
+import qualified Lucid
```
andorp: Why `Lucid` is imported?

fisx: If you are trying to ask "why the qualified?": `href_` is also defined in `Lucid.Missing`, but I need the original one here, that takes a string url instead of a `Path`.

andorp:
LGTM

andorp:
You can merge if travis agrees.

PR #385, opened 2016-04-20
Mark ideas as feasible on the view idea page.

andorp:
As discussed in: #381


src/Frontend/Fragment/Feasibility.hs
```diff
@@ -0,0 +1,49 @@
+{-# LANGUAGE OverloadedStrings   #-}
+{-# LANGUAGE RankNTypes          #-}
+
+{-# OPTIONS_GHC -Werror -Wall #-}
+
+-- | This module contains snippets which are used all over the pages.
```
fisx: is this comment intentional, or a left-over?  if the former, please keep it!  (-:

fisx:
LGTM!

I tried to be fancy, but got stuck.  I will wait for travis and merge.

Thanks!

PR #386, opened 2016-04-20
Sorting idea lists

fisx:
- Includes some cleanup effort in `Frontend.Core`.  First 4 commits do not contain any actual change in functionality.  Just moving stuff around.

### DIFF
--- a/src/Frontend.hs
+++ b/src/Frontend.hs
@@ -283,7 +283,7 @@ topicApi space
 aulaSpace :: ActionM m => IdeaSpace -> ServerT AulaSpace m
 aulaSpace space
     =  ideaApi (IdeaLocationSpace space)
-  :<|> (\qf qs -> makeFrame $ Page.viewIdeas space (qf, qs))
+  :<|> uncurry . makeFrame . Page.viewIdeas space
   :<|> topicApi                   space
 
 type AulaUser =
--- a/src/Frontend.hs
+++ b/src/Frontend.hs
@@ -283,7 +283,7 @@ topicApi space
 aulaSpace :: ActionM m => IdeaSpace -> ServerT AulaSpace m
 aulaSpace space
     =  ideaApi (IdeaLocationSpace space)
-  :<|> uncurry . makeFrame . Page.viewIdeas space
+  :<|> curry (makeFrame . Page.viewIdeas space)
   :<|> topicApi                   space
 
 type AulaUser =
--- a/src/Frontend.hs
+++ b/src/Frontend.hs
@@ -235,10 +235,10 @@ type TopicApi =
        -- view topic details (tabs "Alle Ideen", ..., "Beauftragte Stimmen")
 
        -- view topic details (tabs "Alle Ideen", "Beauftragte Stimmen")
-  :<|> Topic ::> "ideas"              :> IdeasFilterApi :> GetH (Frame ViewTopic)
-  :<|> Topic ::> "ideas" :> "all"     :> IdeasFilterApi :> GetH (Frame ViewTopic)
-  :<|> Topic ::> "ideas" :> "voting"  :> IdeasFilterApi :> GetH (Frame ViewTopic)
-  :<|> Topic ::> "ideas" :> "winning" :> IdeasFilterApi :> GetH (Frame ViewTopic)
+  :<|> Topic ::> "ideas"              :> IdeasFilterApi :> IdeasSortApi :> GetH (Frame ViewTopic)
+  :<|> Topic ::> "ideas" :> "all"     :> IdeasFilterApi :> IdeasSortApi :> GetH (Frame ViewTopic)
+  :<|> Topic ::> "ideas" :> "voting"  :> IdeasFilterApi :> IdeasSortApi :> GetH (Frame ViewTopic)
+  :<|> Topic ::> "ideas" :> "winning" :> IdeasFilterApi :> IdeasSortApi :> GetH (Frame ViewTopic)
   :<|> Topic ::> "delegations"        :> GetH (Frame ViewTopic)
 
        -- create new topic
@@ -249,7 +249,7 @@ type TopicApi =
 type AulaSpace
     =  IdeaApi
        -- browse wild ideas in an idea space
-  :<|> "ideas" :> IdeasFilterApi :> GetH (Frame PageIdeasOverview)
+  :<|> "ideas" :> IdeasFilterApi :> IdeasSortApi :> GetH (Frame PageIdeasOverview)
   :<|> TopicApi
 
 ideaApi :: ActionM m => IdeaLocation -> ServerT IdeaApi m
@@ -278,12 +278,12 @@ topicApi space
   :<|> form . Page.editTopic
   :<|> error "api not implemented: topic/:topic/delegation/create"
   where
-    viewTopicTab tab tid qf = makeFrame $ Page.viewTopic (tab qf) tid
+    viewTopicTab tab tid qf qs = makeFrame $ Page.viewTopic (tab (qf, qs)) tid
 
 aulaSpace :: ActionM m => IdeaSpace -> ServerT AulaSpace m
 aulaSpace space
     =  ideaApi (IdeaLocationSpace space)
-  :<|> makeFrame . Page.viewIdeas space
+  :<|> (\qf qs -> makeFrame $ Page.viewIdeas space (qf, qs))
   :<|> topicApi                   space
 
 type AulaUser =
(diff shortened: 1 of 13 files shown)


fisx:
Status after giving this PR 2h30m:

- some TODOs
- manual testing not unsuccessful (sort menu selection has no effect).

fisx:
...  that just leaves the TODOs.

fisx:
This is ready for review, and I'd be ok with merging it and sorting out the TODOs in a separate PR.

I will keep working on this until further notice.

fisx:
This is good to land, I think.

src/Frontend.hs
```diff
@@ -278,12 +278,12 @@ topicApi space
...
   :<|> error "api not implemented: topic/:topic/delegation/create"
   where
-    viewTopicTab tab tid qf = makeFrame $ Page.viewTopic (tab qf) tid
+    viewTopicTab tab tid qf qs = makeFrame $ Page.viewTopic (tab (qf, qs)) tid
 
 aulaSpace :: ActionM m => IdeaSpace -> ServerT AulaSpace m
 aulaSpace space
     =  ideaApi (IdeaLocationSpace space)
-  :<|> makeFrame . Page.viewIdeas space
+  :<|> (\qf qs -> makeFrame $ Page.viewIdeas space (qf, qs))
```
andorp: Please use `uncurry`.

src/Frontend/Core.hs
```diff
@@ -207,31 +287,78 @@ class Page p => FormPage p where
...
+  where
+    getPage = _formGetPage formHandler
+    processor = _formProcessor formHandler
 
-instance (Page a, Page b) => Page (Beside a b) where
-    isPrivatePage _ = isPrivatePage (Proxy :: Proxy a) || isPrivatePage (Proxy :: Proxy b)
-    extraPageHeaders (Beside a b) = extraPageHeaders a <> extraPageHeaders b
+    guard page = do
+        r <- guardPage page
+        maybe (pure ()) (redirect . absoluteUriPath) r
```
andorp: Please use `<$>`

src/Types.hs
```diff
@@ -102,6 +104,13 @@ instance (SOP.Generic a, SOP.Generic b, SOP.Generic c) => SOP.Generic (Either3 a
 app2 :: (c -> d) -> (a -> b -> c) -> a -> b -> d
 app2 f g x y = f $ g x y
 
+sortOn :: Ord b => Getter a b -> [a] -> [a]
+sortOn l = sortBy (compare `on` view l)
```
andorp: \o/

src/Types.hs
```diff
@@ -102,6 +104,13 @@ instance (SOP.Generic a, SOP.Generic b, SOP.Generic c) => SOP.Generic (Either3 a
...
+
+sortOnDesc :: Ord b => Getter a b -> [a] -> [a]
+sortOnDesc l = reverse . sortOn l
```
andorp: Please indicate the fact for the reverse order.

andorp:
Few comments, later there will be more. :)

src/Frontend/Core.hs
fisx: > Please use `<$>`

i tried.  can you show me how?

src/Types.hs
fisx: > Please indicate the fact for the reverse order.

what do you mean?  the fact that the order is reversed is indicated by the `Desc` in the function name.

andorp: I misread it to `sort on description`.

fisx: > I misread it to `sort on description`.

ah.  do you have a better name?

andorp: Not really. Let's leave it as it is.

src/Frontend/Core.hs
```diff
@@ -454,6 +458,51 @@ ideasFilterQuery = \case
...
+
+instance ToHttpApiData SortIdeasBy where
+    toUrlPiece = \case
+        SortIdeasByAge     -> "age"
+        SortIdeasBySupport -> "sup"
+
+ideasSortQuery :: IdeasSortQuery -> [Idea] -> [Idea]
+ideasSortQuery = f . fromMaybe minBound
+  where
+    f SortIdeasByAge     = age
```
np: Why not `age . sup` then?

src/Frontend/Core.hs
```diff
@@ -454,6 +458,51 @@ ideasFilterQuery = \case
...
+    f SortIdeasBySupport = sup . age
+
+    age = sortOn $ ideaLikes . to length
+    sup = sortOn createdAt
+
+type IdeasQuery = (IdeasFilterQuery, IdeasSortQuery)
```
np: A data-type would have been nice here.

src/Frontend/Core.hs
```diff
@@ -454,6 +458,51 @@ ideasFilterQuery = \case
...
+
+ideasRunQuery :: IdeasQuery -> [Idea] -> [Idea]
+ideasRunQuery (f, s) = ideasSortQuery s . ideasFilterQuery f
+
+listIdeasWithQuery :: IdeaLocation -> IdeasQuery -> URL
+listIdeasWithQuery loc (qf, qs) = (absoluteUriPath . relPath . P.listIdeas $ loc)
+                               <> renderBoth (catMaybes [renderFilter <$> qf, renderSort <$> qs])
+  where
+    renderFilter :: Category -> ST
+    renderFilter v = "category=" <> toUrlPiece v
```
np: Can't we reuse the fact that we already declared "category" as being the name of this query param? I can't find anything obvious.

src/Types.hs
```diff
@@ -107,6 +107,9 @@ app2 f g x y = f $ g x y
 sortOn :: Ord b => Getter a b -> [a] -> [a]
 sortOn l = sortBy (compare `on` view l)
 
+sortOnDesc :: Ord b => Getter a b -> [a] -> [a]
+sortOnDesc l = reverse . sortOn l
```
np: Can you rename Desc to Down as this is the convention in Data.Ord?

And then this is: `downSortOn l = sortOn (l . to Down)`

src/Frontend/Page/Topic.hs
```diff
@@ -279,9 +264,12 @@ viewTopic tab topicId = do
                 pure $ ViewTopicDelegations topic delegations
             _ ->
               do
-                let q = fromMaybe (Nothing, Nothing) $ tab ^? viewTopicTabQuery
-                ideas <- ideasRunQuery q <$> findIdeasByTopic topic
-                ideasAndNumVoters <- ListItemIdeas ctx q <$> (getListInfoForIdea `mapM` ideas)
+                let loc = topicIdeaLocation topic
+                    ideasQuery = fromMaybe (assert False $ error "viewTopic: impossible.")
+                               $ tab ^? viewTopicTabQuery
```
np: You can use: `tab ^?! viewTopicTabQuery`

src/Frontend/Core.hs
fisx: > Why not `age . sup` then?

i was thinking that, too, but i don't think there will ever be two ideas created in the same milisecond.  and if they are, i can live with the fact that their order is not specified.

src/Frontend/Core.hs
fisx: > A data-type would have been nice here.

good point.  i'll give that a shot.

src/Frontend/Core.hs
np: You can't, since `redirect` is not pure.
An explicit bind is shorter: `maybe (pure ()) (redirect . absoluteUriPath) =<< guardPage page`.

Then the `maybe (pure ())` is `mapM_` from `Data.Traversable`:

`mapM_ (redirect . absoluteUriPath) =<< guardPage page`

src/Types.hs
np: I've mentioned elsewhere that haskell uses `Down` in `Data.Ord`.

np:
I'm done commenting.

src/Frontend/Core.hs
fisx: > Can't we reuse the fact that we already declared "category" as being the name of this query param? I can't find anything obvious.

me neither.  if you have no idea, i don't know how to act on this.

src/Frontend/Page/Topic.hs
fisx: > You can use: `tab ^?! viewTopicTabQuery`

yes, but that gives the same bad error messages as fromJust.  i'd rather like to keep this noise.  (unless you write enough tests to convince me we'll never run into that problem ;-)

PR #388, opened 2016-04-21
Fix: event log tests.

fisx:


andorp:
LGTM

src/Arbitrary.hs
```diff
@@ -804,7 +807,9 @@ instance Aeson.ToJSON D3DN where
 -- * event log
 
 instance Arbitrary EventLog where
-    arbitrary = garbitrary
+    arbitrary = EventLog <$> arbWord <*> nonEmpty
+      where
+        nonEmpty = (:) <$> garbitrary <*> garbitrary
```
np: these two `garbitrary` should be `arbitrary` instead.

np:
Done commenting.

src/Arbitrary.hs
fisx: > these two `garbitrary` should be `arbitrary` instead.

that doesn't work, since both generate separate arguments to `(:)`.  `Couldn't match type ‘[a0] -> [a0]’ with ‘[EventLogItem]’`.

np: my guess is that this is an instance of "let should not be generalized". Adding a type sig should fix it.

PR #390, opened 2016-04-21
Enforce a newer cassava package and eliminate the overlapping instance

Mikolaj:


fisx:
LGTM.

I guess I should learn to do it like this to begin with, and not sprinkle tuples all over the code.

Thanks!

PR #395, opened 2016-04-21
Improved sticky footer and delete user

localgrr:
Hooope the sticky footer update doesnt cause any horrid layout problems its getting a bit hacky to the point it might need a re-think.

Delete user page looks OK - please use this layout for any further "confirm" pages


fisx:
Thanks!

PR #396, opened 2016-04-21
Event Log download form.

fisx:


fisx:
Self-contained, ready for review.  Still missing: recording of real events.

src/Frontend.hs
```diff
@@ -310,13 +310,13 @@ type AulaAdmin =
...
        -- event log
-  :<|> "event"  :> GetH (Frame PageAdminSettingsEventsProtocol)
+  :<|> "event"  :> FormHandler PageAdminSettingsEventsProtocol
   :<|> "passwords" :> Capture "schoolclass" SchoolClass :> Get '[CSV] InitialPasswordsCsvH
   :<|> "events" :> Get '[CSV] EventLog
   :<|> "events" :> Capture "space" IdeaSpace :> Get '[CSV] EventLog
 
 
-aulaAdmin :: ActionM m => ServerT AulaAdmin m
+aulaAdmin :: forall m. ActionM m => ServerT AulaAdmin m
```
np: There is no use of `m` in the body here.

src/Frontend/Core.hs
```diff
@@ -274,7 +274,14 @@ class Page p => FormPage p where
...
     formAction :: p -> P.Main
-    -- | Calculates a redirect address from the given page
+    -- | Calculates a redirect address from the given page.
+    --
+    -- Currently, 'FormPage' forces you to redirect after a form has been processed successfully.
+    -- This may not always be what you want.  For instance, a `preview` button of a text field
+    -- containing markdown does not change the state of the server, and multiple re-submissions are
+    -- a feature, not a problem.  For those cases, we may need to extend this part of the API in the
+    -- future.  We could make the result type a @Maybe a@, where 'Nothing' means redirect, and
+    -- 'Just' means not redirect.  Or we could add a more flexible variant of the 'FormPage' class.
```
np: [Context #398 and #83]

We could also say that these extra submissions are not using FormH but PostH and GetH. In your example of the preview button, my view is that it should not be a post, even though I see the point of using the form data.

np:
LGTM

PR #397, opened 2016-04-21
Do not show deleted users.

### LINKED ISSUE #191: Delete user
There is a button on the edit user page, which deletes the user.

 * The user deletion behavior is not defined.
 * There are no design for the user deletion page.

andorp:
Fixes #191.

### DIFF
--- a/src/Action.hs
+++ b/src/Action.hs
@@ -75,7 +75,7 @@ module Action
 where
 
 import Control.Lens
-import Control.Monad (void, when)
+import Control.Monad ((>=>), void, when)
 import Control.Monad.Reader (runReader, runReaderT)
 import Control.Monad.Except (MonadError, throwError)
 import Control.Monad.Trans.Except (runExcept)
@@ -213,13 +213,17 @@ type ActionError m = (MonadError ActionExcept m)
 
 -- * User Handling
 
+checkActiveUser :: ActionUserHandler m => User -> m User
+checkActiveUser user =
+    if isDeletedUser user
+        then logout >> throwError500 "Unknown user identitifer"
+        else pure user
+
 loginByUser :: ActionUserHandler m => User -> m ()
-loginByUser = login . view _Id
+loginByUser = checkActiveUser >=> login . view _Id
 
 loginByName :: (ActionPersist m, ActionUserHandler m) => UserLogin -> m ()
-loginByName n = do
-    Just u <- query $ findUserByLogin n  -- FIXME: handle 'Nothing'
-    loginByUser u
+loginByName u = loginByUser =<< mquery (findUserByLogin u)
 
 -- | Returns the current user ID
 currentUserId :: ActionUserHandler m => m (AUID User)
@@ -251,13 +255,15 @@ currentUser = do
     uid <- currentUserId
     muser <- query (findUser uid)
     case muser of
-        Just user -> pure user
-        Nothing   -> logout >> throwError500 "Unknown user identitifer"
+        Just user | isActiveUser user
+            -> pure user
+        _   -> logout >> throwError500 "Unknown user identitifer"
 
 -- | Modify the current user.
 modifyCurrentUser :: (ActionPersist m, ActionUserHandler m, HasAUpdate ev a)
                   => (AUID User -> ev) -> m a
-modifyCurrentUser ev = currentUserId >>= update . ev
+modifyCurrentUser ev =
+    currentUser >>= checkActiveUser >>= update . ev . view _Id
 
 isLoggedIn :: ActionUserHandler m => m Bool
 isLoggedIn = userState $ to validLoggedIn
--- a/src/Frontend/Page/Admin.hs
+++ b/src/Frontend/Page/Admin.hs
@@ -309,7 +309,7 @@ instance ToHtml AdminViewUsers where
                     renderUserRow (DeletedUser user) = tr_ $ do
                         td_ nil
                         renderUserInfoRow user
-                        td_ "deleted"
+                        td_ "[gelöscht]"
 
                     renderUserRow (ActiveUser user) = tr_ $ do
                         td_ . span_ [class_ "img-container"] $ avatarImgFromMaybeURL (user ^. userAvatar)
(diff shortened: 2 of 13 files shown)


andorp:
I am going to squash the commits.

src/Frontend/Page/Admin.hs
```diff
@@ -292,8 +292,19 @@ instance ToHtml PageAdminSettingsGaPUsersView where
...
-                let renderUserRow :: forall m. (Monad m) => User -> HtmlT m ()
-                    renderUserRow user = tr_ $ do
+                let renderUserRow :: forall m. (Monad m) => UserView -> HtmlT m ()
+                    renderUserRow (DeletedUser user) = tr_ $ do
+                        td_ $ nil
+                        td_ $ user ^. userLogin . unUserLogin . html
+                        td_ (case user ^. userRole of
+                                Student cl    -> toHtml $ showSchoolClass cl
+                                ClassGuest cl -> toHtml $ showSchoolClass cl
+                                _             -> nil)
```
fisx: We have `ideaSpaceToSchoolClass` in `Types`.  This here could be a similar helper.  I thought we already had one, but I can't find it any more.

fisx: (There is no reason to that in this PR, except that I just noticed now.)

src/Frontend/Page/Admin.hs
```diff
@@ -292,16 +292,25 @@ instance ToHtml PageAdminSettingsGaPUsersView where
...
                         td_ (case user ^. userRole of
                                 Student cl    -> toHtml $ showSchoolClass cl
                                 ClassGuest cl -> toHtml $ showSchoolClass cl
                                 _             -> nil)
                         td_ $ roleLabel (user ^. userRole)
                         td_ (toHtmlRaw nbsp)
+
+                let renderUserRow :: forall m. (Monad m) => UserView -> HtmlT m ()
+                    renderUserRow (DeletedUser user) = tr_ $ do
+                        td_ nil
```
fisx: Why not show the avatar of deleted users?  (Just curious, we can keep it this way.)

src/Types.hs
```diff
@@ -105,6 +106,10 @@ instance (SOP.Generic a, SOP.Generic b, SOP.Generic c) => SOP.Generic (Either3 a
 app2 :: (c -> d) -> (a -> b -> c) -> a -> b -> d
 app2 f g x y = f $ g x y
 
+-- TODO: Check for application in the codebase
+fmap2 :: (Functor f, Functor g) => (a -> b) -> f (g a) -> f (g b)
+fmap2 f = fmap (fmap f)
```
fisx: We already have `Data.Functor.Infix.(<$$>)`.

src/Persistent/Idiom.hs
```diff
@@ -127,3 +127,10 @@ setTopicPhase tid phase = modifyTopic tid $ topicPhase .~ phase
 deactivateUser :: AUID User -> AUpdate ()
 deactivateUser uid
     = modifyUser uid (userSettings . userSettingsPassword .~ UserPassDeactivated)
+
+getUserViews :: Query [UserView]
+getUserViews = fmap2 makeUserView getAllUsers
+
+findActiveUser :: AUID User -> EQuery User
+findActiveUser =
+    (maybe404 . (^? activeUser) . makeUserView) <=< maybe404 <=< findUser
```
fisx: Why not define this in `MQuery`?  It seems to me like `maybe404` should only be called inside the application logic, when an `MQuery` is available but an `EQuery` is needed.

src/Action.hs
```diff
@@ -219,7 +219,9 @@ loginByUser = login . view _Id
 loginByName :: (ActionPersist m, ActionUserHandler m) => UserLogin -> m ()
 loginByName n = do
     Just u <- query $ findUserByLogin n  -- FIXME: handle 'Nothing'
-    loginByUser u
+    if isDeletedUser u
+        then logout >> throwError500 "Unknown user identitifer"
+        else loginByUser u
```
fisx: We need to check this for `login`, `loginByUser` as well.  (I wonder if there is a better way to implement these three in terms of each other?)

src/Action.hs
```diff
@@ -251,9 +253,11 @@ currentUser = do
     uid <- currentUserId
     muser <- query (findUser uid)
     case muser of
-        Just user -> pure user
-        Nothing   -> logout >> throwError500 "Unknown user identitifer"
+        Just user | isActiveUser user
+            -> pure user
+        _   -> logout >> throwError500 "Unknown user identitifer"
 
+-- TODO: Check if the current user is not deleted.
```
fisx: if the current user is deleted, something else has gone wrong!  (-:  So if you check it again here, it should be obvious that if the check fails it's an internal error.

src/Persistent/Idiom.hs
andorp: I would like to separate cases, but we don't have different exception for each cases, yet.

fisx:
LGTM.  If you have processed all my comments and the remaining TODOs, you can merge.

src/Action.hs
andorp: Race condition:

 * Somebody changes her profile
 * The admin deletes the user

I see cases when the user gets undeleted by the modification action.

src/Action.hs
andorp: The `login` does not depend on the persistent, and `loginByUser` calls to login, let me tweak a little bit.

src/Frontend/Page/Admin.hs
andorp: The difference on the UI was more pleasant to the eye.

src/Frontend/Page/Admin.hs
np: You can use the roleSchoolClass lens: `td_ $ user ^. userRole . roleSchoolClass . to showSchoolClass . html`

src/Action.hs
fisx: > I see cases when the user gets undeleted by the modification action.

good point!

np:
LGTM

andorp:
Squashed.

fisx:
Thanks!

PR #402, opened 2016-04-22
partly freeze platform for holidays (story 24)

Mikolaj:
https://tree.taiga.io/project/evpa-aula/us/24


src/Frontend/Page/Topic.hs
```diff
@@ -3,6 +3,7 @@
 {-# LANGUAGE FlexibleInstances   #-}
 {-# LANGUAGE LambdaCase          #-}
 {-# LANGUAGE OverloadedStrings   #-}
+{-# LANGUAGE RecordWildCards     #-}
```
fisx: I really should keep the coding policy more updated, but I don't, so technically you can do this.  (-:

But I prefer not to have this extension in the code base.  I think it makes the code harder to read and less rigorous during refactor, and the only advantage I can see is that you don't have to enter the names that you're interested in in the pattern.

src/Frontend/Page/Topic.hs
```diff
@@ -60,7 +61,10 @@ instance Page ViewTopic where
     extraBodyClasses _ = ["m-shadow"]
 
 -- | 10.1 Create topic: Create topic
-data CreateTopic = CreateTopic IdeaSpace [Idea] Timestamp
+data CreateTopic = CreateTopic
+    { _createTopicIdeaSpace   :: IdeaSpace
+    , _createTopicIdeas       :: [Idea]
+    , _createTopicRefPhaseEnd :: Timestamp }
```
fisx: This change I like!

src/LifeCycle.hs
```diff
@@ -20,7 +21,7 @@ import Types
 data PhaseChange
     = RefinementPhaseTimeOut
     | RefinementPhaseMarkedByModerator
-    | AllIdeasAreMarked Timestamp
+    | AllIdeasAreMarked { _phaseChangeVotPhaseEnd :: Timestamp }
```
fisx: *vote*, instead of *vot*?

fisx: ah, i see it was intentional.  in that case i don't mind either way.

src/Types.hs
```diff
@@ -614,6 +622,7 @@ defaultSettings :: Settings
 defaultSettings = Settings
     { _durations = Durations { _elaborationPhase = 21, _votingPhase = 21 }
     , _quorums   = Quorums   { _schoolQuorumPercentage = 30, _classQuorumPercentage = 30 }
+    , _frozen = False
```
fisx: alignment?

src/Types.hs
```diff
@@ -416,21 +416,28 @@ instance SOP.Generic EditTopicData
 -- | Topic phases.  (Phase 1.: "wild ideas", is where 'Topic's are born, and we don't need a
 -- constructor for that here.)
 data Phase =
-    PhaseRefinement { _refPhaseEnd :: Timestamp }
+    PhaseWildIdeaFrozen
+  | PhaseRefinement { _refPhaseEnd :: Timestamp }
                                -- ^ 2. "Ausarbeitungsphase"
+  | PhaseRefFrozen  { _refPhaseLeftover :: Double }
+      -- morally @NominalDiffTime@, but it has no @Read@ instance (and Pico has no @Binary@)
```
fisx: i would have made a newtype, but i like this better.  low-tech!  (-:

fisx:
looks very reasonable so far.  thanks for posting it so early!

Mikolaj:
Hmm, why is this travis check https://travis-ci.org/liqd/aula/builds/125054424  merging my PR with current master (and so failing compilation) instead of testing as is? The line number it's complaining about is not even present in the offending file on my branch...

src/Frontend/Page/Topic.hs
Mikolaj: Fair enough, that's a genuine trade-off you are talking about and I respect your preference. Fixed.

fisx:
if you want travis to test your branch without merging master, you can enable "run on every branch" in travis.  currently, this is disabled to reduce our carbon footprint a little.  if you need it and don't have the permissions on travis, let me know and i'll turn the switch.  alternatively, you could test locally and ignore travis for now.

Mikolaj:
Oh, ok, thank you, I missed the moment we switched it off. That's fine.

fisx:
Cool!

I got it to work and made the form in the admin pages look a bit nicer.  Now I'll read the code.

src/Frontend/Page/Overview.hs
```diff
@@ -130,9 +130,13 @@ instance ToHtml PageIdeasInDiscussion where
                            , href_ . U.listIdeas $ IdeaLocationTopic space (topic ^. _Id)
                            ] $ do
                             img_ [ src_ . U.TopStatic $ "images" </> case topic ^. topicPhase of
-                                      PhaseRefinement _ -> "theme_aus.png"
+                                      PhaseWildIdea     -> "theme_aus.png"  -- FIXME
+                                      PhaseWildFrozen   -> "theme_aus.png"  -- FIXME
```
fisx: This should actually never happen (I tested manually), but it does in the test suite due to inconsistent Arbitrary data.  Sigh.  Let's ignore this for now.   What we would want is this:

```
--- a/src/Frontend/Page/Overview.hs
+++ b/src/Frontend/Page/Overview.hs
@@ -16,6 +16,8 @@ module Frontend.Page.Overview
     )
 where
 
+import Control.Exception (assert)
+
 import Action
 import Frontend.Fragment.IdeaList
 import Frontend.Fragment.QuorumBar ()
@@ -130,14 +132,15 @@ instance ToHtml PageIdeasInDiscussion where
                            , href_ . U.listIdeas $ IdeaLocationTopic space (topic ^. _Id)
                            ] $ do
                             img_ [ src_ . U.TopStatic $ "images" </> case topic ^. topicPhase of
-                                      PhaseWildIdea     -> "theme_aus.png"  -- FIXME
-                                      PhaseWildFrozen   -> "theme_aus.png"  -- FIXME
                                       PhaseRefinement{} -> "theme_aus.png"
                                       PhaseRefFrozen{}  -> "theme_aus.png"
                                       PhaseJury         -> "theme_pruf.png"
                                       PhaseVoting{}     -> "theme_abs.png"
                                       PhaseVotFrozen{}  -> "theme_abs.png"
                                       PhaseResult       -> "theme_ergf.png"
+
+                                      PhaseWildIdea   -> assert False $ error "PageIdeasInDiscussion"
+                                      PhaseWildFrozen -> assert False $ error "PageIdeasInDiscussion"
                                  , class_ "theme-grid-item-image"
                                  ]
                             div_ [class_ "theme-grid-item-text"] $ do

```

src/Frontend/Page/Topic.hs
```diff
@@ -189,14 +201,16 @@ instance FormPage CreateTopic where
 
     redirectOf (CreateTopic _ _ _) = U.listTopicIdeas
 
-    makeForm (CreateTopic space ideas timestamp) =
+    makeForm CreateTopic{ _createTopicIdeaSpace
+                        , _createTopicIdeas
+                        , _createTopicRefPhaseEnd } =
```
fisx: Ok, I agree that this is an improvement.  Will try to start using that myself.  (-:

src/LifeCycle.hs
```diff
@@ -107,6 +132,15 @@ wildIdeaCap _i = \case
     Principal        -> []
     Admin            -> []
 
+wildFrozenCap :: Idea -> Role -> [IdeaCapability]
+wildFrozenCap _i = \case
+    Student    _clss -> [CanLike, CanComment]
```
fisx: This `CanLike` is not correct, but it didn't do any harm because the like buttons are filtered out in two places in `Frontend.Page.Idea`.  (This is not an ideal situation, but I think cleaning this up is a bigger task.  I will just remove this `CanLike` here for now.)

src/Persistent/Idiom.hs
```diff
@@ -136,3 +140,24 @@ getUserViews = makeUserView <$$> getAllUsers
...
+-- sure the form treats vacuus setting of the freeze state as invalid.
+--
+-- Note that this operation is atomic, thus ensuring that all topics
+-- are frozen or all are not frozen. Problem could arise, e.g., when
+-- two admins concurrently freeze and thaw, with different speeds.
+saveAndEnactFreeze :: Timestamp -> Freeze -> AUpdate ()
+saveAndEnactFreeze now shouldBeFrozenOrNot = do
+  dbFreeze .= shouldBeFrozenOrNot
+  let change | shouldBeFrozenOrNot == Frozen = PhaseFreeze now
+             | otherwise = PhaseThaw now
```
fisx: I will change this to a `case` expression: easier to read (IMO anyway), and you can make a list of all cases and get a compiler warning if you introduce a new constructor.

src/Persistent/Pure.hs
```diff
@@ -458,16 +460,26 @@ moveIdeasToLocation ideaIds location =
...
+    Just topic <- liftAQuery $ findTopic (t ^. _Id)
     -- FIXME a new topic should not be able to steal ideas from other topics of course the UI will
     -- hide this risk since only ideas without topics will be visible.
     -- Options:
     -- - Make it do nothing
     -- - Make it fail hard
-    moveIdeasToLocation (pt ^. envWith . protoTopicIdeas) (topicIdeaLocation t)
-    return t
+    moveIdeasToLocation (pt ^. envWith . protoTopicIdeas) (topicIdeaLocation topic)
+    return topic
```
fisx: thanks!  (-:

src/Types.hs
```diff
@@ -640,6 +657,7 @@ defaultSettings :: Settings
 defaultSettings = Settings
     { _durations = Durations { _elaborationPhase = 21, _votingPhase = 21 }
     , _quorums   = Quorums   { _schoolQuorumPercentage = 30, _classQuorumPercentage = 30 }
+    , _freeze    = NotFrozen
```
fisx: I would have used `minBound` here and then control which one it is via the `data` definition, but this is more explicit, and as long as we only have one place where we name the default constructor (which we should!), it's almost no difference when it gets refactored.

fisx:
Nice work, thanks!

I made a lot of noise, please check if I got anything wrong or did you unjustice.  Meanwhile, I will merge this if travis agrees.

src/Frontend/Page/Overview.hs
Mikolaj: Yep, that's what I have in old commits in this PR, but then I needed to add the placeholders due to tests (and in other places), as I reported during the standup.

Mikolaj:
Lots of great catches. Yay, code reviewing actually works. Thanks a lot.

PR #403, opened 2016-04-22
User profile edition with avatar upload

np:


np:
This is not finished yet.

fisx:
Two issues (can possibly be handled elsewhere):

1. we have separate constraints for reading temp files and for temp file cleanup now.  is there a new way to make the mistake of not cleaning them up now?  (i will try to reason that out by myself first.)
2. /user/profile forces the user to upload an image every time they want to change their description (or else internal error).  i suspect this is a deficiency of df, but i am still hoping we can find a quick fix.

fisx:
Ok, 1. was easy: the constraint in `Frontend.Core.form` makes sure that cleanup is always called.  That just makes me wonder why the split in the first place?

fisx:
This PR is a strict improvement over what we have, so @np if you agree we can merge this with the "small" issue 2. above unresolved.

Re. that, I will try to think of a good way to handle it, but feel free to take over any time.

Thanks!

src/Frontend/Page/User.hs
```diff
@@ -262,11 +262,12 @@ instance FormPage EditUserProfile where
 editUserProfile :: ActionM m => FormPageHandler m EditUserProfile
 editUserProfile = FormPageHandler
     { _formGetPage   = EditUserProfile <$> currentUser
-    , _formProcessor = \up ->
+    , _formProcessor = \up -> do
+        uid <- currentUserId
         case up ^. profileAvatar of
-            Nothing -> throwError500 "upload FAILED: no file!"  -- FIXME: status code?
+            Nothing ->
+                update . SetUserProfileDesc uid $ up ^. profileDesc
```
fisx: I guess this is what *should* work, so it's a good change, even though it doesn't make sense with the current behavior of df, df-lucid.

src/Frontend/Page/User.hs
```diff
@@ -266,13 +266,18 @@ editUserProfile = FormPageHandler
...
             Just file -> do
                 let dst = "static" </> "avatars" </> cs (uriPart uid) <.> "png"
                     url = "/" <> dst
                 img <- readImageFile (cs file)
                 case img of
-                    Left e -> throwError500 $ "image decoding failed: " <> e
+                    Left _e ->
+                        -- FIXME: this should be dealt with the Nothing case.
+                        -- throwError500 $ "image decoding failed: " <> e
+                        update . SetUserProfileDesc uid $ up ^. profileDesc
```
fisx: Perfect!  (For now anyway.  I will open an issue.)

PR #405, opened 2016-04-23
Admin can set forward and backward the phase of a topic.

andorp:


andorp:
@fisx Please review and translate the few words marked with `TODO`s. Thanks

src/Action.hs
```diff
@@ -507,6 +507,13 @@ topicInRefinementTimedOut = topicTimeout RefinementPhaseTimeOut
 topicInVotingTimedOut :: (ActionPhaseChange m) => AUID Topic -> m ()
 topicInVotingTimedOut = topicTimeout VotingPhaseTimeOut
 
+topicInVotingSetbackTopicToJuryPhase
```
fisx: `resetVotingToJury`?  `votingToJury`?

fisx: Would it make sense to have a function `dangerousChangePhase`?  Then we could the two force-timeouts, the rest from voting to jury phase, and everything else that we may find we have a need of.

src/Frontend/Page/Admin.hs
```diff
@@ -86,6 +86,11 @@ data AdminEditClass = AdminEditClass SchoolClass [UserView]
 
 instance Page AdminEditClass
 
+data AdminPhaseChangeForTopic = AdminPhaseChangeForTopic
```
fisx: `AdminPhaseChange`?

src/Frontend/Page/Admin.hs
```diff
@@ -633,6 +645,49 @@ adminCreateClass = FormPageHandler (pure AdminCreateClass) q
             , _protoUserDesc      = Markdown nil
             }
 
+adminPhaseChangeForTopic
```
fisx: `adminPhaseChange`?

fisx:
I'm pretty happy with this, but I wonder if there is an easy way to give feedback whether the phase change worked or not.  My favorite would be to have a message queue stored in the session state, and some general mechanism in forms that read those messages and add them to the place where we also render digestive-functors errors.  But I am open to discuss this if anybody has ideas.

There are a few other places where we need that, I think.  I'd prefer just leaving a FIXME for that rather than doing it now.

andorp:
I see what you mean by session state based messages to the user, but that is not applicable to the buttons on the topic overview page, a post only requests goes to the server, and no reload happens on the client side. I intentionally left the details out on the admin site, as I would prefer the phase changes related buttons on the topic overview page. We can convert that solution to a non-post only to be able to do the redirect, or after post we can wait for the response and reload the page.

fisx:
sorry, I missed the buttons on the topic page.  ok, but please still add a FIXME to the admin page in case we leave it in.

the buttons should reload the page with the same trick that works for the vote buttons (call reloadOnClick, without anchor, which may need a tweak of the js function because currently we are only calling it with anchors, i think).

i'm not sure i understand the other solutions you are suggesting.  if you disagree with the js option: could you elaborate?

andorp:
I'll try the JS option.

fisx:
Thanks!

src/Action.hs
```diff
@@ -468,12 +501,37 @@ markIdeaInResultPhase iid rv = do
...
+topicInVotingResetToJury tid = do
+    topic <- mquery $ findTopic tid
+    case topic ^. topicPhase of
+        PhaseVoting _ -> topicPhaseChange topic VotingPhaseSetbackToJuryPhase
+        _             -> pure ()
+
+-- * Admin activities
+
+-- | Make a topic timeout if the timeout is applicable.
+-- FIXME: Only admin can do that
```
Mikolaj: Isn't this now assured via capabilities in `LifeCycle`? Hmm, now that I look at it, `ideaCapabilities` seems barely used --- is it that it needs to be queried somewhere before the user is allowed to call `topicForceNextPhase`?

andorp: Good idea to reuse the `LifeCycle` for authorization too. Now we mainly use those to control the rendering of information to the users.

fisx: Again, to clarify: please don't think about authorization yet, we are still working on the happy flow.

The `ideaCapabilities` should change the shape of the pages, but not impose any real restrictions on an adversary.

I think that there are a lot of places in the code where we have `when (something about user ...) $ do ...` that should check for ideaCapabilities instead.  ideaCapabilities should in turn take *all* information into account that affects the shape of the page, *including* user, user role, user id.

Are there any objections to doing them at all?  I'm not insisting on doing these changes now, but I would like to keep the idea in mind.

andorp: > Are there any objections to doing them at all?

I am totally *for* it. That is a shotgun change, and we should do that after the first release.

Mikolaj: Fully agreed.

PR #406, opened 2016-04-23
Paths & Queries

### LINKED ISSUE #31: use uri-bytestring for urls instead of ST

fisx:
More exhaustive support for uri queries and hashes in `Data.UriPath`, `Frontend.Path`.

Fixes #31, improves usability of buttons vote/comment/report, and resolves this:

```haskell
-- [Frontend.hs]
  :<|> adminEventLogCsv . Just  -- FIXME: with QueryParam, we could melt these two routes into one.
                                -- this isn't a hard task, but we need to extend 'UriPath' type.
```

### DIFF
--- a/src/Data/UriPath.hs
+++ b/src/Data/UriPath.hs
@@ -61,7 +61,7 @@ addHash (SlashFreeUriPart s) = SlashFreeUriPart ("#" <> s)
 (DiffUriParts ps q) </?> q' = DiffUriParts ps (q' : q)
 
 instance IsString UriPath where
-    fromString s = DiffUriParts (ps ++) []  -- TODO: why is this `++` not captured by hlint?
+    fromString s = DiffUriParts (ps <>) []
         where ps = SlashFreeUriPart <$> ST.splitOn "/" (cs s)
 
 relativeUriPath :: UriPath -> ST
--- a/src/Data/UriPath.hs
+++ b/src/Data/UriPath.hs
@@ -46,6 +46,7 @@ instance Monoid UriPath where
 
 infixl 7 </>
 infixl 7 </#>
+infixl 7 </?>
 
 (</>) :: UriPath -> UriPart -> UriPath
 DiffUriParts ps qs </> p = DiffUriParts (ps . (p :)) qs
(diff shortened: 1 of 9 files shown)


andorp:
LGTM

src/Data/UriPath.hs
```diff
@@ -31,31 +34,41 @@ instance s ~ ST => ConvertibleStrings UriPart s where
     convertString = unUriPart
 
 -- | An @UriPath@ is a list of @UriPart@s stored as a difference list.
-newtype UriPath = DiffUriParts { appendUriParts :: [UriPart] -> [UriPart] }
+data UriPath = DiffUriParts
+    { appendUriParts    :: [UriPart] -> [UriPart]
+    , diffUriPartsQuery :: HTTP.Query
```
andorp: @np Are you ok with this change?

fisx:
@andorp if we don't get any more feedback soon, could we merge this and defer any potential changes to another PR?

src/Frontend/Path.hs
```diff
@@ -138,8 +137,8 @@ data Space =
 
 instance SOP.Generic Space
 
-viewIdea :: Idea -> Main
-viewIdea idea = IdeaPath (idea ^. ideaLocation) $ ViewIdea (idea ^. _Id)
+viewIdea :: Idea -> Maybe (AUID Comment) -> Main
```
np: I would have made two functions instead of the `Maybe` (the `Maybe` in `ViewIdea` is fine though)

src/Data/UriPath.hs
```diff
@@ -31,31 +34,40 @@ instance s ~ ST => ConvertibleStrings UriPart s where
...
+    }
 
 instance Monoid UriPath where
-    mempty = DiffUriParts id
+    mempty = DiffUriParts id []
 
-    DiffUriParts ps `mappend` DiffUriParts qs = DiffUriParts (ps . qs)
+    DiffUriParts ps q  `mappend` DiffUriParts ps' q' = DiffUriParts (ps . ps') (q <> q')
```
np: One can question whether this is line OK, for instance merging `foo?bar` and `goo?gar` gives `foo/goo?bar&gar`.
* on one side this might be surprising and unexpected
* however this is actually the only total behavior I can think of

src/Data/UriPath.hs
```diff
@@ -31,31 +34,40 @@ instance s ~ ST => ConvertibleStrings UriPart s where
...
 
 infixl 7 </>
 infixl 7 </#>
 
 (</>) :: UriPath -> UriPart -> UriPath
-DiffUriParts ps </> p = DiffUriParts (ps . (p :))
+DiffUriParts ps qs </> p = DiffUriParts (ps . (p :)) qs
```
np: Same remark.

src/Data/UriPath.hs
np: @andorp @fisx While surprised at first, I like this change now.

PR #407, opened 2016-04-23
Daemons

andorp:
* Creates `Daemon` module
 * Add simple message passing based logging to Action
 * Use timeout daemon in persistent implementation


fisx:
What do you mean by "timeout deamon for persistent implementation"?

src/Action/Implementation.hs
```diff
@@ -57,7 +57,7 @@ instance HasSendMail ActionExcept ActionEnv Action where
     sendMailToAddress addr msg = MkAction $ sendMailToAddressIO addr msg
 
 instance ActionLog Action where
-    logEvent = actionIO . print
+    logEvent msg = actionIO =<< views envLogger ($ msg)
```
fisx: heh.  very cool.

src/Daemon.hs
```diff
@@ -0,0 +1,87 @@
+{-# LANGUAGE TemplateHaskell #-}
+
+{-# OPTIONS_GHC -Werror -Wall #-}
+
+module Daemon
```
fisx: call this `LogDeamon`?

fisx: (or `MsgDeamon`, for consistency?)

fisx: ah.  there are more daemons. never mind.

fisx:
@Mikolaj was planning to work on phase timeouts in #402.  Please make sure that you two don't get in each other's way.  (@andorp, Perhaps you can do only the logging for now, as planned?)

src/Persistent/Implementation/AcidState.hs
```diff
@@ -50,21 +51,18 @@ mkRunPersistOnDisk cfg =
...
-            threadDelay delay_us
+        let checkpoint = do
+                logger cfg "[create acid-state checkpoint, archive]"
+                createCheckpoint st
+                createArchive st
+        let logException (SomeException e) = do
+                logger cfg ("error creating checkpoint or archiving changelog: " <> show e)
 
+        let deamon = timeoutDaemon (logger cfg) "checkpoint" delay_us checkpoint logException
+        tid <- deamon ^. timeoutDaemonStart
```
fisx: ok, this answers my earlier question.  thanks!  (-:

fisx: You're not making use of the `Daemon` class here, though.  The reason for that may be that this class is not particularly useful.  I might have missed something (I have before), but you should either come up with a reason for it (at least a potential, future one) or remove it.

andorp: It is used, I replaced `timeoutDaemonStart` etc with `start`.

fisx: > It is used, I replaced `timeoutDaemonStart` etc with `start`.

ok, good.  (i think i was reading a partially outdated version of the code.)

src/Daemon.hs
```diff
@@ -0,0 +1,81 @@
...
+
+    return $ MessageDaemon loop sendMsg
+
+timeoutDaemon
+    :: SystemLogger
+    -> String
+    -> Int
+    -> (IO ())
+    -> (SomeException -> IO ()) -> Daemon ()
+timeoutDaemon logger name delay_us computation handleException = TimeoutDeamon $ do
```
fisx: It would be nice to make `delay_us` a `(delay_sec :: Double)`.  Feel free to disregard.

fisx: in fact, what about this:

```haskell
data Timespan =
    TimespanUs Int
    TimespanMs Int
    TimespanSec Int
    TimespanHourse Int
    TimespanDays Int
```

(with the correct syntax)

fisx: obviously i meant `hoarse`  (-:

Mikolaj:
@andorp, @fisx: I think the actual daemons and the particulars of freezing and thawing are quite orthogonal, so I hope we can do that in parallel; I will just use the daemons to chase some more data structures that there are currently.

andorp:
@Mikolaj That's great.

fisx:
Doesn't look like it from the number of commits I made, but I like it!  Good to merge if you and travis agree.

Thanks!

PR #417, opened 2016-04-26
Progress bar that includes abstained votes

localgrr:
Looks schmancy


fisx:
Looks great now, thanks!  Your lucid code was good, too.  Sorry for being so opinionated..  (-:

PR #419, opened 2016-04-26
Simple status messages

andorp:
* Machinery to render status messages for the user
 * One message on saving quorums
 * `pageFrame` uses the `Frame` type as a simplification.

### DIFF
--- a/src/Frontend/Core.hs
+++ b/src/Frontend/Core.hs
@@ -52,7 +52,7 @@ module Frontend.Core
     , form
 
       -- * frames
-    , Frame(..), frameBody, frameUser
+    , Frame(..), frameBody, frameUser, frameMessages
     , makeFrame
 
       -- * sort & filter
@@ -416,7 +416,7 @@ form formHandler = getH :<|> postH
 
 -- | Wrap anything that has 'ToHtml' and wrap it in an HTML body with complete page.
 data Frame body
-    = Frame { _frameUser :: User, _frameBody :: body }
+    = Frame { _frameUser :: User, _frameBody :: body, _frameMessages :: [StatusMessage] }
     | PublicFrame               { _frameBody :: body }
   deriving (Show, Read, Functor)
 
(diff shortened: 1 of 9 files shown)


andorp:
@fisx: Please translate the message.

src/Frontend/Core.hs
```diff
@@ -416,23 +416,24 @@ form formHandler = getH :<|> postH
 
 -- | Wrap anything that has 'ToHtml' and wrap it in an HTML body with complete page.
 data Frame body
-    = Frame { _frameUser :: User, _frameBody :: body }
+    = Frame { _frameUser :: User, _frameBody :: body, _frameMessages :: [StatusMessage] }
```
fisx: After logout, we are on a public frame, and yet it makes sense to display a message in the frame like "thank you for playing".  But we can change that once we need it.

src/Frontend/Core.hs
```diff
@@ -416,23 +416,24 @@ form formHandler = getH :<|> postH
...
 
 makeFrame :: (ActionPersist m, ActionUserHandler m, MonadError ActionExcept m, Page p)
           => m p -> m (Frame p)
 makeFrame mp = do
   isli <- isLoggedIn
+  msgs <- flushMessages
   let isPrivate = isPrivatePage mp -- Here 'm' is used as the 'proxy'.
   if | not isli && isPrivate -> redirect . absoluteUriPath $ relPath P.Login
-     | isli     || isPrivate -> Frame <$> currentUser <*> mp
+     | isli     || isPrivate -> Frame <$> currentUser <*> mp <*> pure msgs
```
fisx: ...  or perhaps we should do something: this means that logout flushes the message queue without showing it.  Ok, but not ideal.

src/Frontend/Page/Admin.hs
```diff
@@ -288,7 +288,9 @@ adminQuorum :: ActionM m => FormPageHandler m PageAdminSettingsQuorum
 adminQuorum =
     FormPageHandler
         (PageAdminSettingsQuorum <$> query (view dbQuorums))
-        (update . SaveQuorums)
+        (\qs -> do
+            update $ SaveQuorums qs
+            addMessage "Quorums are saved.")
```
fisx: or:

```haskell
         (update . SaveQuorums >=> const (addMessage "Quorums are saved."))
```

(What would you name something that behaves to `>>` as `>=>` behaves to `>>=`?)

We can keep your version if you prefer.

src/Frontend/Core.hs
```diff
@@ -412,7 +412,7 @@ form formHandler = getH :<|> postH
     processor2 page result = absoluteUriPath . relPath . redirectOf page <$> processor result
 
 
--- * frames
+-- * frame creation
```
fisx: This commit is great, I am glad I recovered the diff even though you did a good job hiding it.  (Spoiiler: `pageFrame` now takes a frame and is much easier to call from the `ToHtml` instance and elsewhere.)

fisx: Please disregard last comment.  It made no sense.

fisx:
Nice work, thanks!

Please merge if you are happy with my changes, or pass back to me with more comments.

src/Frontend/Page/Admin.hs
andorp: I would keep my version, as we will create the `Action.Idiom` module soon.

andorp:
LGTM, first we should merge the #402

PR #420, opened 2016-04-27
Discontinue neutral votes.

fisx:


Mikolaj:
LGTM. But actually, `git grep -i neutr` shows some more stuff, which can be removed if, we want to wipe out any trace of neutral votes (unless we think about easily bringing them back, just in case).

fisx:
Nice catch, thanks!  I will take another look.

andorp:
LGTM

PR #424, opened 2016-04-27
More expressive types for phase timespans.

fisx:
Fixes 2 issues with creating frozen topics:
- see FIXME;
- the leftover time was set to 0 instead of the configured maximum.

### DIFF
--- a/src/Persistent/Pure.hs
+++ b/src/Persistent/Pure.hs
@@ -121,6 +121,7 @@ module Persistent.Pure
     )
 where
 
+import Control.Exception (assert)
 import Control.Lens
 import Control.Monad.Except (MonadError, ExceptT(ExceptT), runExceptT, throwError)
 import Control.Monad.Reader (MonadReader, runReader, asks)
@@ -145,6 +146,7 @@ import qualified Data.Set  as Set
 import qualified Data.Text as ST
 
 import Types
+import qualified LifeCycle
 
 
 -- * state type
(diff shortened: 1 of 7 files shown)


src/Persistent/Pure.hs
```diff
@@ -463,14 +465,15 @@ moveIdeasToLocation ideaIds location =
...
     dbFrozen <- liftAQuery $ view dbFreeze
-    -- FIXME: the invariant that topics start in Ref phases is now spread
-    -- across a couple of different places. To merge it, or when it no longer
-    -- holds, use the full @phaseChange@ machinery (requiring @now@, etc.).
-    when (dbFrozen == Frozen) $ setTopicPhase (t ^. _Id) (PhaseRefFrozen 0)
+    when (dbFrozen == Frozen) $ do
+        case LifeCycle.phaseTrans (t ^. topicPhase) (LifeCycle.PhaseFreeze now) of
+            Just (phase'@(PhaseRefFrozen _), [])
+                -> setTopicPhase (t ^. _Id) phase'
+            bad -> assert False . error $ "addTopic: internal error: " <> show bad
```
Mikolaj: Yeah, thank you for that. If only I wrote `PhaseRefFrozen  {_refPhaseLeftover = 0}` instead of `PhaseRefFrozen 0`, I'd spot my bug outright. :) This could be fixed by fetching the current Duration and inserting it in place of 0, but that's not KISS any more, so the proper, general and extensible way you implemented beats it.

Mikolaj:
LGTM. Merging.

PR #425, opened 2016-04-27
Status messages

fisx:


src/Frontend/Page/User.hs
```diff
@@ -113,11 +113,11 @@ userSettings = FormPageHandler (PageUserSettings <$> currentUser) changeUser
         updatePass uid oldPass newPass1 newPass2
         pure ()
 
-    updateEmail uid em = do
+    updateEmail uid em = do  -- TODO: if email hasn't changed, do not add message.
```
np: I suggest that `setUserEmail` in `Persist` returns whether or not the email has changed. We could have a variation of `.~` and `?~`.

However I'm also wondering if the lack of message be concerning, a single message saying that the changes are now saved might be enough.

fisx: i like the second option.  will do, thanks!

fisx:
oops, closed by accident, and then got very confused by the github message "successfully merged and closed".  but it doesn't look like it got merged.

i'll leave this closed and make a new PR.

PR #431, opened 2016-04-28
More field validation

andorp:


src/Frontend/Validation.hs
```diff
@@ -55,6 +60,15 @@ inRange mn mx =
     isBetween n = mn <= n && n <= mx
 
 
+-- * simple validators
+
+-- TODO: Translate
+nonEmpty :: (Monad m, Monoid v, IsString v) => FieldName -> Form v m String -> Form v m String
+nonEmpty name = TD.validate checkNonEmpty
+  where
+    checkNonEmpty [] = TD.Error . fromString $ unwords [name, ":", "can not be empty"]
```
fisx: "darf nicht leer sein"

src/Frontend/Core.hs
```diff
@@ -163,7 +164,8 @@ type FormHandler p = FormH '[HTML, PlainText] (Frame (FormPageRep p)) (FormPageR
 semanticDiv :: forall m a. (Monad m, Typeable a) => a -> HtmlT m () -> HtmlT m ()
 semanticDiv t = div_ [makeAttribute "data-aula-type" (cs . show . typeOf $ t)]
 
-type DfForm a = forall m. Monad m => DF.Form (Html ()) m a
+type DfFormM m a = DF.Form (Html ()) m a
```
fisx: Elsewhere we use `SomeThingM` for constraint types.  I assume you needed the unquantified version somewhere, but perhaps we can give up on the quantified version altogether, and always quantify explicitly?  If that's not good, what about `DfForm` for `DfFormM` and `DfFormA` for `DfForm`?  (`A` for "all".  Hm, I still prefer removing `DfForm`.)

src/Frontend/Page/Idea.hs
```diff
@@ -311,6 +312,12 @@ instance ToHtml IdeaVoteLikeBars where
             PhaseVotFrozen{}  -> toHtml $ voteBar nil
             PhaseResult       -> toHtml $ voteBar nil
 
+validateMarkdown :: Monad m => FieldName -> DfFormM m String -> DfFormM m Document
+validateMarkdown name = fmap (Markdown . cs) . nonEmpty name
+
+-- TODO: Translation
+validateIdeaTitle :: Monad m => DfFormM m String -> DfFormM m ST.Text
+validateIdeaTitle = fmap cs . validate "Idea title" (many1 (alphaNum <|> space))
```
fisx: "Titel der Idee"

src/Frontend/Page/Idea.hs
```diff
@@ -337,9 +345,10 @@ instance FormPage EditIdea where
     redirectOf (EditIdea idea) _ = U.viewIdea idea
 
     makeForm (EditIdea idea) =
+        -- TODO: Translation
```
fisx: Nothing to translate here except "Idee", and that looks fine to me.

src/Frontend/Page/Idea.hs
```diff
@@ -442,7 +451,8 @@ instance FormPage CreatorStatement where
     redirectOf (CreatorStatement idea) _ = U.viewIdea idea
 
     makeForm (CreatorStatement idea) =
-        "statement-text" .: (Markdown <$> DF.text (unMarkdown <$> creatorStatementOfIdea idea))
+        -- TODO: Translate
```
fisx: "Statement des Autors" (I suspect that will be changed by QA, but I can't think of anything better.)

tests/AulaTests.hs
```diff
@@ -161,3 +161,6 @@ boundary mn mx = frequency
     , (1, pure mx)
     , (98, choose (mn, mx))
     ]
+
+nonEmptyMarkdown :: Gen Document
+nonEmptyMarkdown = Markdown . (cs :: String -> ST) . getNonEmpty <$> arb
```
fisx: I like that this is a special case of `nonEmpty`.  I guess it'd be nice to run pandoc here later, but even if we don't (which is my current preference) it's good to make that explicit in one place.

src/Frontend/Page/Idea.hs
```diff
@@ -418,11 +418,13 @@ instance FormPage JudgeIdea where
...
 
     makeForm (JudgeIdea IdeaFeasible _ _) =
+        -- TODO: Field validation
         Feasible
         <$> "jury-text" .: (Markdown <$$> (`justIfP` (not . ST.null)) <$> DF.text Nothing)
     makeForm (JudgeIdea IdeaNotFeasible _ _) =
+        -- TODO: Translation
         NotFeasible
-        <$> "jury-text" .: (Markdown <$> DF.text Nothing)
+        <$> "jury-text" .: (validateMarkdown "Jury text" (DF.string Nothing))
```
fisx: "Anmerkungen zur Durchführbarkeit"

src/Frontend/Page/Idea.hs
```diff
@@ -384,17 +384,17 @@ createOrEditPage showDeleteButton cancelUrl v form p = semanticDiv p $ do
...
     type FormPageResult CommentIdea = Comment
 
     formAction (CommentIdea idea mcomment) = U.commentOrReplyIdea idea mcomment
 
     redirectOf (CommentIdea idea _) = U.viewIdeaAtComment idea . view _Id
 
     makeForm CommentIdea{} =
-        "comment-text" .: (Markdown <$> DF.text Nothing)
+        -- TODO: Translation
+        "comment-text" .: (CommentContent <$> validateMarkdown "Comment" (DF.string Nothing))
```
fisx: "Verbesserungsvorschlag"

fisx: (The correct translation of that word is "suggestion for improvement", and it's part of the didactic concept to use that word instead of "Kommentar".  "you all smell bad" is also a valid "Kommentar", but doesn't suggest an improvement. :-)

tests/Frontend/CoreSpec.hs
```diff
@@ -263,8 +263,10 @@ postToForm (F g) = do
         let frm = makeForm page
         env <- runFailOnError $ (`payloadToEnv` payload) <$> getForm "" frm
 
-        (_, Just payload') <- runFailOnError $ postForm "" frm (\_ -> pure env)
-        liftIO $ payload' `shouldBe` payload
+        (_, mpayload) <- runFailOnError $ postForm "" frm (\_ -> pure env)
+        case mpayload of
+            Nothing       -> fail "Form validation has failed."
+            Just payload' -> liftIO $ payload' `shouldBe` payload
```
fisx: It would be even nicer if we could somehow get the validation errors here.  Just an idea, feel free to ignore.

fisx:
LGTM so far!  I added the translations as comments, hope that's ok?  If you agree that that's easier, I'll do it directly in the code again from now on.

src/Frontend/Core.hs
andorp: Removed.

src/Frontend/Validation.hs
```diff
@@ -63,11 +64,17 @@ inRange mn mx =
...
-nonEmpty name = TD.validate checkNonEmpty
-  where
-    checkNonEmpty [] = TD.Error . fromString $ unwords [name, ":", "can not be empty"]
-    checkNonEmpty xs = TD.Success xs
+nonEmpty = TD.validate . checkNonEmpty
+
+optionalNonEmpty
+    :: (Monad m, Monoid v, IsString v)
+    => FieldName -> Form v m (Maybe String) -> Form v m (Maybe String)
+optionalNonEmpty = TD.validateOptional . checkNonEmpty
```
fisx: I wonder how this is different from no validation?

fisx:
LGTM, still!  (-:

thanks!

src/Frontend/Validation.hs
andorp: Comes from two fact: the antisymmetry of the digestive functors, and the placeholder for further markdown validation.

PR #434, opened 2016-04-29
Field validation on Topic, User and Login pages.

andorp:


src/Frontend/Fragment/Category.hs
```diff
@@ -25,16 +25,16 @@ import qualified Data.Text as ST
 import qualified Lucid
 import qualified Text.Digestive.Form as DF
 import qualified Text.Digestive.Lucid.Html5 as DF
-import qualified Text.Digestive.Types as DF
 
 
 -- | FIXME: 'makeFormSelectCategory', 'formPageSelectCategory' should be a subform.  (related: `grep
 -- subform src/Frontend/Page/Topic.hs`.)
+-- FIXME: Error and non selected category are inseparable cases.
```
andorp: This could be achieved with a custom `instance Enum a => Enum (Maybe a)` instance:

````haskell
instance Enum a => Maybe (Enum a) where
    toEnum 0 = Nothing
    toEnum n = Just (toEnum (n-1))

    fromEnum Nothing = 0
    fromEnum (Just n) = (fromEnum n) - 1
````

and also some work with the JS code.

**Comment by andorp** [comment]
I am finished with this PR, going to implement the EventLog and System messages on top of this one.

File: src/Frontend/Fragment/Category.hs (hunk 1 above)
Comment (fisx) [inline src/Frontend/Fragment/Category.hs]: > and also some work with the JS code.

that's the key here.  and since the error can only arise from internal errors or adversarial behavior, i would like to leave this as is for now.

File: src/Frontend/Page/Idea.hs [hunk 2]
```diff
@@ -396,7 +387,7 @@ instance FormPage CommentIdea where
     redirectOf (CommentIdea idea _) = U.viewIdeaAtComment idea . view _Id
 
     makeForm CommentIdea{} =
-        "comment-text" .: (CommentContent <$> validateMarkdown "Verbesserungsvorschlag" (DF.string Nothing))
+        "comment-text" .: (CommentContent <$> validateMarkdown "Verbesserungsvorschlag" (Markdown <$> DF.text Nothing))
 
     -- FIXME styling
```
fisx: I don't think these are very helpful either.  Styling issues are kept track of elsewhere, and having them here just risks that they will remain after styling has been fixed.

src/Types.hs
```diff
@@ -115,6 +115,11 @@ instance (SOP.Generic a, SOP.Generic b, SOP.Generic c) => SOP.Generic (Either3 a
 app2 :: (c -> d) -> (a -> b -> c) -> a -> b -> d
 app2 f g x y = f $ g x y
 
+infixr 9 <..>
+
+(<..>) :: (c -> d) -> (a -> b -> c) -> a -> b -> d
+(<..>) = app2
```
fisx: I like this!  (-:

(Would `..` have been even better?)

fisx:
LGTM.

My biggest concern is the FIXME noise.  (-:

I made a few edits and did the translations, and will merge if travis agrees.

Thanks!

src/Frontend/Page/Idea.hs
andorp: Not part of this PR :)

src/Types.hs
andorp: Probably it clashes with the `1 .. 10` notation. :)

PR #438, opened 2016-05-02
2015 05 cc markdown ui

localgrr:
You need to update your git submodules
I've only added very simple markdown css - I've used a reset stylesheet so there are no default HTML styles any more (there are pros and cons, this is a con) I figure I would add back in markdown styles in a need to use basis, let me know whats going to come up.


fisx:
d3 is in submodules/, showdown is in static/third-party.  i'm not sure there is a good reason for either, but can we keep this consistent?

fisx:
also: lots of nit-picks.  otherwise LGTM.  and this was *fast*!  (-:

PR #447, opened 2016-05-04
EventLog

fisx:


fisx:
I just realized that the log daemon is not living in `Action`, and for good reasons: `Action` is not multi-threaded.

We could still store the event log in AulaData if we chose not to use `Logger`, but I think we should have one sink of all events that are going to moderators, admins, sys-admins, students, ....  This way it'll be a local change to, say, start sending sms out to moderators if a wild idea has reached the quorum, and an easy one if sys admins already get notified on that channel about disk failure.

We could write a ToRecord instance and forget / revert all the beautiful changes above.  But that would be neither very fast (not necessarily a problem) nor easily revertible (the event log is human-readable and destroys some information w.r.t. both the cold and the warm event log.

My favorite option right now is to send the cold event log items to the logger and use them to find an efficient and revertible serialization.

Comments, insights, opinions etc. very welcome.

np:
A quick comment: @fisx  :+1: for keeping the event log out of AulaData and use something better suited for that.

src/Action.hs
```diff
@@ -86,6 +89,21 @@ module Action
 
     , module Action.Smtp
     , sendMailToRole
+
+    -- * moderator's event log
```
andorp: I have the feeling that we should introduce the `Action.Idiom` module here.

fisx: Probably.

In any case, *some* re-structure of `Action` wlll be due soon.  And of `Pure`.

fisx:
I am going to merge so I can make another release, but please somebody review this and have an opinion.  It was quite a ride.

src/Action/Implementation.hs
```diff
@@ -64,6 +71,14 @@ instance HasSendMail ActionExcept ActionEnv Action where
 instance ActionLog Action where
     log msg = actionIO =<< views envLogger ($ msg)
 
+    readEventLog = do
+        cfg <- viewConfig
+        rows :: [EventLogItemCold]
```
andorp: We need to add exception handling here.

andorp: Or the `actionIO` could wrap exception...

src/Logger/EventLog.hs
```diff
@@ -22,79 +22,90 @@ import Data.String.Conversions
...
+  deriving (Generic)
 
-data EventLogItem = EventLogItem IdeaSpace Timestamp User EventLogItemValue
-  deriving (Eq, Ord, Show, Read, Generic)
+data EventLogItem' user topic idea comment =
+    EventLogItem' IdeaSpace Timestamp user (EventLogItemValue' user topic idea comment)
+  deriving (Eq, Show, Generic)
 
-data URLEventLogItem = URLEventLogItem URL EventLogItem
+data EventLogItemValue' user topic idea comment =
```
andorp: Please rename `EventLogItemValue'` to `EventLogItemValue` .

src/Action/Implementation.hs
```diff
@@ -64,6 +71,14 @@ instance HasSendMail ActionExcept ActionEnv Action where
...
+            <- fmap adecode . LBS.lines <$> actionIO (LBS.readFile (cfg ^. logging . eventLogPath))
+        EventLog (cs $ cfg ^. exposedUrl) <$> (warmUp `mapM` rows)
+      where
+        adecode = fromMaybe (error "readEventLog: inconsistent data on disk.") . Aeson.decode
```
andorp: We should log the the line that has parse errors.

src/Action/Implementation.hs
fisx: You mean "event log file not found"?  I can't think of a sane situation in which that happens, but you're probably right that we should catch it.

src/Logger/EventLog.hs
fisx: good idea.  that's an artefact of a previous state of the PR.  i'll take care of this.

src/Action/Implementation.hs
fisx: also good point.

src/Action/Implementation.hs
andorp: Something deletes it from the disk, the location is not mounted, disk is full etc...

PR #448, opened 2016-05-04
Generic shrinking.

andorp:


src/Arbitrary.hs
```diff
@@ -492,7 +634,7 @@ instance Arbitrary Document where
     arbitrary = arbMarkdown
 
     shrink (Markdown "") = []
-    shrink _ = [Markdown ""]
+    shrink _ = [Markdown "", Markdown "x"]
```
andorp: The three big test partitions are the empty, one, n characters in a document. If a test fails on a non-empty string, printing out one character long document is nicer.

src/Arbitrary.hs
```diff
@@ -78,119 +84,176 @@ import Types
...
-
-    p :: Proxy Arbitrary
-    p = Proxy
+    subs = apInjs_POP (hcpure proxyArb (scale scaling arbitrary))
 
 garbitrary :: forall a. (Generic a, All2 Arbitrary (Code a)) => Gen a
 garbitrary = garbitrary' (max 0 . subtract 10)
 
+gshrink :: forall a . (Generic a, All2 Arbitrary (Code a)) => a -> [a]
+gshrink = List.map to . shrinkSOP . from
```
andorp: I am going to make a PR in basic-sop from this.

fisx:
LGTM.  Looking forward to extremely short counter-example pages in failing tests from now on!  (-:

PR #450, opened 2016-05-05
Topic description

andorp:
* Refactor in validation module
* Intorducing Description type
* Intorducing Frontend.Constant module for global constants


tests/AulaTests.hs
```diff
@@ -163,5 +166,11 @@ boundary mn mx = frequency
...
 
+nonEmptyST :: Gen ST
+nonEmptyST = (cs :: String -> ST) . getNonEmpty <$> arb
+
 nonEmptyMarkdown :: Gen Document
-nonEmptyMarkdown = Markdown . (cs :: String -> ST) . getNonEmpty <$> arb
+nonEmptyMarkdown = Markdown <$> nonEmptyST
+
+validTopicDescription :: Gen Description
+validTopicDescription = Description . Text.take Constant.topicDescMaxLength <$> nonEmptyST
```
fisx: Why not make the `Arbitrary` instance valid?

src/Frontend/Validation.hs
```diff
@@ -48,60 +55,81 @@ type FieldParser a = Parsec String () a
...
+    g . f = FieldValidator (unFieldValidator g <=< unFieldValidator f)
+
+instance Arrow FieldValidator where
+    arr f   = FieldValidator (DF.Success . f)
+    first f = FieldValidator $ \(x,y) -> case unFieldValidator f x of
+                DF.Success z -> DF.Success (z,y)
+                DF.Error   e -> DF.Error e
+
+markFieldName :: FieldName -> FieldValidator a b -> FieldValidator a b
+markFieldName fieldName v = FieldValidator $ \x -> case unFieldValidator v x of
```
fisx: `labelValidationError`?  `prefixLabel`?  `markErrorWithFieldName`?

src/Frontend/Validation.hs
```diff
@@ -48,60 +55,81 @@ type FieldParser a = Parsec String () a
...
+    errorString = filter (/= '\n') . errorMsgs . errorMessages
     -- | Parsec uses 'ParseError' which contains a list of 'Message's, which
     -- are displayed if a parse error happens. Also it gives control to the
     -- client code to make their translation of those connectors. The German
     -- translations here are probably not helping to form perfect phrases in
     -- all situations.
     errorMsgs = showErrorMessages "oder" "unbekannt" "erwartet" "unerwartet" "zu kurz"
 
+calcValidator :: (Monad m) => FieldName -> FieldValidator a b -> a -> Result (HtmlT m ()) b
+calcValidator n v = errorToHtml . unFieldValidator (markFieldName n v)
```
fisx: what does this do?  i think it also needs another name.

fisx:
LGTM!  I made a lot of noise, though.  Please take a quick look and merge if you agree.

tests/AulaTests.hs
andorp: In the `Arbitrary` module many we generate data with noise. I don't want to start clean that module in different PRs. This generator is used in form validation, that is the reason to be.

fisx: summarizing a brief off-line discussion: removing noise from the arbitrary instances to accomodate application logic is problematic because it introduces assumptions that are hard to look up.  worse, if different parts of the application logic have overlapping domains, we will be tempted to generate the intersection, punching holes in our data that are hard to track.

however:

1. we already do if (accomodate app logic) in `src/Arbitrary.hs`, and it's confusing to sometimes do sometimes have the nice-noise generators there and sometimes in other places like this one
2. white noise is mostly useless

in the long run, it would be nice to have more sophisticated testing (andorp: feel free to write a few lines on your ideas for the record :-).  for now, i will try to find the time soon to move these happy-noise generators into `src/Arbitrary.hs` and copy this reasoning to the module haddock.

andorp:
````haskell
data Contract a b = Contract {
      gen :: Gen a
    , box :: a -> b 
    , post :: b -> Prop 
    }

test :: Contract a b -> Spec
test = ...
````
I have the gut feeling this construction forms a category, which at least has products.

Apply TDD, would mean the define those contract, and test them against the function, can lead us a framework where we easily can put up complex test cases.

andorp:
LGTM

PR #451, opened 2016-05-05
Show 'table div' if the idea has reached the quorum.

andorp:


andorp:
Real design in: https://marvelapp.com/ehhb43#10108433

fisx:
Looks good, but it indeed lacks frontend work.  @localgrr could you take over?

andorp:
LGTM

PR #457, opened 2016-05-06
Edit, delete comments

andorp:


src/Frontend/Page/Idea.hs
```diff
@@ -104,6 +107,13 @@ data ReportComment = ReportComment Comment
 
 instance Page ReportComment where
 
+data EditComment
+    = EditComment Idea Comment
```
fisx: you don't need `Idea`, because is part of the `CommentKey` that's part of the `Comment`.  I think.

src/Frontend/Page/Idea.hs
```diff
@@ -527,6 +558,33 @@ commentIdea loc ideaId =
             return comment)
         "Der Verbesserungsvorschlag wurde gespeichert."
 
+editComment :: ActionM m => IdeaLocation -> AUID Idea -> AUID Comment -> FormPageHandler m EditComment
+editComment loc ideaId commentId =
+    formPageHandlerWithMsg
+        (equery $ do
+            idea <- maybe404 =<< findIdea ideaId
+            comment <- maybe404 $ idea ^. ideaComments . at commentId
```
fisx: We should use `findComment` here as below, right?

src/Frontend/Page/Idea.hs
fisx: just learned that `Idea` is needed for its title.  In that case we should still simplify this type and only carry around the title.  Less confusing.

fisx: Let's ignore these questions for now.  Should happen naturally in a bigger comment refactoring.

andorp:
LGTM

PR #460, opened 2016-05-09
Phase timeout

np:


np:
This is not really finished but I'd done for today.

* the display of leftover time is wrong: we see a negative number of seconds. We might be calling diffTimestamp the wrong way around. Also we need to display a number of days.
* some english text
* the `fromNow` iso could be used more

fisx:
strangely, the voting phase duration is fine, only the refinement phase duration is off by 76 days.

fisx:
correction: it's always `-55` days, no matter what the phase duration.

fisx:
i think this would be a good time to write some tests around the timestamp / timespan unit conversion functions...

fisx:
I think I've finished this and will merge now.  Please take a look at my commits if there is anything you'd like to change.

np:
LGTM thanks for finishing this!

PR #462, opened 2016-05-09
Context dependent CreateTopic and EditTopic menu items.

andorp:


src/LifeCycle.hs
```diff
@@ -96,6 +116,7 @@ data IdeaCapability
     | CanAddCreatorStatement
     | CanEdit
     | CanMoveBetweenTopics  -- also move between (and into and out of) topics
+                            -- the idea can be moved to a topic
```
fisx: what is this line for?  i thought it was covered by the above comment.

andorp: You are right, it is a leftover. It is not clear where this capability should be defined. One can add an ideas to a topic and it a capability of an idea, or can can modify topic to add many ideas. We need a clarification here. Anyway I am going to remove this line in an Errata. Thanks for spotting that out.

fisx:
LGTM!

PR #464, opened 2016-05-10
Render "on the table" icon for only wild ideas.

andorp:


fisx:
LGTM, except the `wildPhase` function seems a bit ad hoc.  I will wait for #460 to land and rebase.

fisx:
@andorp please take another look, but with the changes from #460 this should be in the spirit of your change?

andorp:
LGTM

PR #465, opened 2016-05-10
Student can add creator statement only for winning ideas.

andorp:


fisx:
Nice and compact!  (-:

PR #467, opened 2016-05-10
Moderators can edit creator statement.

andorp:


src/LifeCycle.hs
```diff
@@ -83,6 +85,11 @@ phaseTrans PhaseResult PhaseThaw{} = Just (PhaseResult, [])
 phaseTrans _ _ = Nothing
 
 
+-- * Capabilities api
+
+anyCapOf :: (Eq a) => [a] -> [a] -> Bool
+anyCapOf required caps = any (`elem` caps) required
```
fisx: I would prefer this to be inlined instead of giving it a name.  The code is easier to understand than the function name, and it doesn't require to look up the implementation.

If you insist on keeping it, I would prefer the same argument order as with `elem`, and perhaps the name `anyElem` or `elemAny`.  Nothing in the type signature requires `a ~ Capability`.

andorp:
LGTM

PR #468, opened 2016-05-10
Event logging and tests.

fisx:


fisx:
That was the second time today that we lost 1-2 hours finding stale-event-log issues.  This time it hit @np; I've written down his fix.

andorp:
LGTM

PR #471, opened 2016-05-10
Prefilter ideas based on tab.

andorp:


src/Frontend/Path.hs
```diff
@@ -285,8 +288,8 @@ ideaPath loc mode root =
 -- factored out.
 space :: Space -> UriPath -> UriPath
 space ListTopics                  root = root </> "topic"
-space (ViewTopicIdeasVoting tid)  root = root </> "topic" </> uriPart tid </> "ideas" </> "voting"
-space (ViewTopicIdeasWinning tid) root = root </> "topic" </> uriPart tid </> "ideas" </> "winning"
+--space (ViewTopicIdeasVoting tid)  root = root </> "topic" </> uriPart tid </> "ideas" </> "voting"
+--space (ViewTopicIdeasWinning tid) root = root </> "topic" </> uriPart tid </> "ideas" </> "winning"
```
fisx: what's this about?

fisx:
- The error in the test suite has to do with the fact that the paths are very confused about the distinction between wild ideas and topics.  That's two contexts in which ideas can be listed, but they behave differently (in a space, there are no tabs for feasible / winning ideas, for instance).  See also the `TODO` that i added.
- @andorp has noticed that in topic idea lists, it looks like the m-active class gets lost on the tab links when filtering / sorting.  I haven't investigated that yet.

fisx:
I've unassigned myself, so feel free to take over again.  If you don't, I'll get back to this later today or tomorrow.

fisx:
the error message is now pretty straight-forward: we are trying to open a topic tab when browsing wild ideas, and the assertion in `listIdeas'` triggers.

fisx:
This is getting scary.  I had to rename some more test cases since (I suspect) the generated data is inconsistent. I fixed some inconsistencies, but still got errors and non-termination (?!).  Manual testing seems fine.

This was taking too much time.  I think we need to consider replacing `src/Arbitrary.hs` with a more randomized version of `tests/AulaData/Stories/*`.

fisx:
I am still getting this:

```haskell
  tests/Frontend/PathSpec.hs:60: 
  1) Frontend.Path, Paths and handlers, Every path has a handler
       uncaught exception: AssertionFailed (src/Frontend/Path.hs:235:5-10: Assertion failed
       ) (after 97 tests)
       *** Failed! (after 97 tests): 
       Exception:
         src/Frontend/Path.hs:235:5-10: Assertion failed
       Space SchoolSpace (ListIdeasInSpace (Just (IdeasQuery {_ideasQueryF = IdeasWithCat {_catFilter = CatTeaching}, _ideasQueryS = SortIdeasBySupport})))
```

I don't understand how this test can trigger this error, and why the last line shows a value that should be *returned* by the crashing `listIdeas'`, and that looks perfectly fine.

fisx:
No, the latter makes sense: rendering the mock-generated page values triggers the same problem that is triggered in the `renderMarkup` tests that I just disabled.  I think.

fisx:
> - @andorp has noticed that in topic idea lists, it looks like the m-active class gets lost on the tab links when filtering / sorting.  I haven't investigated that yet.

this one is solved in e45d958.

fisx:
- manual testing: success
- test suite with `-a 6000`: success
- i will leave 09588a0 alone and just make a note to not use sensei for a bit whenever it starts running off with all the memory
- travis is being very quiet right now, so i'll merge with just the local test.

Makefile
```diff
@@ -25,20 +25,20 @@ unregister-full:
 
 # only aware of aula sources
 sensei: .phony aula.unregister
-	$(EXEC) sensei $(AULA_SOURCES) tests/Spec.hs $(SENSEI_DEFAULT_ARGS) $(SENSEI_ARGS)
+	$(EXEC) sensei -j5 $(AULA_SOURCES) tests/Spec.hs $(SENSEI_DEFAULT_ARGS) $(SENSEI_ARGS)
```
andorp: Can you point me to the documentation for '-j5'?

fisx: > Can you point me to the documentation for '-j5'?

what about `--ghc-options="-Werror -Wall -O0"` then?  or `docker exec -it`?  `grep -Hni`?  i think the documentation of the clis should not be part of the Makefile.

andorp: I don't want include it in the makefile. I was searching for -j and tried manually with sensei, but sensei cli --help, its documentation, and its source code is not really helpful. I would like what does the "-j5" meam?

fisx: oops, sorry, i misunderstood your comment.

https://downloads.haskell.org/~ghc/latest/docs/html/users_guide/flag-reference.html#idp46760134970192

you sumbled over the fact that it's not a sensei flag, but a ghc flag.  (-:

(there is also something about cabal and -j, but i couldn't find a good reference for that.)

so, should I add a comment to the Makefile, or is it ok to expect people to find and read the ghc docs?

andorp: np, thanks :)

andorp: I am ok with the reference being here only in these comments.

Issue #216, opened 2016-03-15
Event log


andorp:
* Define the event log format
 * Define combinators
 * Use event logging at errors, redirection, login, logout

andorp:
@fisx is this implemented already?

fisx:
not done.  the only thing we have is a user-facing event log type ("idea has been edited" and things like that, not "disk full", "database inconsistent", ...).

fisx:
Still missing afterin #473:

```haskell
  | EventLogTopicNewPhase         topic Phase Phase
  | EventLogIdeaNewTopic          idea (Maybe topic) (Maybe topic)
```

fisx:
(for ideas-new-topic, we should implement moving ideas first (via the `...`-menu. it may be different on the UI from editing topics.)

PR #473, opened 2016-05-11
Throw event log events

fisx:
Related: #216 (not sure we should close that yet)

Missing:

```haskell
  | EventLogTopicNewPhase         topic Phase Phase
  | EventLogIdeaNewTopic          idea (Maybe topic) (Maybe topic)
```

### DIFF
--- a/src/Action.hs
+++ b/src/Action.hs
@@ -26,7 +26,6 @@ module Action
     , ActionCurrentTimestamp(getCurrentTimestamp)
     , ActionSendMail
     , ActionAddDb
-    , ActionPhaseChange
     , ActionError
     , ActionExcept(..)
     , ActionEnv(..), envRunPersist, envConfig, envLogger
@@ -195,8 +194,6 @@ instance ThrowSendMailError ActionExcept where
 type ActionSendMail = HasSendMail ActionExcept ActionEnv
 
 type ActionAddDb m = (ActionUserHandler m, ActionPersist m, ActionCurrentTimestamp m)
-
-type ActionPhaseChange m = (ActionAddDb m, ActionSendMail m)
 
 type ActionM m =
       ( ActionLog m
@@ -419,7 +416,7 @@ createTopic proto = do
 editTopic :: ActionM m => AUID Topic -> EditTopicData -> m ()
 editTopic topicId topic = do
     update $ EditTopic topicId topic
-    eventLogUserEditsTopic =<< equery (maybe404 =<< findTopic topicId)
+    eventLogUserEditsTopic =<< mquery (findTopic topicId)
 
 createIdea :: Create Idea
 createIdea proto = do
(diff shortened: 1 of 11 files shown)


fisx:
ready for review and merging!

src/Action.hs
```diff
@@ -343,15 +345,15 @@ deleteUser = update . DeactivateUser
 
 -- * Phase Transitions
 
-topicPhaseChange :: (ActionPhaseChange m) => Topic -> PhaseChange -> m ()
```
andorp: Please delete the type synonym if not used anymore.

src/Action.hs
```diff
@@ -422,26 +416,56 @@ createTopic proto = do
     eventLogUserCreatesTopic topic
     pure topic
 
+editTopic :: ActionM m => AUID Topic -> EditTopicData -> m ()
+editTopic topicId topic = do
+    update $ EditTopic topicId topic
+    eventLogUserEditsTopic =<< equery (maybe404 =<< findTopic topicId)
```
andorp: Please use `mquery`.

src/Action.hs
```diff
@@ -422,26 +416,56 @@ createTopic proto = do
...
+
 
 -- * Vote Handling
 
-likeIdea :: ActionAddDb m => AUID Idea -> m ()
-likeIdea ideaId = currentUserAddDb_ (AddLikeToIdea ideaId) ()
+likeIdea :: ActionM m => AUID Idea -> m ()
+likeIdea ideaId = do
+    currentUserAddDb_ (AddLikeToIdea ideaId) ()
+    do idea <- equery $ maybe404 =<< findIdea ideaId
```
andorp: Please compute use one `equery` as it composes.

src/Action.hs
```diff
@@ -642,42 +665,47 @@ eventLogUserEditsTopic topic = do
...
 eventLogUserVotesOnComment ::
-      (ActionUserHandler m, ActionCurrentTimestamp m, ActionLog m)
-      => Idea -> Comment -> Maybe Comment -> UpDown -> m ()
-eventLogUserVotesOnComment idea comment mcomment v = do
-    uid <- currentUserId
+      (ActionUserHandler m, ActionCurrentTimestamp m, ActionPersist m, ActionLog m)
+      => KeyOf Comment -> UpDown -> m ()
+eventLogUserVotesOnComment ck@(CommentKey _ ideaId parentIds _) v = do
+    uid      <- currentUserId
+    idea     <- equery $ maybe404 =<< findIdea ideaId
```
andorp: It seems that this part should go under one `equery`.

src/Action.hs
andorp: Please do so in the subsequent occurrences too.

fisx:
I have followed your comments in `likeIdea`, but reject the one on `eventLogUserVotesOnComment`, for the code in the latter is only a work-around until the `EventLogUserVotesOnComment` constructor takes a comment key; see FIXME there.

PR #476, opened 2016-05-12
2015 05 cc responsiveness

localgrr:


fisx:
Definitely an improvement.  (:

- I still think it's weird that the writing in the buttons under comments in the idea detail view differs in size.
- "Prozessverwaltung" in the admin pages context menu header on the left of the page still gets chopped off when narrowing the window from "full-screen" towards "phone"

But both can be handled elsewhere.

Thanks!

PR #479, opened 2016-05-13
Library for delegated voting

fisx:


fisx:
putting this here to keep track of the status of the discussion.

@andorp assigned this to you in the hope you'll enjoy thinking about this.  It's only due in iteration 2, and we can talk about it before you spend serious time on it, and you can just un-assign you if you prefer.

src/Data/Voting.hs
```diff
@@ -0,0 +1,100 @@
...
+identity :: Voter -> Delegation Voter 'VoterKind 'VoterKind
+identity v = Delegation v v
+
+delegate :: Voter -> Voter -> Delegation Voter 'VoterKind 'VoterKind
+delegate from to = Delegation from to
+
+instance Eq v => Category (Delegation v) where
+    id = IdentityDelegation
+    NoDelegation . x = NoDelegation
+    x . NoDelegation = NoDelegation
```
fisx: why not?

```haskell
x . NoDelegation = x
```

can you explain into the source.

src/Data/Voting.hs
```diff
@@ -0,0 +1,100 @@
...
+    IdentityDelegation . (Delegation a b) = Delegation a b
+    (Delegation a b) . IdentityDelegation = Delegation a b
+    (Delegation b1 c) . (Delegation a b0)
+      | b0 == b1  = Delegation a c
+      | otherwise = NoDelegation
```
fisx: so `NoDelegation` is the miss of all partial functions here.  (not really a comment just thinking out loud.)

src/Data/Voting.hs
```diff
@@ -0,0 +1,100 @@
...
+        Nothing  -> Map.insert to (Set.singleton from) coDmap
+        Just to' -> let coDmap0 = Map.update (deleteValue from) to coDmap
+                        coDmap1 = case Map.lookup to' coDmap0 of
+                                    Nothing -> Map.insert to' (Set.singleton from) coDmap0
+                                    Just ds -> Map.insert to' (Set.insert from ds) coDmap0
+                    in coDmap1
+    deleteValue d ds = let ds' = Set.delete d ds in if Set.null ds' then Nothing else Just ds'
+
+canVote :: Voter -> DelegationMap -> Bool
+canVote v (DelegationMap dmap) = isNothing $ Map.lookup v dmap
```
fisx: the name is misleading: any voter can always vote.  if she has delegated away her vote before, that implicitly withdraws the delegation.  so this should be called `hasDelegated`, perhaps?

src/Data/Voting.hs
```diff
@@ -0,0 +1,100 @@
...
+
+getDelegators :: Voter -> CoDelegationMap -> [Voter]
+getDelegators v (CoDelegationMap dmap) = fix voters v
```
fisx: we should introduce and use the following terminology:

- delegate: someone who has been delegated somebody else's vote to.
- delegatee: someone who has delegated her vote to somebody else.

src/Data/Voting.hs
```diff
@@ -0,0 +1,100 @@
...
+  where
+    voters rec v = v : maybe [] (mconcat . map rec) (Set.toList <$> Map.lookup v dmap)
+
+data Votings = Votings {
+        votings :: Map Voter (Voter, Vote)
+    }
+
+setVote :: Voter -> Vote -> [Voter] -> Votings -> Votings
```
fisx: the two function names `setVote`, `vote` are confusing.  they do the same thing, but one of them is a helper for the other.  ideally, the helper should be defined in the `where` clause of the other.  if both are needed globally, one should be called something that makes it explicit that it's a "more internal" function.  `vote` and `vote'`?  Or `vote` and `voteInternal`?  or have an internal module for the latter and call it the same?  my preference is the `'`

src/Data/Voting.hs
```diff
@@ -77,24 +102,41 @@ setDelegation from to (Delegations (DelegationMap dmap) (CoDelegationMap coDmap)
...
-vote voter vote (Delegations dmap coDmap) votings
-  | canVote voter dmap = setVote voter vote (getDelegators voter coDmap) votings
-  | otherwise          = votings
+-- * monadic api
+
+class Monad m => DelegationM m where
+    setDelegation :: Voter -> Voter -> m ()
+    canVote       :: Voter -> m Bool
+    getDelegators :: Voter -> m [Voter]
+    voteFor       :: Voter -> Vote -> Voter -> m ()
```
fisx:
```haskell
delegateTo ...
isDelegatee ...
isDelegate ...
```

(on second thoughts, `delegatee` is not much better than `delegator`.  undecided on this one.)

fisx: let me try that again.

```haskell
delegateTo ...
delegateeOf ...
delegateOf ...
voteFor ...
```

not entirely certain what `voteFor` does, but i suspect i might want it to be hidden deeper in this module, not exposed in the monadic api.

src/Data/Voting.hs
```diff
@@ -50,12 +52,13 @@ instance Eq v => Category (DCat v) where
...
       | otherwise = NoDelegation
 
+
 -- * pure model
 
 data DelegationMap = DelegationMap {
         _delegationMap
             :: Map
-                Voter -- who delegates
+                (Voter, Topic) -- who delegates in which topic
```
fisx: s/Topic/Scope/?  scopes extend down-path: if you delegate in a topic, you delegate in all ideas in that topic (unless something overrules there).

src/Data/Voting.hs
```diff
@@ -62,10 +62,18 @@ instance Eq v => Category (DCat v) where
 data DelegationMap = DelegationMap {
         _delegationMap
             :: Map
-                (Voter, Topic) -- who delegates in which topic
-                Voter -- to whom delegates
+                Voter -- who delegates in which topic
+                (Map Topic Voter) -- to whom delegates
```
fisx: comments need update here.  this great evidence that comments should *never* carry information, because it'll be *always* outdated.  (-:

src/Data/Voting.hs
```diff
@@ -109,19 +117,21 @@ setDelegationPure from topic to (Delegations (DelegationMap dmap) (CoDelegationM
...
-delegatedForTopic from topic to (DelegationMap dmap) =
-    maybe False (to ==) $ Map.lookup (from, topic) dmap
+delegatedForIdea :: Voter -> Idea -> Voter -> DelegationMap -> Bool
+delegatedForIdea from idea to dmap =
+    maybe False (to ==) $ lookupDMap from (TopicIdea idea) dmap
+
 
-canVotePure :: Voter -> Idea -> DelegationMap -> Bool
-canVotePure v i (DelegationMap dmap) = isNothing $ Map.lookup (v, (Topic i)) dmap
+canVotePure :: Voter -> Idea -> DelegationMap -> TopicTree -> Bool
```
fisx: shouldn't be `TopicTree`, but `TopicDag`!

fisx:
I'm intrigued.  Some of my comments are answered by subsequent commits.

fisx:
possibly relevant: http://hackage.haskell.org/package/ixset-typed

src/Data/Voting.hs
```diff
@@ -0,0 +1,518 @@
...
+
+getVotePure :: Voter -> Idea -> Votings -> Maybe (Voter, Vote)
+getVotePure v i (Votings vmap) = Map.lookup (v,i) vmap
+
+-- * monadic api
+
+class Monad m => DelegationM m where
+    setDelegation :: Voter -> Topic -> Voter -> m ()
+    canVote       :: Voter -> Idea  -> m Bool
+    getSupporters :: Voter -> Topic -> m [Voter]
```
np: `getSupporters :: Voter -> m (Map Topic [Voter])` ?

fisx: i guess we need to look at the way this called.  my expectation is that andor's version is more convenient, but i don't trust my expectation that much here.

andorp:
Closing in favour of #615

PR #480, opened 2016-05-13
Shifting phases

### LINKED ISSUE #472: better status message when moving phases in the admin pages
currently, when moving won't work, we still get a success message.  ideally, all moves should work, actually, but if the shift is not successful, that should be transparent to the user.

fisx:
Fixes #472, but does more.

Most notably:
- we can now shift back through all phases from result to refinement
- jury statements are not removed any more when switching back to jury phase.  (i think this means that the next edit will trigger another phase shift.  we should think about this some more.)

### DIFF
--- a/src/LifeCycle.hs
+++ b/src/LifeCycle.hs
@@ -36,7 +36,6 @@ import Types
 
 data PhaseChange
     = RefinementPhaseTimeOut
-    | RefinementPhaseMarkedByModerator
     | AllIdeasAreMarked { _phaseChangeVotPhaseEnd :: Timestamp }
     | VotingPhaseTimeOut
     | VotingPhaseSetbackToJuryPhase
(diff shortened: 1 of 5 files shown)


fisx:
it turns out that shifting back from *any* phase was implemented as follows:

- check if we're in voting phase
- if yes, move back to jury phase
- if no, do nothing

i will replace `topicInVotingResetToJury` with something more general-purpose now.

src/LifeCycle.hs
```diff
@@ -70,8 +69,6 @@ thawPhase now = (phaseStatus     %~ thawStatus)
 phaseTrans :: Phase -> PhaseChange -> Maybe (Phase, [PhaseAction])
 phaseTrans (PhaseRefinement ActivePhase{}) RefinementPhaseTimeOut
     = Just (PhaseJury, [JuryPhasePrincipalEmail])
-phaseTrans (PhaseRefinement ActivePhase{}) RefinementPhaseMarkedByModerator
```
andorp: Removing this does not do any harm. But it is defined in the requirements:
https://media.taiga.io/attachments/3/c/3/3/f2ddcaf7f28d89e2cfccdc793c519951a79b17c9797e84ee01b6ad195bab/160212_aula_process_overview_user_flows.pdf

    NTH: if a global switch is set, instead of
    a fixed duration for the phase, the
    moderator will mark a topic as done so
    it can move to the jury phase"

fisx: thanks, that's a good point.  the above constructor didn't implement this requirement, though, hence removing it strictly improves the situation (less dead code, same feature coverage).

i've implemented this feature (in a slightly liberal interpretation, plus the analogous one for voting phase) in #483.  took 3'39'' with testing.  (-:

Mikolaj:
LGTM, as well.

PR #484, opened 2016-05-16
Make active tab/menu selection more type safe.

fisx:
I don't have a strong opinion here, but this rules out a bug that was expensive to fix (see https://github.com/liqd/aula/pull/471/commits/e45d958af4f5e4f72a0e75e09c48269237bbfaae).


fisx:
The commit referenced in the PR description had nothing to do with this PR.  I fixed that, now it points to the fix of the issue that motivated #484.

fisx:
much simpler!  (-:

fisx:
thx

PR #488, opened 2016-05-17
Finishing touch at "Refactor: phase change actions"

np:


fisx:
yes, that makes sense.  thanks!

PR #490, opened 2016-05-17
Misc capability fixes.

fisx:


src/LifeCycle.hs
```diff
@@ -219,27 +219,32 @@ isCreatorOf u = (u ==) . view createdBy
...
   deriving (Enum, Eq, Ord, Show, Read, Generic)
 
 instance SOP.Generic CommentCapability
 
-canDeleteComment :: AUID User -> Role -> Comment -> Bool
-canDeleteComment uid role comment = uid `isCreatorOf` comment || role == Moderator
-
-commentCapabilities :: AUID User -> Role -> Comment -> [CommentCapability]
-commentCapabilities uid role comment
+commentCapabilities :: AUID User -> Role -> Comment -> Phase -> [CommentCapability]
```
andorp: When the `commentCapabilities` are getting more complicated we should change to the tabular form as we did with the ideas.

andorp:
LGTM

PR #491, opened 2016-05-17
Background phase transition

np:


src/Persistent/Idiom.hs
```diff
@@ -157,13 +156,8 @@ saveAndEnactFreeze :: Timestamp -> Freeze -> AUpdate ()
...
   dbFreeze .= shouldBeFrozenOrNot
   let change = case shouldBeFrozenOrNot of
-          NotFrozen -> PhaseThaw   now
-          Frozen    -> PhaseFreeze now
-      freezeOrNot topic = do
-          case phaseTrans (topic ^. topicPhase) change of
-            Nothing -> return ()  -- ignored, probably repeated freeze or thaw
-            Just (phase', actions) -> do
-                let !_A = assert (null actions) ()  -- wrong monad to execute'em
-                setTopicPhase (topic ^. _Id) phase'
```
fisx: so, the assertion is now eliminated?  i think i'll try to re-add it, hope that's ok?

fisx: hum, you have very thoroughly removed that option, it seems.

np: yes the assertion is no longer necessary

fisx:
LGTM!  i reverted the removal of the assertion and did the usual nit-picking.  will merge later tonight if it's still open and green, and nobody objects.

Thanks!

PR #494, opened 2016-05-17
Show "create topic" button in idea if quorum is reached.

fisx:


src/Frontend/Page/Idea.hs
```diff
@@ -210,7 +210,14 @@ instance ToHtml ViewIdea where
 
             when (has _PhaseWildIdea phase && ideaReachedQuorum ideaInfo) $ do
                 -- FIXME: design; see https://marvelapp.com/ehhb43#10108433
-                div_ [class_ "voting-buttons"] "Idee kann auf den Tisch."
+                div_ [class_ "voting-buttons"] $
+                    if role == Moderator
```
andorp: Let use the `ideaCapabilities` here.

andorp:
LGTM

src/Frontend/Page/Idea.hs
```diff
@@ -211,7 +211,7 @@ instance ToHtml ViewIdea where
             when (has _PhaseWildIdea phase && ideaReachedQuorum ideaInfo) $ do
                 -- FIXME: design; see https://marvelapp.com/ehhb43#10108433
                 div_ [class_ "voting-buttons"] $
-                    if role == Moderator
+                    if CanCreateTopic `elem` userCaps
```
fisx: how did i get that wrong?!  thanks for fixing it!

PR #504, opened 2016-05-18
Password validation is only needed on new passwords, remove it on current password (fixes #492)

### LINKED ISSUE #492: Password verification tweaks.
- passwords shouldn't be checked for allowed length etc. in login form (otherwise, if length requirements change, some passwords in the DB won't work any more).
- demo passwords should satisfy the allowed length etc. just like the passwords entered in 'change pass'.

np:


andorp:
LGTM

np:
Can you have a look at the second commit.

andorp:
LGTM

PR #523, opened 2016-05-18
Apply custom ordering of idea spaces.

andorp:


src/Frontend/Page/Overview.hs
```diff
@@ -26,7 +29,6 @@ import Persistent (findWildIdeasBySpace, getListInfoForIdea, findTopicsBySpace)
 import qualified Data.Text as ST
 import qualified Frontend.Path as U
 
-
```
andorp: Sorry :)

fisx:
LGTM.  We could make a `ST`-based implementation for perormance, but I'd like to do that only when benchmarks support it's a good idea.

I'll merge when travis is green.

andorp:
LGTM

PR #526, opened 2016-05-19
Type safety for paths

andorp:
Smart constructors should be introduced for every path we use as to achieve the full type safeness. I would do that in a different PR to minimize the size of the lousy coupled changes.


fisx:
Perfect!  (-:

fisx:
(we *could* do something like `data AllowMethods = Get | Post` and then `Main '[Get, Post'`, but we can think about that when resolving the resp. FIXME.)

Issue #517, opened 2016-05-18
Deleting ideas 


rittermo:
On the Edit idea page: 
* Deleting ideas does not work (it redirects to the idea detail page and the idea still exists) 
* The delete button should ask for a confirmation before deleting the idea

PR #529, opened 2016-05-19
Delete idea.

andorp:
Fixes: #517


fisx:
I haven't really looked at the code yet, but I think what it currently does is not only to remove the user-contributed content, but the entire idea detail view page.

The idea behind the `delete` button is to remove offensive content entered by students, without doing too much damage to harmless or constructive content.  So I think what we want to do is just hide the title and description of the idea, and keep showing everything else even if the idea is deleted.

@rittermo what's your call?

rittermo:
@fisx No, a moderator could already edit the title and the description to do that. We need a way of completely deleting ideas too. This could be useful for test content, unintended duplicates etc.

fisx:
@rittermo it won't be very useful for that without considerable extra work.  also, it would introduce an assymetry between deleting comments and deleting ideas which i find confusing.  my understanding was that deleting content is needed because content can be offensive, and we're ready for that.  

if we start thinking about content in the aula system as not append-only, but disappearable, things will get a lot more complicated everywhere.

are you sure this is a requirement?

andorp:
I see two solutions here:

 * Mark idea as deleted. Do not list that idea anymore in the idea lists and do not render the title and the description of that idea, if we open it somehow, **but display** the related comments.
 * Mark idea as deleted. Do not list that idea anymore in the idea lists and **do not render** the title and the description and comments of that idea.

Both cases we just mark the idea, and we don't remove it from the database. That is my preference as well.

fisx:
in conclusion:

- deleted ideas are not accessible via any idea list, whether in topics, wild idea lists, user profiles, or in some place in the system we haven't thought of here.
- when accessing an idea via the idea we've written down somewhere and the manually entered URL, the page will show a deleted idea, not a 404.  (this can also happen if student opens a topic, moderator deletes an idea from it, and student selects that idea, in that order.)
- i personally like that page to still list the comments, @rittermo and @andorp like the page to contain no data except "it's gone".  let's do the latter.

fisx:
Two issues:

1. after positive confirmation, the idea gets deleted, but the redirect doesn't work.
1. we can still edit deleted ideas.  the edit page should look the same as the view page on deleted ideas.

I will look into 1.

fisx:
i think i've addressed everything except for the js redirect.  will look at that later.

fisx:
Ready for merge IMO.

@andorp please take a look at my changes and merge if you're happy.

andorp:
LGTM

PR #531, opened 2016-05-19
Refactor freeze

### LINKED ISSUE #496: Refactor saveAndEnactFreeze
This has been done in 90272b4 in #491, but I'm not convinced that the `assert` is really not necessary.  Therefore, the PR has this commit reverted.

We should add a quickcheck property that replaces it, then cherry-pick this commit and use it in the production .code.  It's much nicer!

fisx:
Fixes #496.

### DIFF
--- a/src/LifeCycle.hs
+++ b/src/LifeCycle.hs
@@ -24,6 +24,7 @@ module LifeCycle
     )
 where
 
+import Control.Exception
 import Control.Lens
 import Data.Monoid
 import GHC.Generics (Generic)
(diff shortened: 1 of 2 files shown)


Mikolaj:
Yes, I think in the current setup, this is dead code. I think `PhaseFreeze` can be removed.

fisx:
Good point!  I'm getting more and more comfortable with this change.  (-:

andorp:
LGTM

Mikolaj:
(lgtm) (forgot to say so long time ago)

Issue #536, opened 2016-05-20
report user profiles


fisx:
like comments and ideas, user profiles (descriptions and avatar images) can contain offending content.  the user profile needs a report button for everybody.

fisx:
related: #517, #514, #537

PR #540, opened 2016-05-20
Report User

andorp:
Fixes: #536

### DIFF
--- a/src/Action/Smtp.hs
+++ b/src/Action/Smtp.hs
@@ -10,7 +10,8 @@
 {-# LANGUAGE TemplateHaskell        #-}
 
 module Action.Smtp
-    ( EmailMessage(..), msgSubject, msgBody, msgHtml
+    ( EmailMessage(..), msgSubjectLabel, msgSubjectText, msgBody, msgHtml
+    , EmailSubjectLabel(..)
     , HasSendMail(sendMailToAddress)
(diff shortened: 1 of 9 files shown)


src/Action/Smtp.hs
```diff
@@ -98,6 +104,9 @@ sendMailToAddressIO logger receiver msg = do
         Left (e :: IOException) ->
             throwSendMailError $ IOErrorRunningSendMail e
   where
+    subjectLabel = case msg ^. msgSubjectLabel of
+        IdeaSpaceSubject is -> cs $ showIdeaSpace is
+        UserLoginSubject ul -> ul ^. unUserLogin
```
fisx: two throughts (no need to act on them):

1. I think like `where` only if it is used in a function that has a very small body (3 lines at most) and does everything in the `where` block.  In this case, I prefer `let`, because it lets the author write things down in the order in which they need to be read in order to be understood.
1. I would have instantiated `HasUILabel` with `EmailSubjectLabel`, but arguably this is the only place where it's ever needed, so I'm not sure about this any more.

fisx:
LGTM, manual test works, and I like the email subject refactoring!

Will merge when green.

src/Action/Smtp.hs
andorp: 1. I agree. I will keep that in mind.
2. It is only used here and in emails, no UI element is rendered from it.

PR #550, opened 2016-05-23
Event log

### LINKED ISSUE #216: Event log

fisx:
Fixes #216


tests/Persistent/ApiSpec.hs
```diff
@@ -100,7 +102,12 @@ addDbSpecProp imp name getXs addX propX =
 
 addDbSpec :: (Foldable f, Arbitrary proto, HasAUpdate ev a)
           => PersistenceImpl -> String -> Query (f a) -> (EnvWith proto -> ev) -> Spec
-addDbSpec imp name getXs addX = addDbSpecProp imp name getXs addX (\_ _ -> passes)
+addDbSpec imp name getXs addX = addDbSpecProp imp name getXs addX id (\_ _ -> passes)
+
+-- | like 'addDbSpec', but takes a function that strips extra info from the translates result value.
```
andorp: Probably typo `translates`?

src/Persistent/Pure.hs
```diff
@@ -542,7 +556,7 @@ moveIdeaToTopic ideaId mTopicId =
 setTopicPhase :: AUID Topic -> Phase -> AUpdate ()
 setTopicPhase tid phase = withTopic tid . topicPhase .= phase
 
-addTopic :: Timestamp -> AddDb Topic
```
fisx: i'll add a comment here explaining things.

fisx: Also, this should keep the same type, and we should add a function `addTopicYieldLocs`.  That will make lots of the follow-up changes unnecessary.

Thanks, @andorp for the idea!

andorp:
LGTM

PR #552, opened 2016-05-23
Fix not feasable icon

### LINKED ISSUE #532: Icon size mismatch in idea list
(Make sure #510 is merged if you process this)

2016-05-19-180320_1600x900_scrot

The `feasible` has the right size, the `not feasible` is too large.  Would be nice if the table could be made a bit bigger.

(I artificially added lots of icons, so it's ok that they don't fit into one row any more.)

localgrr:
Fixes: #532


fisx:
LGTM.  please merge when green.

fisx:
github won't let me remove the label.  :-(

PR #553, opened 2016-05-23
Phase timeout info on view topic page.

### LINKED ISSUE #511: Tweak timeout phase transitions.
- add "and at <exact time>" to topic's phase info.
- check for off-by-one errors: i think i've seen "this phase is still running for -1 days".
- ~~change timeout interval for phase transition to "always around midnight CEST".~~

fisx:
Fixes #511.

### DIFF
--- a/src/Config.hs
+++ b/src/Config.hs
@@ -34,6 +34,8 @@ module Config
     , logging
     , logLevel
     , eventLogPath
+    , unsafeTimestampToLocalTime
+    , aulaTimeLocale
     )
 where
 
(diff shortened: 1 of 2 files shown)


andorp:
* The time is displayed in UTC, we should use the local time zone.
 * I think seconds should not be displayed.

fisx:
I refuse to eliminate the `unsafePerformIO` call in this case!  (-:

andorp:
LGTM, but I open an issue :)

PR #555, opened 2016-05-23
Reset Password Page

### LINKED ISSUE #190: Reset Password Page
* There is no design for the reset password page.
 * There is no exact description how the reset password should work.
 * What should happen if we reset a password for a user who has no email address?

andorp:
Fixes #190


src/Frontend/Page/Admin.hs
```diff
@@ -828,6 +837,36 @@ instance FormPage AdminPhaseChange where
...
+
+    -- TODO: Styling
+    -- TODO: Translation
+    formPage v form p@(PageAdminResetPassword usr pwd) = adminFrame p . semanticDiv p $ do
+        h3_ "Reset password for user"
+        form $ do
+            div_ $ do
+                p_ . toHtml $ "Reset password for user: " <> usr ^. userLogin . unUserLogin
+                p_ . toHtml $ "New password:" <> (pwd ^. unInitialPassword)
+                p_ "Are you sure???"
```
fisx: Please don't use punctuation like this.  It hurts my ears.  (-:

fisx:
LGTM!  Please merge when green.

PR #559, opened 2016-05-24
Improved quorum table icons fixes #519

### LINKED ISSUE #519: Add icon for ideas that have reached the quorum
If an idea has reached the quorum, a table icon should appear with the label “kann auf den Tisch” on the idea item in topic overview and the idea detail page. (see here https://marvelapp.com/ehhb43#10108424 for topic overview and here https://marvelapp.com/ehhb43#10108433 idea detail page)
A div is already rendered by Haskell.

localgrr:


fisx:
LGTM, except for one small glitch and the usual nit-picks.  please merge when green!

PR #566, opened 2016-05-24
Made form errors more prominent fixes #502

### LINKED ISSUE #502: Better display of form validation errors
Currently validation errors are displayed very bleakly at the beginning of the form. This should be changed to either 

a) (The nicest version in terms of UX Design) Display the validation error next to form field that is concerned and make them and the form field stand out visually by adding a red color for the text and possibly a red border for the form field.

b) (Probably the easier solution but not so nice in terms of UX) Display a list of validation errors at the beginning of the form as a list in red color and give the list a label/headline like “Es traten folgende Fehler in den Eingaben auf:”

localgrr:


fisx:
nice.  it looks so easy. (-:

fisx:
LGTM.

localgrr:
Too easy B)

PR #569, opened 2016-05-24
Mark only non marked ideas.

andorp:


src/Action.hs
```diff
@@ -446,7 +446,8 @@ topicForcePhaseChange dir tid = do
   where
     -- this implicitly triggers the change to voting phase.
     makeEverythingFeasible topic = do
-        ideas <- query $ findIdeasByTopic topic
+        let nonMarkedAsFeasible = not . has (ideaJuryResult . _Just . ideaJuryResultValue . _Feasible)
```
fisx: This overwrites all ideas marked as `NotFeasible`, so all negative decisions by the dean will still be overwritten.  What we want is that all decision by the dean are accepted.

fisx:
Looks good (almost :-).  If you fix the one bug and travis likes it, please merge.

fisx:
Perfect!

Issue #545, opened 2016-05-23
LifeCycle needs refactoring


fisx:
https://github.com/liqd/aula/blob/d6ed640f37107db408355a073e5c9b798f07a538/src/Frontend/Page/User.hs#L186

Issue #487, opened 2016-05-17
Disable some buttons / menu entries / .... by gray-out instead of by hiding.


fisx:
This will avoid the situation where users look for a button that they know has been somewhere, because they will find it and know it's disabled instead of lost.

Probably works best on things disabled because of the situation (quorum not reached, wrong phase, ...), and not because of the missing role.

fisx:
i guess what we can do here: @np changes the `when <something about caps...> $ do ...` into code that conditionally attributes a new css class to the grayed-out menu entries, and then @localgrr can change the name of the css class to something that works for her and define its layout?

fisx:
#572 is put on hold for now (see discussion there).

PR #570, opened 2016-05-25
Lifecycle refactor

fisx:
Could be a solution to #487.
Please look at this after #545 has been resolved.


fisx:
2016-05-25-100637_1600x900_scrot

fisx:
See also: 2016-05-24-ap-inactive-buttons

PR #572, opened 2016-05-25
Life cycle refactor 2

andorp:


andorp:
I think this functionality should be discussed further.

src/LifeCycle.hs
```diff
@@ -137,9 +150,24 @@ userCapabilities = \case
...
-
-ideaCapabilities :: AUID User -> Role -> Idea -> Phase -> [Capability]
-ideaCapabilities = phaseCap
+ideaCapabilities :: AUID User -> Role -> Idea -> Phase -> [Clickable Capability]
+ideaCapabilities u r i p =
+    let activeCaps = nub $ ideaCapabilitiesInPhase u r i p
+        allCaps    = nub $ ideaCapabilitiesInAllPhases u r i
+    in (Clickable <$> activeCaps) <> (GrayedOut <$> (allCaps \\ activeCaps))
+
+-- ASSUMPTION: The list of a phase should be part of the AllPhase list.
```
fisx: Can you rephrase this assumption?  My best guess is you mean this (but I suspect I got it wrong):

```haskell
-- The union of the dea caps of all phases.
```

tests/LifeCycleSpec.hs
```diff
@@ -0,0 +1,34 @@
...
+import LifeCycle
+
+
+spec :: Spec
+spec = do
+    tag Large . it "sameItemDoesNotAppear as clickable and grayed out in ideaCapabilities" . property .
+        forAllShrinkDef arb $ \userId ->
+        forAllShrinkDef arb $ \role ->
+        forAllShrinkDef arb $ \idea ->
+        forAllShrinkDef arb $ \phase ->
```
fisx: you can probably make this one function in a 4-tuple?  if you want to, you can define `curry4` in `Types` to make this more compact.  (-:

fisx: (`uncurry4`?  never get these two right the first time..)

src/Frontend/Page/Idea.hs
```diff
@@ -169,6 +169,13 @@ linkToIdeaLocation idea = do
              IdeaLocationSpace{} -> "Zum Ideenraum"
              IdeaLocationTopic{} -> "Zum Thema"
 
+-- ASSUMPTION: The caps list does con contain the same value marked as
+-- Clickable and GrayedOut
```
fisx: i would prefer if this assumption is encoded as an assertion.  (can the test suite then use this assertion to fail in case of error, rather than their independent test function?)

fisx: i tried this: a4ca3e7

fisx: Also, I'm wondering if there is a better name?  `canGrayOut`?  No, that's no better.  I appreciate the rule that eliminators are called like the type, `maybe` does it, too. I just still find it hard to read.

fisx:
Let's discuss my comments and changes off-line, but I think this can be merged soon.

src/LifeCycle.hs
andorp: thanks

tests/LifeCycleSpec.hs
andorp: I think in this form we run 4x more tests, but I am not insisting on that.

fisx: not important, leave it.

src/Frontend/Page/Idea.hs
fisx: i like, in this order:

1. eitherClickable
1. eitherClickableGrayedOut
1. clickable

all are ok for me.

fisx:
@andorp i've played with your changes, add a few more, and opened #579.  could you take a last look at my commits here and object or merge?  thanks!

fisx:
Ok, changed my mind: this is making the code base too complicated for too little benefit.  If we want to have a bad solution for some grayed-out UI elements here and there, we should aim for a *simple*, *local* bad solution, not something striving for generality and beauty unsuccessfully.

I will close the issue for now, but leave the PR open in the hope we'll find the time to think of something better.

andorp:
Closing this, as we closed other experiments in the past.

PR #581, opened 2016-05-26
Misc fixes.

fisx:


andorp:
I have a small concert. As `CanVote` introduced, there are three interpretation for it. One in idea context, one in topic context, and one in user context. When we take the union of those contexts thing could go wrong, but it is still not the case.

src/LifeCycle.hs
```diff
@@ -127,7 +127,7 @@ instance SOP.Generic Capability
 
 userCapabilities :: Role -> [Capability]
 userCapabilities = \case
-    Student    _clss -> []
+    Student    _clss -> [CanVote]
```
fisx: ok, thanks.

this makes me think perhaps we should merge all `*Capabilities` functions into one after all that takes a context of super-type of all individual functions.  that would rule out issues like this.

andorp:
LGTM

PR #592, opened 2016-05-27
Simplify caps computation.

### LINKED ISSUE #582: LifeCycle: merge `*Capabilities` functions into `capabilities`.
- this will reduce the risk of missing to call one of the four accidentally.
- we should also unravel the functions that are not in decision table shape.
- it should be difficult to miss where to look for a certain capability in a certain context.
- see also: discussion in #581.

fisx:
Fixes #582.

### DIFF
--- a/src/Frontend/Fragment/VotesBar.hs
+++ b/src/Frontend/Fragment/VotesBar.hs
@@ -44,7 +44,7 @@ instance ToHtml IdeaVoteLikeBars where
                         , capCtxComment = Nothing
                         }
                 \\ case mode of
-                    IdeaVoteLikeBarsPlain       -> [CanLike, CanVote]
+                    IdeaVoteLikeBarsPlain       -> [CanLike, CanVoteOnIdea]
                     IdeaVoteLikeBarsWithButtons -> []
 
             likeBar :: Html () -> Html ()
@@ -108,7 +108,7 @@ instance ToHtml IdeaVoteLikeBars where
             user = ctx ^. renderContextUser
 
             voteButtons :: Html ()
-            voteButtons = when (CanVote `elem` caps) .
+            voteButtons = when (CanVoteOnIdea `elem` caps) .
                 div_ [class_ "voting-buttons"] $ do
                     voteButton vote Yes "dafür"
                     voteButton vote No  "dagegen"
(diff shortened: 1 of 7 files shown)


src/Frontend/Fragment/VotesBar.hs
```diff
@@ -36,11 +36,13 @@ minBarSegmentWidth = 5
...
-                        (ctx ^. renderContextUser . userRole)
-                        idea
-                        phase
+        let caps = capabilities CapCtx
+                        { capCtxRole    = ctx ^. renderContextUser . userRole
+                        , capCtxPhase   = Just phase
+                        , capCtxUser    = Just $ ctx ^. renderContextUser . _Id
+                        , capCtxIdea    = Just idea
+                        , capCtxComment = Nothing
+                        }
```
andorp:
```haskell
phaseVotingCap :: Idea -> Role -> [Capability]
phaseVotingCap i = \case
    Student    _clss -> [CanVote | isFeasibleIdea i]

topicVotingCaps :: Role -> [Capability]
topicVotingCaps = \case
    Student    _clss -> [CanVote]

userCapabilities :: Role -> [Capability]
userCapabilities = \case
    Student    _clss -> [CanVote]
````

Resulting:
The student always can Vote on an idea even if is not feasible.

**Comment by fisx** [comment]
can i have this back?

File: src/LifeCycle.hs [hunk 2]
```diff
@@ -96,7 +96,9 @@ phaseTrans _ _ = Nothing
 data Capability
     -- Idea
     = CanLike
-    | CanVote
+    | CanVoteInGeneral
+    | CanVoteOnIdea  -- can vote on one idea
+    | CanVoteOnTopic -- can vote on one topic
```
fisx: `CanVoteInGeneral` is a super-type of the other two.  I wonder if there is a way to avoid that?

src/LifeCycle.hs
```diff
@@ -96,9 +96,7 @@ phaseTrans _ _ = Nothing
 data Capability
     -- Idea
     = CanLike
-    | CanVoteInGeneral
-    | CanVoteOnIdea  -- can vote on one idea
-    | CanVoteOnTopic -- can vote on one topic
+    | CanVote
```
fisx: ah, there is!  (-:

src/LifeCycle.hs
```diff
@@ -139,11 +139,11 @@ capabilities ctx = mconcat $
...
-    r  = ctx ^. to capCtxRole
-    mp = ctx ^. to capCtxPhase
-    mu = ctx ^. to capCtxUser
-    mi = ctx ^. to capCtxIdea
-    mc = ctx ^. to capCtxComment
+    r  = capCtxRole    ctx
+    mp = capCtxPhase   ctx
+    mu = capCtxUser    ctx
+    mi = capCtxIdea    ctx
+    mc = capCtxComment ctx
```
fisx: fair enough.  (-:

fisx:
LGTM.  Thanks for fixing this!

PR #593, opened 2016-05-28
Voting logic

### LINKED ISSUE #512: Voting phase needs clarification
1. in voting phase and result phase, votes should be sorted by support.  however, support is not defined based on quorum votes (aka likes) any more, but on votes.  there is a function that gives a number to the support of an idea based on numbers of yes votes, no votes, and voters, resp.  a first shot at this function could be `#yes / #no`
1. what should the final function be here?
1. is "mark as winner" going away entirely?
1. can creator's statements be entered for non-winning ideas?

fisx:
Fixes #512.

### DIFF
--- a/src/Persistent/Idiom.hs
+++ b/src/Persistent/Idiom.hs
@@ -166,25 +166,23 @@ countVotes v = length . filter (has v) . map (view ideaVoteValue) . Map.elems . 
 -- "accepted" (automatically determined) and "winning" (decided by moderator) is necessary because
 -- two accepted ideas may be contradictory.  Identifying such contradictions requires language
 -- skills beyond those of the aula system.
-ideaAccepted :: AUID Idea -> EQuery Bool
+ideaAccepted :: IdeaStats -> Bool
 ideaAccepted = ideaAcceptedByMajority
 
-ideaAcceptedByMajority :: AUID Idea -> EQuery Bool
-ideaAcceptedByMajority iid = do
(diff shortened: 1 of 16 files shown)


fisx:
I will merge this to make a new release; @andorp, could you review post mortem? Thanks!

src/Persistent/Idiom.hs
```diff
@@ -167,22 +167,15 @@ countVotes v = length . filter (has v) . map (view ideaVoteValue) . Map.elems .
...
-ideaAcceptedByMajority :: IdeaStats -> Bool
-ideaAcceptedByMajority (IdeaStats idea _ _ numVoters) = nyes > nno && ntotal >= quo
+ideaAccepted (IdeaStats idea _ _ numVoters) = _ideaAcceptedByMajority
   where
-    quo    = numVoters `div` 3
-    nyes   = countVotes _Yes idea
-    nno    = countVotes _No idea
-    ntotal = nyes + nno
+    _ideaAcceptedByMajority = nyes > nno && ntotal >= quo
+    _ideaAcceptedByQuorum   = nyes >= quo && nno < quo
```
np: Is it because we don't know yet what to choose?

np:
LGTM, eventually it would be nice to have a separate module holding only the voting core rules.

fisx:
yes to the separate module.

re. the different quorum rules: yes, they were options discussed with the client, and I decided to implement both of them so we won't forget about it.  i'll add a comment.

PR #611, opened 2016-05-30
Sanitize HTML in forms accepting markdown.

### LINKED ISSUE #469: Markdown fields escape HTML
I think i confirmed that the backend is making EG ```<haskell> in to &lt;haskell&gt;```

fisx:
Fixes #469.


fisx:
this is fun!  i found two issues just by making `Document` abstract and following the type errors.  (-:

fisx:
ready for review.  this got slightly bigger than i hoped, but i think we can use this as is for the time being.

src/Data/Markdown/HtmlWhiteLists.hs
```diff
@@ -0,0 +1,982 @@
...
+      "TR",
+      "TT",
+      "U",
+      "UL",
+      "VAR"
+      ],
+
+  "denied": [
+      { "key": "APPLET",
+        "reason": "disallow because allows scripting" },
```
andorp: Is is here for documentation purposes?

fisx: yes i assume so.  this is just copied verbatim from caja, and i'm only using the white lists.  as i indicate in the haddock of `Markdown` though, it is possible that there are some subtleties in the semantics of these json files that i've missed.

PR #612, opened 2016-05-31
Authorization

np:


src/Arbitrary.hs
```diff
@@ -216,39 +219,40 @@ instance Arbitrary ViewIdea where
...
-    arbitrary = CreatorStatement <$> arb
-    shrink (CreatorStatement x) = CreatorStatement <$> shr x
+    arbitrary = garbitrary
+    shrink    = gshrink
 
 instance Arbitrary ReportComment where
-    arbitrary = ReportComment <$> arb
-    shrink (ReportComment x) = ReportComment <$> shr x
+    arbitrary = garbitrary
+    shrink    = gshrink
```
fisx: this is nicer, but it may reduce test coverage, because we are not very confident with the `garbitrary` implementation (see issue over at sop-basic).  i would prefer to keep the ugly cases here because they are less likely to cause follow-up work.

np: I only added those when I needed to change them because I was adding arguments. I agree with your point but I would consider this to be a fair use.

np:
I've rebased once more.

fisx:
result of off-line conversation: missing:

1. tests
2. idea space access (student in class 7a must not see ideas in 6a).

src/Frontend/Core.hs
```diff
@@ -221,52 +235,113 @@ percentVotes idea numVoters vv = {- assert c -} v
...
-  deriving (Eq, Read, Show)
-
--- | Calculates the render context for role sensitive page rendering
-renderContext :: (ActionPersist m, ActionUserHandler m) => m RenderContext
-renderContext = RenderContext <$> currentUser
-
-
 -- * pages
 
+data AccessResult
```
fisx: can you shuffle the definitions in this section to tell a story and match the export list?

src/Frontend/Core.hs
```diff
@@ -221,52 +235,113 @@ percentVotes idea numVoters vv = {- assert c -} v
...
+  = AccessGranted
+  | AccessDenied { _accessDeniedMsg :: ST, _accessDeniedRedirect :: Maybe URL }
```
fisx: two constructors `AccessDenied`, `AccessDeniedRedir` instead of the maybe?  no strong opinion here, just an idea.

src/Frontend/Core.hs
```diff
@@ -221,52 +235,113 @@ percentVotes idea numVoters vv = {- assert c -} v
...
+accessDeferred = pure AccessDeferred
+
+redirectLogin :: Applicative m => m AccessResult
+redirectLogin = accessDenied "Not logged in" $ Just P.Login
 
-instance Page ST where
-    isPrivatePage _ = True -- safer default, might need to be changed if needed
+userPage :: Applicative m => AccessInput any -> m AccessResult
+userPage LoggedIn{}  = accessGranted
+userPage NotLoggedIn = redirectLogin
```
fisx: rename to `pageForLoggedIn`?  it's more clumsy, but some pattern in the names would be nice.  (hence also: `pageForAdmin` etc.)

src/Frontend/Core.hs
```diff
@@ -221,52 +235,113 @@ percentVotes idea numVoters vv = {- assert c -} v
...
+pageForRole _ NotLoggedIn = redirectLogin
+
+publicPage :: Applicative m => any -> m AccessResult
+publicPage _ = accessGranted
+
+adminPage :: Applicative m => AccessInput any -> m AccessResult
+adminPage = pageForRole Admin
+
+authNeedPage :: Applicative m => (p -> m AccessResult) -> AccessInput p -> m AccessResult
+authNeedPage withPage = \case
```
fisx: s/withPage/pageProp/?  or pageP?

src/Frontend/Core.hs
```diff
@@ -221,52 +235,113 @@ percentVotes idea numVoters vv = {- assert c -} v
...
+authNeedCaps :: [Capability] -> Getter p CapCtx -> Applicative m => AccessInput p -> m AccessResult
+authNeedCaps needCaps' capCtx = authNeedPage $ \p ->
+    let
+        needCaps = Set.fromList needCaps'
+        haveCaps = Set.fromList $ capabilities (p ^. capCtx)
+        diffCaps = needCaps `Set.difference` haveCaps
+    in
+    if Set.null diffCaps
+        then accessGranted
+        else accessDenied ("Missing capabilities " <> cs (show diffCaps)) Nothing
```
fisx: no need to react, just saying: i'd have used `case` here.

src/Frontend/Core.hs
```diff
@@ -312,20 +387,19 @@ class Page p => FormPage p where
...
+    , _formPageRepAction :: ST
+    , _formPageRepPage   :: p
+    }
 
 instance (Show p) => Show (FormPageRep p) where
     show (FormPageRep _v _a p) = show p
 
 instance Page p => Page (FormPageRep p) where
-    isPrivatePage _ = isPrivatePage (Proxy :: Proxy p)
+    isAuthorized = isAuthorized . fmap _formPageRepPage
```
fisx: pretty!  (-:

src/Frontend/Core.hs
```diff
@@ -430,14 +501,35 @@ data Frame body
...
+                            throwError500 "AccessDeferred should not be used with LoggedIn"
+        else do
+            -- PublicFrame <$> mp <*> flushMessages
+            access <- isAuthorized (NotLoggedIn :: AccessInput p)
+            case access of
+                AccessGranted -> PublicFrame <$> mp <*> flushMessages
+                AccessDenied s u -> handleDenied s u
+                AccessDeferred -> throwError500 "AccessDeferred should not be used with NotLoggedIn"
+  where
+    handleDenied s u = throwServantErr $ (maybe Servant.err403 err303With u) { errBody = cs s }
```
fisx: i would have put this in a `let` binding for more linear reading.

src/Frontend/Fragment/Comment.hs
```diff
@@ -22,10 +22,9 @@ import qualified Frontend.Path as U
 
 
 data CommentWidget = CommentWidget
-    { _cwRenderContext :: RenderContext
-    , _cwIdeaCaps      :: [Capability]
-    , _cwComment       :: Comment
-    , _cwPhase         :: Phase
+    { _cwCapCtx   :: CapCtx
+    , _cwIdeaCaps :: [Capability]
```
fisx: is this worth a `FIXME`?  i would have thought that caps are implicit in capctx.  if this is just to avoid redundant computations, there should be a comment saying so.

src/Frontend/Page/Idea.hs
```diff
@@ -88,70 +89,125 @@ import qualified Types (MoveIdea)
...
+    , _jiTopic  :: Topic
+    }
+  deriving (Eq, Show, Read, Generic)
+
+instance SOP.Generic JudgeIdea
+
+makeLenses ''JudgeIdea
 
 instance Page JudgeIdea where
+    isAuthorized = authNeedCaps [CanMarkFeasiblity] jiCtx
```
fisx: rename cap to `CanJudge`.  (not necessarily in this PR, but since it's an easy change...)

src/Frontend/Page/Idea.hs
```diff
@@ -88,70 +89,125 @@ import qualified Types (MoveIdea)
...
-  deriving (Eq, Show, Read)
+data ReportComment = ReportComment { _rcCtx :: CapCtx, _rcComment :: Comment }
+  deriving (Eq, Show, Read, Generic)
+
+instance SOP.Generic ReportComment
+
+makeLenses ''ReportComment
 
 instance Page ReportComment where
+    -- You can report as soon as you can view the comment/idea.
```
fisx: `as soon as` => `iff`

fisx: tell me if i'm being annoying!  (-:

src/Frontend/Page/Idea.hs
```diff
@@ -368,33 +415,35 @@ instance FormPage CreateIdea where
...
-    makeForm (EditIdea idea) =
+    makeForm ei =
         ProtoIdea
         <$> ("title"         .: validateIdeaTitle (DF.text . Just $ idea ^. ideaTitle))
         <*> ("idea-text"     .:
                 validate "Idee" markdownV ((idea ^. ideaDesc) & _Markdown %%~ (DF.text . Just)))
         <*> ("idea-category" .: makeFormSelectCategory (idea ^. ideaCategory))
         <*> pure (idea ^. ideaLocation)
+      where
+        idea = ei ^. eiIdea
```
fisx: would have used a view pattern, or kept the simple pattern.

src/Frontend/Page/Login.hs
```diff
@@ -27,7 +27,10 @@ data PageHomeWithLoginPrompt = PageHomeWithLoginPrompt LoginDemoHints
   deriving (Eq, Show, Read)
 
 instance Page PageHomeWithLoginPrompt where
-    isPrivatePage _ = False
+    isAuthorized = \case
+        -- Redirect from login if the user is already logged in.
+        NotLoggedIn -> accessGranted
+        LoggedIn{}  -> accessDenied "You are already logged in" (Just U.ListSpaces)
```
fisx: nice!

src/Frontend/Core.hs
np: I would be in favor to move this into a new module ideally along with the capabilities from LifeCycle. AccessControl? Authorization?

src/Frontend/Core.hs
np: I started with your version but switched to this to reduce duplication. I propose to change the smart constructor `accessDenied` to have a `Nothing` and add `accessRedirected` for the `Just` version.

src/Frontend/Core.hs
np: Then maybe changing `pageForRole` to `rolePage`.

src/Frontend/Core.hs
np: onPage?

src/Frontend/Core.hs
np: case on a `Bool` or something with guards?

src/Frontend/Page/Overview.hs
```diff
@@ -101,6 +103,8 @@ instance ToHtml PageOverviewOfWildIdeas where
             div_ [class_ "ideas-list"] $ toHtml ideasAndNumVoters
 
 instance Page PageOverviewOfWildIdeas where
+    -- Any logged in user is authorized since findWildIdeasBySpace is already selecting the right view
+    isAuthorized = userPage
```
fisx: does that mean that we can visit enemy idea spaces, only they will appear empty?  will that be fixed by the ideaspace caps change?

src/Frontend/Page/Overview.hs
```diff
@@ -148,7 +145,9 @@ instance ToHtml PageOverviewOfTopics where
                                 span_ [class_ "theme-grid-item-link"]
                                     "view topic"
 
-instance Page PageOverviewOfTopics
+instance Page PageOverviewOfTopics where
+    -- Any logged in user is authorized since findTopicsBySpace is already selecting the right view
+    isAuthorized = userPage
```
fisx: see last comment.

src/Frontend/Fragment/Comment.hs
np: So, the `CapCtx` which was used to create the `cwIdeaCaps` was less precise than `cwCapCtx`. The one in `cwCapCtx` has a `Just comment`.

So if we were to reuse the caps coming out of `cwCapCtx` for idea-level conditions then there wouldn't be redundant computations either. However we would rely on the monotonic property of `capabilities`. Namely if a `c0 c1 : CapCtx` where `c1` is more defined than `c0` then `capabilities c0` should be a subset of `capabilities c1`.

If we agree on this analysis then we can throw away `cwIdeaCaps`.

src/Frontend/Page/Idea.hs
np: Ack.

src/Frontend/Page/User.hs
```diff
@@ -39,31 +39,43 @@ import qualified Text.Digestive.Lucid.Html5 as DF
 data PageUserSettings = PageUserSettings User
   deriving (Eq, Show, Read)
 
-instance Page PageUserSettings
+instance Page PageUserSettings where
+    -- The use settings page always goes to the profile of the current logged in user.
```
fisx: does that mean you rely on the URL to not contain the uid?  because i think that's changed, and even if not it might change in the future.

src/Frontend/Page/User.hs
```diff
@@ -39,31 +39,43 @@ import qualified Text.Digestive.Lucid.Html5 as DF
...
+    isAuthorized = userPage -- Are profiles public?
 
 -- | 8.X User profile: Editing the public profile
-data EditUserProfile = EditUserProfile RenderContext User
+data EditUserProfile = EditUserProfile CapCtx User
   deriving (Eq, Show, Read)
 
-instance Page EditUserProfile
+instance Page EditUserProfile where
+    -- Can the admin edit any profile through that endpoint?
```
fisx: yes, admin and moderator all can edit user profiles.  they are like ideas in that they may be offensive and in need of censorship.

src/Frontend/Page/User.hs
```diff
@@ -39,31 +39,43 @@ import qualified Text.Digestive.Lucid.Html5 as DF
...
 
 -- | 8.X Report user profile
 data ReportUserProfile = ReportUserProfile User
   deriving (Eq, Show, Read)
 
-instance Page ReportUserProfile
+instance Page ReportUserProfile where
+    -- If you can view the profile of a user then you can report on it.
+    -- Any user who is logged in can view the profile of any other user.
+    isAuthorized = userPage
```
fisx: missing empty line.  (there was one above that i missed, i think.)

src/Frontend/Path.hs
```diff
@@ -106,7 +106,7 @@ import Types
 
 -- * types
 
--- FIXME: Introduce AllowGet only
+-- FIXME: Introduce AllowGet only, or better make it a list of [GET|POST].
```
fisx: :+1:

fisx:
done with the review.  looks good, but i have change requests.  some can be ignored, some might be bugs.  back to you npou!

src/Frontend/Core.hs
fisx: both fine with me.

src/Frontend/Core.hs
fisx: if you have a reason for the current implementation that i missed, please don't change anything.  otherwise the names you suggest are fine.

src/Frontend/Core.hs
fisx: accessToPage?

`page-p` is lisp and prolog, but `on` in haskell i think has a different intuition.

since it's just a local name, please make a call and i'll be happy with it.  `withPage` is good enough, if none of the others seem better to you.

src/Frontend/Core.hs
fisx:
```haskell
... in case needCaps `Set.difference` haveCaps of
   [] -> ...
   diffCaps -> ...
```

src/Frontend/Fragment/Comment.hs
fisx: if you document the monotonic property at least as a comment with `FIXME: write tests for this` in LifeCycle, I agree with this change.

fisx:
s/makeFrame/runHandler/

that's my preference.  alternatives: runPage, finalizePage, wrapPage, checkAuth, checkPageAuth.  framePage.  soo many!  (-:

src/Frontend/Core.hs
np: Done.

src/Access.hs
```diff
@@ -0,0 +1,411 @@
...
+{-# LANGUAGE MultiWayIf            #-}
+{-# LANGUAGE OverloadedStrings     #-}
+{-# LANGUAGE Rank2Types            #-}
+{-# LANGUAGE ScopedTypeVariables   #-}
+{-# LANGUAGE TemplateHaskell       #-}
+{-# LANGUAGE TypeFamilies          #-}
+{-# LANGUAGE TypeOperators         #-}
+{-# LANGUAGE ViewPatterns          #-}
+
+module Access
```
andorp: That is a good name.

src/Arbitrary.hs
```diff
@@ -568,7 +572,7 @@ instance Arbitrary IdeaLocation where
 -- * user
 
 instance Arbitrary User where
-    arbitrary = garbitrary <**> (set userRole <$> garbitrary)
+    arbitrary = garbitrary
```
andorp: thx

src/Frontend/Fragment/Comment.hs
andorp: #614

src/Frontend/Page/User.hs
```diff
@@ -39,31 +39,49 @@ import qualified Text.Digestive.Lucid.Html5 as DF
...
+data EditUserProfile = EditUserProfile CapCtx User
   deriving (Eq, Show, Read)
 
-instance Page EditUserProfile
+instance Page EditUserProfile where
+    -- Can the admin edit any profile through that endpoint?
+    isAuthorized = authNeedPage $ \_ (EditUserProfile ctx u) ->
+        if isOwnProfile ctx u
+            then accessGranted
+            else accessDenied "You can only edit your own profile"
```
andorp: Translation

src/Frontend/Path.hs
andorp: [GET, POST] is isomorphic to [POST|GET].
What is the interpretation for types like: [GET], [GET, GET]

andorp:
It seems that all the Page types has `CapCtx`, probably it could go under the `Page` typeclass in future PRs.

andorp:
LTGM

src/Frontend/Page/Idea.hs
fisx: done.

src/Frontend/Path.hs
fisx: interesting questions.  but i think you can implement set membership tests in typed lists.  details to be discussed once we get around to playing with this...

fisx:
LGTM, too!  I'll merge when green.

src/Frontend/Page/User.hs
np: It hasn't changed yet but with 58d2aa51cf5b5c36af82bf949130af2ab3a4deaa we no longer have this assumption.

src/Frontend/Path.hs
np: Like in Servant with the content types `Get '[JSON,JSON,HTML]`

PR #615, opened 2016-06-03
Delegation

andorp:


src/Types.hs
```diff
@@ -285,12 +285,21 @@ type instance Proto IdeaLike = ()
 data IdeaVote = IdeaVote
     { _ideaVoteMeta  :: MetaInfo IdeaVote
     , _ideaVoteValue :: IdeaVoteValue
+    , _ideaVoteVoter :: AUID User
```
fisx: rename to `_ideaVoteDelegate`?  also document which of delegate and delegator is stored in meta info.

fisx: also comment on the fact that the delegation path can be lost over time.  (i don't see a problem with it now, but it should be made explicit.)

src/Types.hs
```diff
@@ -740,6 +749,15 @@ data DelegationContext =
 
 instance SOP.Generic DelegationContext
 
+data DelegationContextFull =
```
fisx: rename `DelegationContext` to `DScope` in all its variants?

tests/DelegationSpec.hs
```diff
@@ -0,0 +1,170 @@
...
+        spaces = DlgCtxIdeaSpace <$> unIdeaSpaces u
+        topics = DlgCtxTopicId   <$> unTopics     u
+        ideas  = DlgCtxIdeaId    <$> unIdeas      u
+
+-- * delegation program
+
+data DelegationDSL where
+    SetDelegation       :: AUID User -> DelegationContext -> AUID User      -> DelegationDSL
+    Vote                :: AUID User -> AUID Idea         -> IdeaVoteValue  -> DelegationDSL
+    CheckNoOfDelegatees :: AUID User -> DelegationContext -> Int -> DelegationDSL
```
fisx: rename to `CheckNumVotingPower`.  it's number of delegatees (aka delegators) plus oneself.

andorp:
Closing in favour of #629

PR #625, opened 2016-06-07
Encrypt the passwords with scrypt

np:


src/Arbitrary.hs
```diff
@@ -638,8 +639,9 @@ instance Arbitrary InitialPassword where
     shrink = gshrink
 
 instance Arbitrary EncryptedPassword where
-    arbitrary = garbitrary
-    shrink    = gshrink
+    arbitrary = mk <$> arb <*> arb
+      where
+        mk salt pass = ScryptEncryptedPassword . getEncryptedPass $ encryptPass' (Salt salt) (Pass pass)
```
fisx: i wonder why you don't want the other two cases, but i don't feel strongly either way.

src/Types.hs
```diff
@@ -29,6 +29,7 @@ where
 import Control.Lens
 import Control.Monad
 import Control.Monad.Trans.Except (ExceptT, runExceptT)
+import Crypto.Scrypt
```
fisx: (we should really break up this module...)

fisx:
LGTM & works.  will merge when green.  thanks!

src/Arbitrary.hs
np: Which other two?

src/Frontend/Page/User.hs
```diff
@@ -115,7 +115,8 @@ instance FormPage PageUserSettings where
             <*> ("old-password"  .:
                     -- while we need to check that the old password is the correct
                     -- one, we do not need to validate it against the rules for new passwords.
-                    DF.checkM "Das alte Passwort ist nicht korrekt" checkUserPassword
+                    -- TODO: why can't the user provide a Nothing here to circumvent the check?
```
np: The user can provide nothing but `checkPwdAllOrNothing` is checking for that.

src/Arbitrary.hs
fisx: my mistake.  this is the type for encrypted passwords only, and it is contained in the `UserPass`, where there are other constructors.  i got the two confused.

src/Frontend/Page/User.hs
fisx: ok makes sense.  i'll add a comment about that.  (unless you want to do it...?)

np: To me the comment line 98 was enough.

PR #627, opened 2016-06-07
Refactor IdeaView page.

### LINKED ISSUE #624: Streamline & refactor detail headers
The buttons, progress bar and indicator areas are getting very messy in detail header. I propose something like this
```html
</main-heading>
</sub-heading>
</progress-bar>
<icon-list>
    </icon-icon1></icon-icon2>
</icon-list>
<admin-buttons>
    </btn-cta></btn-cta>
</admin-buttons>
```
And we just keep it like that for every type of detail heading turning off buttons/indicators where appropriate

fisx:
Fixes #624

- lose anchor in feasibilityVerdict.  i don't think it was used for anything.
- separate vote bars, buttons.
- re-arrange buttons and indicators into resp. exclusive divs.


fisx:
this is my first shot at #624.  @localgrr what do you think?

localgrr:
I'm happy with this @fisx

fisx:
LGTM.  since i've written a bit of non-trivail code: anybody care to take a look and merge?

src/Frontend/Page/Idea.hs
```diff
@@ -287,33 +287,44 @@ instance ToHtml ViewIdea where
...
+            -- explanation by the dean why the idea is feasible or not (if available)
+            feasibilityVerdict idea
 
-            -- creator statement
-            when (any (`elem` caps) [CanAddCreatorStatement, CanEditCreatorStatement]) $ do
-                div_ [class_ "creator-statement-button"] $ do
+            -- buttons
+            toHtml $ ideaVoteLikeButtons ctx stats
+
+            when (has _PhaseWildIdea phase && ideaReachedQuorum stats && canCreateTopic) $ do
```
andorp: The `ideaReachedQuorum stats`  is redundant, it is already in the definition if canCreateTopic.

andorp: The `CanCreateTopic` should be in `topicCapabilites`.

src/Frontend/Page/Idea.hs
```diff
@@ -287,33 +287,44 @@ instance ToHtml ViewIdea where
...
+                div_ [class_ "table-actions m-no-hover"] $ do
                     button_ [ class_ "btn-cta m-valid"
-                            , onclick_ $ U.creatorStatement idea
+                            , onclick_ $ U.Space spc U.CreateTopic
```
andorp: I would prefer the usage of `U.createTopic` .

andorp:
LGTM

src/Frontend/Page/Idea.hs
fisx: you're right about the redundancy in the condition.  this kind of mistakes happens if you introduce too many names whose definitions are *almost* trivial, and no longer than the names themselves.  i've removed some of these names in this case.

src/Frontend/Page/Idea.hs
fisx: that doesn't exist?

fisx:
ah, you did it yourself.  never mind.

PR #628, opened 2016-06-07
2016 06 cc more bugfixes

localgrr:
Variety of small mostly mobile bugfixes


fisx:
1. what's the `if admin` in the frame about?  we can do it, but i would like to understand.
1. could you take a look at 8b4bf1e?  sorry for squeezing this in here, i thought it was a smaller deal until it was too late.  i hope it's ok that i collapsed some `div_` levels that seemed unnecessary to me.

fisx:
off-line communication re. `if admin`: the admin pages are not responsive, and fixing that would take a lot of work, so we will fix this by not giving admins a responsive ui.  they will have to work with the desktop ui on all devices.

fisx:
@localgrr, admins *are* still able to use aula on their phones, right?  it's just not quite as nice?

fisx:
@localgrr could you test admin-on-mobile and the div-nesting changes and merge?  I'm happy with everything now.

PR #634, opened 2016-06-08
Use CanDelegate instead of CanVote to display delegation buttons on the profile page (fixes #630)

### LINKED ISSUE #630: delegation on user profiles does not work.
i think the reason is that we are using the `CanVote` capability, which is only avaiable in certain phases, which are not set in user profiles, only in topics and ideas.  i suggest we add a new capability `CanDelegate`, which is technically independent from `CanVote`, and thus from phases.

np:


andorp:
LGTM

PR #635, opened 2016-06-08
Remove ideaSpaceToSchoolClass and use lenses instead

np:


src/DemoData.hs
```diff
@@ -202,7 +202,7 @@ universe rnd size = do
 
     ideaSpaces <- nub <$> generate (numberOfIdeaSpaces size) rnd arbitrary
     mapM_ (update . AddIdeaSpaceIfNotExists) ideaSpaces
-    let classes = mapMaybe ideaSpaceToSchoolClass ideaSpaces
+    let classes = ideaSpaces ^.. each . _ClassSpace
```
andorp: https://hackage.haskell.org/package/lens-4.14/docs/src/Control.Lens.Fold.html#%5E..

andorp:
LGTM

fisx:
LGTM, too.  @np, could you make travis happy and merge?

I would also like to have the link into the lens docs in the code, but I suspect you disagree there.  (-:

np:
Sorry for wishful coding... @andorp the naming convention was to call it Erratatata though :)

PR #637, opened 2016-06-08
Render delegatees on user profile page.

andorp:


andorp:
Discovered design issue during implementaiton: #638

src/Frontend.hs
```diff
@@ -325,7 +325,8 @@ aulaSpace space
 
 type AulaUser =
        "ideas"       :> GetH (Frame PageUserProfileCreatedIdeas)
-  :<|> "delegations" :> GetH (Frame PageUserProfileDelegatedVotes)
+  :<|> "delegations" :> "global" :> GetH (Frame PageUserProfileDelegatedVotes)
+  :<|> "delegations" :> "class"  :> GetH (Frame PageUserProfileDelegatedVotes)
```
fisx: it would be nice to have the class here explicitly, as it won't be uniquely identified by the uid any more soon.  but since this implies changes to the UI, i'm ok with postponing the change here, too.

src/Frontend/Page/User.hs
```diff
@@ -252,40 +259,61 @@ instance ToHtml PageUserProfileDelegatedVotes where
...
-    <*> pure [] -- FIXME
+delegatedVotesGlobal userId = delegatedVotes userId DScopeGlobal
+
+delegatedVotesClass :: (ActionPersist m, ActionUserHandler m)
+      => AUID User -> m PageUserProfileDelegatedVotes
+delegatedVotesClass userId = do
+    user <- mquery (findUser userId)
+    case user ^? userRole . _Student of
+        -- TODO: Translation
+        Nothing -> throwError500 "User is not a student"
```
fisx: "Nutzer ist kein Schüler".  but i don't think we need to translate internal errors, or errors that can only be triggered by programming mistakes or malicious behaviour around the frontend.

src/Frontend/Page/User.hs
```diff
@@ -252,40 +259,61 @@ instance ToHtml PageUserProfileDelegatedVotes where
...
+        Just cl -> delegatedVotes userId (DScopeIdeaSpace (ClassSpace cl))
+
+delegatedVotes :: (ActionPersist m, ActionUserHandler m)
+      => AUID User -> DScope -> m PageUserProfileDelegatedVotes
```
fisx: i think this already works fine with any `DScope`, even topic and idea.  if so: nice!  (-:

src/Frontend/Page/User.hs
```diff
@@ -54,8 +58,11 @@ data PageUserProfileCreatedIdeas = PageUserProfileCreatedIdeas CapCtx UserView L
 instance Page PageUserProfileCreatedIdeas where
     isAuthorized = userPage -- Are profiles public?
 
+type DelegationInfo = [(User, [User])]
```
fisx: newtype?

fisx:
looks good so far.  made a few comments, could you take another look?  thanks!

src/Frontend/Page/User.hs
andorp: I think that everything can reach the user should be translated, even if we hope it never happen.

np: I'm on @fisx side here, internal errors do not have to treated as part of the UI, hence no need to translate them.

PR #639, opened 2016-06-08
User profile only shows the ideas visible to the current user (fixes #631)

### LINKED ISSUE #631: filter ideas in user profile by availability for reading by current user.
status now: if you are a student and visit the profile of a student of another class, you can see the ideas living in that class, but when you click on them, you get "access denied".  the ideas should just be invisible.

np:


fisx:
LGTM, thanks.

this will become more interesting with students who are also moderators and should be able to see beyond their own class(es), but not by much i think.

PR #640, opened 2016-06-08
Adding support for multiple roles, part 1

np:


np:
This is ready for review but not for merging. We need to decide how to address the remaining TODOs, some are related to delegation, some are UI related, some are test related.

andorp:
LGTM so far.

src/Types.hs
```diff
@@ -1299,8 +1313,26 @@ unsafeEmailAddress local domain = InternalEmailAddress $ Email.unsafeEmailAddres
...
+hasRole user role = role `Set.member` (user ^. userRoleSet)
+
+isAdmin :: User -> Bool
+isAdmin = (`hasRole` Admin)
+
+roleScope :: Getter Role RoleScope
+roleScope = to $ \r ->
+    case r ^? roleSchoolClass of
+        Nothing -> SchoolScope
+        Just cl -> ClassesScope $ Set.fromList [cl]
```
fisx: There is `Set.singleton`, but i don't have a preference here.

src/Types.hs
```diff
@@ -1365,6 +1397,7 @@ showIdeaSpace :: IdeaSpace -> String
 showIdeaSpace SchoolSpace    = "school"
 showIdeaSpace (ClassSpace c) = showSchoolClass c
 
+-- FIXME HasUILabel: there is already an instance for SchoolClass.
```
fisx: `HasUILabel` is about human-readable, this is about machine-readable, but the distinction is not very clear and it's possible we got it wrong in a place or two.  I"ll look into this.

src/Access.hs
```diff
@@ -415,8 +415,8 @@ isSameSchoolClass ctx = isJust . isSameSchoolClass' (ctx ^. capCtxUser)
 isSameSchoolClass' :: User -> User -> Maybe SchoolClass
 isSameSchoolClass' user user' = if c == c' then c else Nothing
```
fisx: this can be changed to the obvious `sharesSomeSchoolClasses`, but IRC at least one of the use cases is to filter ideas in the user profile for those the visitor has access two.  for that we'll need `isInAccessibleSchoolClass :: Idea -> User -> Maybe SchoolClass`.

src/Frontend/Core.hs
```diff
@@ -487,9 +487,10 @@ pageFrame frame = do
         link_ [rel_ "stylesheet", href_ $ P.TopStatic "css/all.css"]
 
         -- | disable the meta tag for admins, since admin pages are not working on mobile devices.
-        case frame ^? frameUser . userRole of
-            Just Admin -> meta_ [name_ "viewport", content_ "width=1024"]
-            _          -> meta_ [name_ "viewport", content_ "width=device-width, initial-scale=1"]
+        let viewport_content
+                | allOf frameUser isAdmin frame = "width=1024"
```
fisx: should this be `anyOf`?   an admin who is also moderator can see the admin pages, so she shouldn't be switched to mobile.  see comment above.

fisx:
not done reading, but very happy so far.

there is at least one more place where this is going to require re-design on UI level, and that's the user profile under `delegations`, in the DScope navigation buttons.  the only places you marked with TODOs are those in the admin pages where user roles are manipulated.  i'm not worried about either of them, but certain are we that we haven't missed anything?

fisx:
sorry, didn't have time to fix the last ghc error...

src/Types.hs
```diff
@@ -676,6 +677,19 @@ data Role =
...
 instance SOP.Generic Role
 
+data RoleScope
+  = ClassesScope (Set SchoolClass)
+  | SchoolScope
+  deriving (Eq, Ord, Show, Read, Generic)
+
+instance SOP.Generic RoleScope
+
+instance Monoid RoleScope where
```
fisx: you should make a comment here and tell the story you told in the call this morning.  it's very insightful.

src/Types.hs
np: Ack.

src/Access.hs
np: We now have

```
commonSchoolClasses :: User -> User -> Set SchoolClass
haveCommonSchoolClass :: CapCtx -> User -> Bool
```

src/Types.hs
np: Done.

fisx:
LGTM, will merge when green.  thanks!

There are a few `TODO` still in there, but they all relate to delegation and we should do them separately.

Issue #556, opened 2016-05-24
First login procedure for students


fisx:
Users on login should:

~~1. if password is initial, be asked to change it by being redirected to the user settings with a status message "please change your password".  [solved]~~
2. if logged in for the first time (determined by a dedicated flag), chose a delegate.  this is an option and it should be possible to turn it off in `Config` or in the admin pages.  This should not apply to users that have no student role.

fisx:
take this out of it.2?  we're done with password-change part, only thing missing is the delegate-your-vote part.

fisx:
updated description.  agreed with @rittero to keep this in the maybe pile.

fisx:
not a requirement any more, for now.

PR #641, opened 2016-06-08
Force user to change initial password.

andorp:
After login it redirect to the user settings page form the spaces and the delegations pages.

### DIFF
--- a/src/Frontend.hs
+++ b/src/Frontend.hs
@@ -172,7 +172,7 @@ type AulaMain =
 
 aulaMain :: ActionM m => ServerT AulaMain m
 aulaMain =
-       runHandler Page.viewRooms
+       forceChangeInitialPassword (runHandler Page.viewRooms)
   :<|> aulaSpace
 
   :<|> aulaUser
@@ -180,7 +180,7 @@ aulaMain =
   :<|> aulaAdmin
 
   :<|> error "api not implemented: \"delegation\" :> \"edit\" :> FormHandler ()"
-  :<|> runHandler Page.viewDelegationNetwork
+  :<|> forceChangeInitialPassword (runHandler Page.viewDelegationNetwork)
 
   :<|> runHandler (pure PageStaticImprint)
   :<|> runHandler (pure PageStaticTermsOfUse)
--- a/src/Frontend.hs
+++ b/src/Frontend.hs
@@ -41,7 +41,6 @@ import Thentos.Frontend.State (serveFAction)
 import Action (ActionM, UserState, ActionEnv(..), logout, phaseTimeout)
 import Action.Implementation (Action, mkRunAction)
 import Config
-import Data.UriPath
 import Daemon
 import Logger.EventLog
 import Frontend.Core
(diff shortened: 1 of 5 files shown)


src/Frontend.hs
```diff
@@ -172,15 +172,15 @@ type AulaMain =
 
 aulaMain :: ActionM m => ServerT AulaMain m
 aulaMain =
-       runHandler Page.viewRooms
+       forceChangeInitialPassword (runHandler Page.viewRooms)
```
fisx: why these two?  i'm particularly confused by the one wrapping the delegation network.  is that needed?

shouldn't we redirect to a dedicated end-point after login, and that end-point redirects to the actual landing page, whichever that is?  that way if we re-visit that page the check won't happen again, only after the next login.

src/Frontend/Core.hs
```diff
@@ -431,6 +432,15 @@ runHandler mp = do
     handleDenied s u = throwServantErr $ (maybe Servant.err403 err303With u) { errBody = cs s }
         -- FIXME log these events as INFO, should we do this here or more globally for servant errors.
 
+-- ASSUMPTION: The user is logged in.
+forceChangeInitialPassword :: ActionM m => m a -> m a
+forceChangeInitialPassword handler = do
+    x <- handler
+    user <- currentUser
+    if has (userSettings . userSettingsPassword . _UserPassInitial) user
+        then do addMessage "Please change your password." -- TODO: Translation
```
fisx: "Bitte ändere dein Passwort, damit niemand dein Login klauen kann."

fisx:
this is related to #556, but it does not offer the delegation page yet, so we should leave that open.

i had a few comments, otherwise LGTM.  @andorp if you have time before monday you can fix it, otherwise first come first serve i guess.  thanks!

src/Frontend.hs
andorp: My first idea was that. I wanted to code that into the `runHandler`. What would happen with that approach:

 1. Render login page
 1. User logs in
 1. runHandler on list ideas checks if there is an initial password, it redirects to the user setting page
 1. the server receives a GET request for user settings page
 1. runHandler on the user profile page checks if there is initial password, it redirects to the user settings page
 1. and so on

andorp: We could save as a session state if we rendered redirected once in the session on case of initial password.

andorp: Guarded list idea spaces and not guarded delegation network pages made me feel wierd as I tested out manually:

 - I can't open idea lists page but I can see my delegation? Hmm that's wierd.

I aimed the consistency and force the user to make change his initial password.

fisx: note that "profile page" and "settings page" are two different things.

but i still don't get it.  i guess i'll try what i have in mind and i'll see what you mean in 10 minutes (-:

fisx:
@andorp what do you think?

andorp:
I like it.

andorp:
LGTM

PR #645, opened 2016-06-09
Type safety for paths

andorp:


fisx:
I merged master, and found a few things in path that I would like to see cleaned up.  Mostly it's documentation of function names that are not self-explanatory, and I think we should pay more attention to the order in which we define and then export things.

@andorp could you take another look at `git grep TODO src/Frontend/Path.hs`?

andorp:
@fisx: Please review, I did a rebase.

Issue #578, opened 2016-05-26
refresh after click is wobbly.


fisx:
sometimes, the reload fetches the old page before the server can update.  is there a way to wait until the post request triggered by the click returns, and only trigger the reload then?

fisx:
I suspect `runPostHandler` in #648 triggered the death of the already fragile way of doing server communication we've come up with in `static/js/custom.js:reloadOnClick` and `src/Lucid/Missing.hs:postButton_`.  I don't understand why, but I think this is a good time to find a better fix.

I propose to change `reloadOnClick` as follows:

1. inspect `event.currentTarget.parentElem` to get to the contents of the form.
1. construct an ajax request and and it to the server asynchronously.
1. in doing so, register a success event handler that triggers the reload.
1. `event.preventDefault(); event.stopImmediatePropagation(); event.stopPropagation();` as necessary.

this will get rid of the race condition and avoid unnecessary waiting.  and it should solve whatever is going wrong in the tip of #648.

PR #648, opened 2016-06-10
Authorization Part 2

np:


np:
Only the last commit (b8272517b1aae1bdee09c3b44d3b8dd4a0ba9be3) is to be reviewed.

np:
I mean it is not finished at all but I'm open to remarks.

fisx:
so the test suite error is "aula-eventlog.json no such file".  `aulaServerWithEventLog` did set up the event log and remove the file, and it appears that somewhere in this PR the server starts to write to the event log less, so the file doesn't get recreated.

hypothesis: the initial test data is still added to the server, but something changed in the writing of the event protocol.  either the events aren't logged any more, or something goes wrong while that happens and we don't detect it.

i may be still far off, though.  need to read up on what I did so many days ago...

src/Access.hs
```diff
@@ -402,6 +412,28 @@ authNeedCaps needCaps' getCapCtx = authNeedPage $ \_ p ->
...
+authNeedCaps :: [Capability] -> Getter p CapCtx -> AccessCheck p
+authNeedCaps needCaps getCapCtx = authNeedPage $ \_ p -> authNeedCaps' needCaps (p ^. getCapCtx)
+
+makeLenses ''NeedCap
+makePrisms ''NeedCap
+
+makeLenses ''DelegateTo
+makePrisms ''DelegateTo
+
+-- FIXME: dependent types needed here
```
fisx: This is not a FIXME for aula, but for ghc...

fisx: (it's great to have dependent type signatures in comments, though.)

src/Frontend/Core.hs
```diff
@@ -125,6 +129,8 @@ type instance CaptureData User               = AUID User
 type instance CaptureData IdeaJuryResultType = IdeaJuryResultType
 type instance CaptureData Role               = Role
 
+data PostResult a = UnsafePostResult
```
fisx: why is this unsafe?

src/Frontend/Core.hs
```diff
@@ -401,33 +464,36 @@ form formHandler = getH :<|> postH
 -- | Wrap anything that has 'ToHtml' and wrap it in an HTML body with complete page.
 data Frame body
     = Frame { _frameUser :: User, _frameBody :: body, _frameMessages :: [StatusMessage] }
-    | PublicFrame               { _frameBody :: body, _frameMessages :: [StatusMessage] }
+    | PublicFrame               { _frameBody :: body, _frameMessages :: [StatusMessage] } -- TODO remove messages
```
fisx: what does this TODO mean?

src/Frontend/Core.hs
np: By unsafe I mean that there should not be more occurrences of it. The goal of this type is to make sure one call `runPostHandler` .

src/Frontend/Core.hs
np: I'm not really sure it makes sense to display messages to not logged in users. If the user is not logged in then the user state is empty therefor the list of messages is empty therefore we don't need to have it here.

src/Frontend/Core.hs
fisx: ah, i see.  like this?: https://github.com/liqd/aula/pull/648/commits/e95c58c0bbc8222c6974e9f11241164664dfd6f5

src/Frontend/Core.hs
fisx: I think it does make sense:

https://github.com/liqd/aula/pull/648/commits/5a953e2e65ca6fcb11ee497e5e968f43e70b3529

any objections?

src/Frontend/Core.hs
```diff
@@ -240,6 +249,42 @@ instance Page p => Page (Frame p) where
...
+instance (KnownSymbol s, Page a) => Page (QueryParam s b :> a) where
+    isAuthorized = error "IMPOSSIBLE"
+
+instance Page a => Page (FormReqBody :> a) where
+    isAuthorized = error "IMPOSSIBLE"
+
+instance Page a => Page (Get c a) where
+    isAuthorized = error "IMPOSSIBLE"
+
+instance Page a => Page (Post c a) where -- TODO
```
fisx: is there a reason why `Get` above doesn't get a `TODO`, but `Post` and `PostResult` do?

fisx:
i think we are getting closer to something that we can start cleaning up.  (-:

fisx:
re. e3ba4f9: it would be nice to have a debug log entry for every call to runHandler' that contains the type rep of the page.  that would justify keeping the Typeable constraint around everywhere.

src/Frontend/Core.hs
```diff
@@ -154,6 +154,22 @@ type GetH = Get '[HTML, PlainText]
 type PostH p = Post '[HTML] (PostResult p)
 type FormHandler p = FormH '[HTML, PlainText] (Frame (FormPageRep p)) (FormPageResult p)
 
+-- | A void type for end-points that respond with 303 and thus never return any values.
+data Redirect
```
fisx: We could introduce a smart constructor `redirect :: URL -> Redirect` that throws an exception.  (thanks @np)

src/Access.hs
```diff
@@ -369,6 +369,11 @@ data NeedCap (cap :: Capability) = NeedCap { _needCapCtx :: CapCtx }
 data NeedAdmin = NeedAdmin  -- TODO: should constructors 'NeedAdmin', 'NeedCap', ... be renamed 'unsafe*'?
   deriving (Generic)
 
+-- | TODO: should this go elsewhere?  is it a function we want to have around?  is it implemented right?
+adminIsAuthorized :: Applicative m => AccessInput p -> m AccessResult
+adminIsAuthorized (LoggedIn u _) | u `hasRole` Admin = accessGranted
+adminIsAuthorized _                                  = accessDenied Nothing
+
```
np: use `adminPage` instead.

src/Access.hs
```diff
@@ -429,8 +431,8 @@ makePrisms ''NeedCap
 makeLenses ''DelegateTo
 makePrisms ''DelegateTo
 
--- FIXME: dependent types needed here
-needCap :: {-cap :: -}Capability -> AccessCheck (NeedCap cap)
+-- (the type signature would be more interesting with dependent types.)
```
np: There is a way to fake this kind of dependent type in Haskell but this might not be worth the effort.
The issue here being that: `needCap CanEdit :: AccessCheck (NeedCan CanVote)`

src/Frontend/Core.hs
```diff
@@ -467,10 +468,11 @@ data Frame body
     | PublicFrame               { _frameBody :: body, _frameMessages :: [StatusMessage] } -- TODO remove messages
   deriving (Show, Read, Functor)
 
--- The first function 'mp' computes the page from the current logged in user.
--- WARNING, 'mp' CAN be run before authorization, this 'mp' SHOULD NOT modify the state.
--- The second function 'mr' computes the result and performs the necessary action on the
--- state.
+-- | Check authorization of a page action and wrap the page in a 'Frame'.  Returns a new action.
```
np: This is actually not putting the Frame, runHandler does that.

src/Frontend/Core.hs
```diff
@@ -253,31 +253,34 @@ instance Page p => Page (Frame p) where
...
-    isAuthorized = error "IMPOSSIBLE"
+    isAuthorized = error "IMPOSSIBLE: instance Page a => Page (Get c a)"
 
-instance Page a => Page (Post c a) where -- TODO
-    isAuthorized = error "IMPOSSIBLE"
+instance Page a => Page (Post c a) where -- TODO: we should probably not have this instance, then we
+                                         -- won't be able to accidentally sneak in an unprotected
+                                         -- 'Post' handler.  ame with 'Get' above and 'PostResult'
+                                         -- below.
+    isAuthorized = error "IMPOSSIBLE: instance Page a => Page (Post c a)"
```
np: I propose to make a single instance for `Page c (PostResult a)` instead of two. We get the guarantee that `Post` cannot be used as-is.

fisx: sounds good!

src/Access.hs
np: Also `adminPage` is redirecting to the login page if the user is not logged in.

fisx: that's at least as good as what i did here.  i just didn't think about it.  (-:

src/Frontend/Core.hs
```diff
@@ -150,9 +167,20 @@ instance Aeson.ToJSON (PostResult a) where toJSON = Aeson.gtoJson
...
 -- for more details).
-type GetH = Get '[HTML, PlainText]
-type PostH p = Post '[HTML] (PostResult p)
-type FormHandler p = FormH '[HTML, PlainText] (Frame (FormPageRep p)) (FormPageResult p)
+type GetH p = Get '[HTML, PlainText] (GetResult p)
+type PostH' p r = Post '[HTML, PlainText] (PostResult p r)
+type PostH p = PostH' p ()
+type FormHandler p =
+       GetH (Frame (FormPageRep p))
+  :<|> FormReqBody :> PostH' (Frame (FormPageRep p)) (Frame (FormPageRep p)) -- Redirect
```
fisx: why can't we use type `Redirect` type here, then?  (just trying to understand.)

src/Frontend/Core.hs
```diff
@@ -282,35 +310,36 @@ instance (KnownSymbol s, Page a) => Page (Capture s c :> a) where
...
     isAuthorized = error "IMPOSSIBLE: instance Page a => Page (FormReqBody :> a)"
 
-instance Page a => Page (Get c a) where
-    isAuthorized = error "IMPOSSIBLE: instance Page a => Page (Get c a)"
+-- We must not generalize this instance to a `Page (Get c a)` instance.
+instance Page a => Page (Get c (GetResult a)) where
+    isAuthorized = error "IMPOSSIBLE: instance Page a => Page (Get c (GetResult a))"
+
+-- We must not generalize this instance to a `Page (Get c a)` instance.
+instance Page a => Page (Get c (Frame a)) where
```
fisx: does this mean that `Frame` is part of the security mechanism?  i think i'm asking weird questions.  bear with me.  (-:

src/Frontend/Core.hs
```diff
@@ -499,6 +501,9 @@ form formHandler = getH :<|> postH
         v <- getForm fa (processor1 page)
         pure $ FormPageRep v fa page
 
+    -- If form is not filled out successfully, 'postH' responds with a new page to render.  This
+    -- makes it impossible to re-use 'runPostHandler': we need the user info and the page computed
+    -- pre-authorization and forwarded by 'coreRunHandler'.
```
fisx: i'm undecided if this text clarifies or confuses things?

src/Backend.hs
```diff
@@ -44,20 +44,32 @@ type DelegationsApi = GetJ DelegationNetwork
...
+instance Page NeedEmptyUserMap where
+    isAuthorized = publicPage
+{-
+    FIXME: since we don't have access to the db here.
+    isAuthorized _ = do
+        noUsers <- query $ views dbUserMap Map.null
+        if noUsers
+            then accessGranted
+            else accessDenied $ Just "Can only be used when there are no users!"
+-}
```
fisx: In thentos, we have a credential *is coming from localhost*.  I like this better because it doesn't require any additional work to avoid having to race attackers after deployment.  For now I'd like to leave this a FIXME, but I'll add this thought to it.

Makefile
```diff
@@ -71,9 +72,11 @@ wc:
...
 
 content:
-	curl -XPOST http://localhost:8080/api/manage-state/create-init
-	curl -XPOST http://localhost:8080/api/manage-state/create-demo
-	curl -XPOST http://localhost:8080/api/manage-state/create-votes
+	curl -XPOST $(AULA_URL)/api/manage-state/create-init
+	curl -c cookie-jar -F /login.user=admin -F /login.pass=pssst $(AULA_URL)/login
+	curl -b cookie-jar -XPOST $(AULA_URL)/api/manage-state/create-demo
+	curl -b cookie-jar -XPOST $(AULA_URL)/api/manage-state/create-votes
+	rm cookie-jar
```
fisx: awesome!  i've wanted to figure out how to do this for a long time.  thank you!  (-:

fisx:
manual testing:

```
$ make content
curl -XPOST http://localhost:8080/api/manage-state/create-init
[]curl -c cookie-jar -F /login.user=admin -F /login.pass=pssst http://localhost:8080/login
curl -b cookie-jar -XPOST http://localhost:8080/api/manage-state/create-demo
[]curl -b cookie-jar -XPOST http://localhost:8080/api/manage-state/create-votes
Rolle Administrator benötigt.rm cookie-jar
```

fisx:
when deleting an idea as moderator, we end on a blank page.  looks like a `toHmtl ()`.  perhaps something wrong with the redirect?  same symptom with comment votes.

fisx:
i put a `sleep 1` into the Makefile rule and can't see the message any more.

src/Frontend/Core.hs
np: Redirect looks like a good fit for the second parameter of `PostH'` (which could replace `PostH` if we add `()` at a few places). However this does not work because of validation errors, in case of errors we do not redirect but display `Frame (FormPageRep p)`.

fisx:
can't reproduce it without the `sleep 1` any more now either.  strange!

fisx:
the issue with (probably) `runPostHandler` is indeed only reproducible on this branch, but not on master.  i'll investigate.

src/Frontend/Core.hs
np: This might be worth a second try what breaks if you remove it?

fisx: good idea.  nobody misses anything.  (-:

fisx:
outsourced the problem to #578.

so...  LGTM!

andorp:
LGTM too

PR #642, opened 2016-06-09
Email based reset password page

andorp:
~~Fixes #605.  Fixes #574.~~

obsoleted by #723

PR #657, opened 2016-06-14
Redirect to spaces after changing initial password.

andorp:


fisx:
LGTM.

caveats:

1. we will need to touch this again in #642 
1. if we move away from this page without changing the initial password and come back later, we get redirected to the idea space overview on change, even though we have visited the page deliberately, not for finalizing login.

both minor points though.  merging.  thanks!

PR #663, opened 2016-06-15
Allow pages to have extra footer elems rather than page headers.

fisx:
i hear this is how js people load code these days, and we're only using it for that.


andorp:
LGTM

PR #664, opened 2016-06-15
Display Delegation information on DelegationNetwork page.

andorp:


src/Frontend/Page/User.hs
```diff
@@ -60,12 +60,13 @@ data PageUserProfileCreatedIdeas = PageUserProfileCreatedIdeas CapCtx UserView L
 instance Page PageUserProfileCreatedIdeas where
     isAuthorized = userPage -- Are profiles public?
 
-newtype DelegationInfo = DelegationInfo [(User, [User])]
+-- Represents two step long delegation paths from a user.
+newtype DelegationPath = DelegationPath [(User, [User])]
```
fisx: `DelegationTree`

src/Frontend/Page/Delegation.hs
```diff
@@ -80,21 +83,19 @@ topicDelegation tid = formPageHandlerWithMsg
 
 -- | 13. Delegation network
 -- FIXME: Render all the delegations not just the current user related ones.
-data PageDelegationNetwork = PageDelegationNetwork [Delegation]
+data PageDelegationNetwork = PageDelegationNetwork DScope [DScope] [DelegationInfo]
```
fisx: also interesting here: which DScopes are parents of whom?

fisx: (not urgent, could be done in this PR or in a separate one.)

src/Frontend/Page/Delegation.hs
```diff
@@ -152,7 +153,43 @@ instance ToHtml PageDelegationNetwork where
...
+    type FormPageResult  PageDelegationNetwork = PageDelegationNetworkPayload
+
+    formAction (PageDelegationNetwork scope _ _)      = U.delegationViewScope scope
+    redirectOf _ (PageDelegationNetworkPayload scope) = U.delegationViewScope scope
+
+    makeForm (PageDelegationNetwork actualDScope dscopes _delegations) =
+        PageDelegationNetworkPayload
+        <$> ("scope" .: DF.choice delegationScopeList (Just actualDScope))
+      where
+        -- TODO: Better visual instead of show
```
fisx: yes, good idea.  if there is no HasUILabel instance, just write one.

src/Types.hs
```diff
@@ -804,6 +812,24 @@ data DScope =
 
 instance SOP.Generic DScope
 
+readEitherCS :: (ConvertibleStrings String c, Read a) => String -> Either c a
+readEitherCS = either (Left . cs) Right . readEither
```
fisx: move to mini-prelude in Types.hs?

src/Frontend/Page/Delegation.hs
```diff
@@ -152,7 +153,43 @@ instance ToHtml PageDelegationNetwork where
...
+        form $ do
+            inputSelect_ [] "scope" v
+            DF.inputSubmit "Show delegations!"
+            p_ . toHtml $ show delegations
+            img_ [src_ . U.TopStatic $ "images" </> "delegation_network_dummy.jpg"]
+
+viewDelegationNetwork :: ActionM m => Maybe DScope -> FormPageHandler m PageDelegationNetwork
+viewDelegationNetwork mscope = formPageHandler
+    (do user <- currentUser
+        let scope = fromMaybe DScopeGlobal mscope
```
fisx: this is a perfect use case for a view pattern, as `mscope` is not needed.

fisx: (just saying :-)

fisx:
LGTM!  for details see comments.

exec/RenderHtml.hs
```diff
@@ -91,6 +90,7 @@ pagesForm f =
     , f (Proxy :: Proxy PageUserSettings)
     , f (Proxy :: Proxy PageAdminSettingsEventsProtocol)
     , f (Proxy :: Proxy PageDelegateVote)
+--    , f (Proxy :: Proxy PageDelegationNetwork) -- FIXME
```
fisx: i commented that back in and it works.  why did you take it out?

src/Frontend.hs
```diff
@@ -159,7 +159,7 @@ type AulaMain =
 
        -- delegation network
   :<|> "delegation" :> "edit" :> FormHandler PageDelegateVote
-  :<|> "delegation" :> "view" :> GetH (Frame PageDelegationNetwork)
+  :<|> "delegation" :> "view" :> QueryParam "score" DScope :> FormHandler PageDelegationNetwork
```
fisx: you still need to fix this typo (`score` => `scope`).

exec/RenderHtml.hs
fisx: no need to take it back in, btw.  i've already done this in my PR.  just curious.

andorp: I just inserted there for future work.

fisx:
one more thing: we already have `DelegationNetwork`, which does the same as your new `DelegationInfo`.  either one of them should go, but it's ok if you only write a FIXME about that.

exec/RenderHtml.hs
andorp: I am happy if it works without any additional work. Previously it was a Page instance, not a FormPage.

src/Frontend/Page/Delegation.hs
andorp: Fixed in: 53e9bfd

fisx:
entering the page at dspace idea-1, but no such idea exists, and active menu entry is global.

src/Frontend/Page/Delegation.hs
```diff
@@ -166,14 +166,13 @@ instance FormPage PageDelegationNetwork where
         <$> ("scope" .: DF.choice delegationScopeList (Just actualDScope))
       where
         -- TODO: Better visual instead of show
```
fisx: dangling TODO.  i'll remove that.

fisx:
i see only "schule" and "idee" dscopes in the menu.  is that intended?

src/Persistent/Pure.hs
```diff
@@ -606,6 +608,31 @@ addDelegation = addDb dbDelegationMap
 deleteDelegation :: AUID Delegation -> AUpdate ()
 deleteDelegation did = dbDelegationMap . at did .= Nothing
 
+delegationScopeTree :: User -> Query (Tree DScopeFull)
+delegationScopeTree user = unfoldTreeM discover DScopeGlobalFull
+  where
+    discover :: DScopeFull -> Query (DScopeFull, [DScopeFull])
+    discover DScopeGlobalFull = do
+        schoolIdeas <- DScopeIdeaFull <$$> findWildIdeasBySpace SchoolSpace
+        let ideaSpaces = DScopeIdeaSpaceFull . ClassSpace <$> (user ^.. userSchoolClasses)
```
fisx: moderator should be able to see all classes.  we have that functionality somewhere, let me find it...

fisx: no, what i meant was `RoleScope` but that's different in that it drops the difference between "SchoolScope" and "all classes".  we need that distinction here.

src/Persistent/Pure.hs
```diff
@@ -606,6 +608,31 @@ addDelegation = addDb dbDelegationMap
...
+        pure (DScopeGlobalFull, schoolIdeas <> ideaSpaces)
+
+    -- FIXME: Impossible: As the Global is the School space
+    discover s@(DScopeIdeaSpaceFull SchoolSpace) = do
```
fisx: this should be removed, and the next case should cover this (just relax the pattern).

fisx:
LGTM.  will merge when green.

thanks!  (-:

fisx:
Ok, travis now runs 17 mins.  Thanks for this tweak!

PR #668, opened 2016-06-16
Unwobble buttons

### LINKED ISSUE #578: refresh after click is wobbly.

fisx:
Fixes #578.

This was surprisingly simple.  We now don't reload the page after a fixed timeout, but after the response on the POST has been received.


andorp:
LGTM

PR #675, opened 2016-06-16
delegation network visualization

fisx:


src/Persistent/Idiom.hs
```diff
@@ -207,6 +207,8 @@ ideaVoteSupportByAbsDiff idea = Support $ countVotes _Yes idea - countVotes _No
 
 -- * voting
 
+-- | The number of votes a 'User' has in a 'DScope', including own vote, direct delegations,
+-- transitive delegations, and implicit delegations from ancestor 'DScope's.
```
fisx: @andorp correct?

andorp: correct

PR #680, opened 2016-06-17
Delegation network speed up.

andorp:


src/Data/Delegation.hs
```diff
@@ -0,0 +1,163 @@
...
+
+--  Type parameters for this module is 'U' and 'S', fixing them
+-- is necessary as a workaround for an issue of 'dericeSafeCopy'.
+--
+--  The deriveSafeCopy can not add the Ord constraints when
+-- creates the SafeCopy class for the 'DelegationMap S U', which
+-- constrant is needed to be able to copy the inner map.
+--
+--  The type constants U and S are created to be able to make this
+-- module parametric easier if the issue got resolved.
```
fisx: :+1: helpful docs, good solution!

fisx:
LGTM.

i've made a quick load test using `make content-deleg`, and it appears the biggest cost is now imposed by logging to stdout.  loading the delegation page still takes a bit too long, but it's a big improvement and good enough for now.

Issue #682, opened 2016-06-19
delegations on user profile


fisx:
1. the user profile always has empty delegations.  this is suspicious, but i haven't looked into whether it's just missing demo data or whether there is something wrong with constructing or rendering it.
1. scope buttons should use the DScope selection widget from the network browser page.  (this may be a duplicate of another issue we've already open.)

fisx:
`userDelegateeListsMap` may require changing when we do this.

Issue #652, opened 2016-06-13
Show delegated votes


rittermo:
We need a solution that indicates if I already have delegated my vote to a particular user. 

Nadine designed a solution for the user profile, which shows the delegations already made in the dropdown and a little icon next to the user name, which shows the delegations on click.

![8 1 user profile delegate_02](https://cloud.githubusercontent.com/assets/15341015/16007029/83d7bc54-3170-11e6-90ef-e183708caed8.jpg)

This is somewhat dependent on the solutions we chose in #651

fisx:
related: #682

fisx:
Fixed in #699.  The "delegate" buttons are now toggles, and they read "withdraw" if a delegation is active.

@rittermo please re-open if you are unhappy with this solution.

Issue #651, opened 2016-06-13
Class-wide delegations for multiple classes


rittermo:
Now that we have multiple classes per student, the button to delegate class-wide needed to be adapted. Nadine made two designs. I favor the second one with the dropdown as the first one with the overlay would mean we have a separate page.

![8 1 user profile delegate vote_stufe](https://cloud.githubusercontent.com/assets/15341015/16006975/27798b22-3170-11e6-99f2-15c5aa789f60.jpg)
![8 1 user profile delegate_02](https://cloud.githubusercontent.com/assets/15341015/16006976/27a42ad0-3170-11e6-9fcd-3dec5bf33b8b.jpg)

fisx:
Fixed in #699.  Well, I claim it is fixed..  (-:

I've made a third design, consisting of a list of toggle buttons where the buttons originally were.  I agree Nadine's designs are nicer, but I suggest to take the short-cut and just tweak the css of what we have.

@rittermo: If you disagree, please re-open, and ideally move to it.3.

PR #687, opened 2016-06-20
Delegation in user profile

fisx:
related: #652, #651, #682

user profile has a list of buttons now, one for each class and one for school, all of them toggle.

### DIFF
--- a/src/Action.hs
+++ b/src/Action.hs
@@ -554,12 +554,12 @@ delegateTo :: ActionM m => DScope -> AUID User -> m ()
 delegateTo scope delegate = do
     user <- currentUser
     addWithCurrentUser_ AddDelegation (Delegation scope (user ^. _Id) delegate)
-    eventLogUserDelegates True scope delegate
+    eventLogUserGivesDelegation scope delegate
 
 withdrawDelegationTo :: ActionM m => DScope -> AUID User -> m ()
 withdrawDelegationTo scope delegate = do
     update $ WithdrawDelegation delegate scope
-    eventLogUserDelegates False scope delegate
+    eventLogUserWithdrawsDelegation scope delegate
 
 -- ASSUMPTION: Idea is in the given idea location.
 voteIdeaComment :: CommentKey -> Create_ CommentVote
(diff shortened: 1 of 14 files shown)


fisx:
i don't understand the error in the test suite.  @andorp can you easily see what's going on there?  i thought that path should not return 405?  perhaps something about an invalid user id in the path?  (seems unlikely to me, though.)

anyway.  giving up for today!

src/Frontend/Path.hs
```diff
@@ -736,6 +736,8 @@ isPostOnly = \case
       case m of
           UserDelegateVoteOnSchoolSpace    -> True
           (UserDelegateVoteOnClassSpace _) -> True
+          UserWithdrawDelegationOnSchoolSpace    -> True
+          (UserWithdrawDelegationOnClassSpace _) -> True
```
fisx: D'OH!  (-:

thanks...

src/Frontend/Path.hs
```diff
@@ -25,7 +25,7 @@ module Frontend.Path
     ( HasPath(..)
     , Top(..)
     , AllowedMethod(..)
-    , Main
+    , Main(Broken)
```
andorp: Please hide the `Broken` constructor.

src/Frontend/Page/User.hs
```diff
@@ -291,14 +324,14 @@ delegatedVotesClass userId = do
 
 delegatedVotes :: (ActionPersist m, ActionUserHandler m)
       => AUID User -> DScope -> m PageUserProfileDelegatedVotes
-delegatedVotes userId scope = do
+delegatedVotes userId _scope = do  -- TODO: about dropping _scope on the floor here, see #682.  see also: userDelegateeListsMap
```
andorp: Leftover `TODO` item.

src/Action.hs
```diff
@@ -867,8 +858,12 @@ eventLogUserVotesOnComment ck@(CommentKey _ ideaId parentIds _) v = do
 
 eventLogUserDelegates ::
       (ActionUserHandler m, ActionPersist m, ActionCurrentTimestamp m, ActionLog m)
-      => DScope -> AUID User -> m ()
-eventLogUserDelegates scope delegateId = do
+      => Bool -> DScope -> AUID User -> m ()
+eventLogUserDelegates setOrWithdraw scope delegateId = do
```
andorp: I would introduce two functions on top of this one. The two would be `eventLogUserSetsDelegate` and `eventLogUserWithdrawsDelegate`.

src/Frontend/Path.hs
fisx: why?  won't do it!  (-:

i don't see any benefit (it's called "broken" and shouldn't even be locally defined in this module, and i was confused by the fact it was hidden and have wasted some time figuring out why.)

src/Frontend/Page/User.hs
fisx: there are a few more.  i would like to resolve them separately.

src/Action.hs
fisx: good point.  will do.

Issue #597, opened 2016-05-29
Cancel button for create topic


rittermo:
On the create topic page, it would nice to have a cancel button labelled “Abbrechen”  which takes me back to the topics list.

fisx:
related #603

PR #688, opened 2016-06-20
Cancel button on create and edit topic pages.

andorp:
Fixes: #597 

- On create topic page it points to the list topics page.
- On edit topic page goes back to the topic itself.


fisx:
you could have made the `cancelButton` combinator powerful enough to handle both cases at this point, but ok.  (-:

LGTM, thanks!

fisx:
(never mind, `cancelButton` is only interesting because we don't have to worry about where to aim it.)

PR #689, opened 2016-06-20
Delegation.

fisx:


fisx:
ready for review.  @andorp?  you don't need to read the `d3-aula.*` stuff, that's still work in progress, i will work on that more with @localgrr later today.

Issue #599, opened 2016-05-29
Sorting for created ideas on user profile


rittermo:
The list of ideas that a user has created, which is shown on the user‘s profile, should have the sorting  dropdown to sort the ideas by popularity or date.

PR #701, opened 2016-06-21
Filter ideas on user profile.

andorp:
Fixes: #599


src/Frontend/Fragment/WhatListPage.hs
```diff
@@ -0,0 +1,43 @@
+{-# LANGUAGE DataKinds         #-}
+{-# LANGUAGE DeriveGeneric     #-}
+{-# LANGUAGE TemplateHaskell   #-}
+
+module Frontend.Fragment.WhatListPage
```
fisx: we can keep this "fragment", but i wouldn't have started this module: it doesn't have a `ToHtml` instance, and it's very tightly related to the idea lists module it is coming from.

fisx: anyway, small modules are good.

fisx:
Cool!  It looks so easy.  (-:

LGTM, can be merged when green.

PR #717, opened 2016-06-23
2016 06 cc header tidy

### LINKED ISSUE #643: tidy up button design.
in #627, in `ViewIdea` we move all buttons into one block so they can be designed together.  We should do the same thing for all applicable pages, in particular:

- `PageOverviewOfWildIdeas`
- `PageOverviewOfTopics`
- `PageUserProfile*`

(there are a number of admin pages that could benefit from that, but usability for admins is not a priority.)

localgrr:
Fixes #643.

Some general header and user profile changes, none were specific issues. I think buttons are too big in idea admin headers, at least for desktop. It makes sense they are really big for tablets. Made an adjustment for that


.gitignore
```diff
@@ -23,3 +23,4 @@ aula.yaml*
 /static/avatars/*.png
 /hpc
 cookie-jar
+src/Backend.hs
```
fisx: nope, we still want this file to stick around...  if you have local changes, you just need to add around them manually.  (i use git-cola for that, but i always suspected there are better tools, too.)

fisx:
LGTM.  can be merged when green.

PR #724, opened 2016-06-23
Avatar URLs are no longer stored

np:
* Avatar URLs are computed only from the user id and avatar dimension
* Convenience functions are provided to access different dimensions


src/DemoData.hs
```diff
@@ -172,20 +162,27 @@ genCommentVote comments_in_context students = do
     let action = addWithUser . AddCommentVote $ comment ^. _Key
     action student <$> arb
 
-updateAvatar :: User -> URL -> forall m . ActionM m => m ()
-updateAvatar user url = update $ SetUserAvatar (user ^. _Id) url
+updateAvatar :: User -> FilePath -> forall m . ActionM m => m ()
+updateAvatar user file = readImageFile file >>= \case
+    Right pic -> saveAvatar (user ^. _Id) pic
+    Left err -> fail err
```
andorp: We should use `addMessage` and not fail here.

np: I don't mind the use of addMessage but so far this is in DemoData.

np:
Notice that this commit also fixes the following issue:

* New user creates contents but has no avatar yet (`metaCreatedByAvatar = Nothing` on the new created objects)
* The user then changes avatar.
* Contents created before that still has no image.

src/DemoData.hs
andorp: Sorry, missed the demo data part. In this case I am ok just having the `fail` combinator here.

andorp:
LGTM

PR #725, opened 2016-06-23
No empty brackets in Role UI label for empty school name.

andorp:


src/Types.hs
```diff
@@ -519,6 +519,10 @@ data SchoolClass = SchoolClass
     }
   deriving (Eq, Ord, Show, Read, Generic)
 
+nilSchoolClass :: SchoolClass -> Bool
+nilSchoolClass (SchoolClass _ "") = True
+nilSchoolClass _                  = False
```
fisx: i will merge this, but:

1. i think there is a prettier solution with some of the more recent pattern-related language extensions.  will look into it or add a FIXME.
1. i will add a FIXME that this shouldn't be happening in the first place.  `""` is no substitute for `Nothing`.  (i'll look into this again also.)

andorp: I totally agree. The `Role` should be a parametric type.

PR #722, opened 2016-06-23
Delegate on topic, idea level

fisx:
Fixes #720.

PR #727, opened 2016-06-23
Delegate on topic, idea level

### LINKED ISSUE #720: delegate on topic, idea level.
we have the functionality for that, but we need to change the design.  this involves reproducing the mechanics of the category filter for lists of users of dynamic length.

2016-06-23-142656_1600x900_scrot

fisx:
Obsoletes #722.  Fixes #720.


src/Action.hs
```diff
@@ -550,6 +551,10 @@ voteOnIdea ideaId voteVal = do
         addWithCurrentUser_ (AddVoteToIdea ideaId delegatee)
                             (ProtoIdeaVote voteVal (voter ^. _Id))
 
+delegateOrWithdraw :: ActionM m => DScope -> Maybe (AUID User) -> m ()
+delegateOrWithdraw scope (Just delegate) = delegateTo           scope delegate
+delegateOrWithdraw _scope Nothing         = error "withdrawDelegationTo scope delegate"  -- FIXME
```
andorp: A `pure ()` should be better here.

fisx: could be.  i wanted to make it obvious that we should resolve #730.

PR #729, opened 2016-06-23
Render voting power for the explicit scope only.

andorp:


fisx:
LGTM!

i really like this solution!  we need to think about the expensive `votingPower`, too, but we don't need to do it now.

can be merged when green.

Issue #728, opened 2016-06-23
Refactor delegate topic level, idea level form: show current delegate when opening.


fisx:
There is some code duplication here:

- https://github.com/liqd/aula/blob/f7d163e6fc337a5f274200ddb768e9355e0b7d36/src/Frontend/Page/Delegation.hs#L96
- https://github.com/liqd/aula/blob/f7d163e6fc337a5f274200ddb768e9355e0b7d36/src/Frontend/Page/Delegation.hs#L110

if we want to pre-select the current delegate of the current user, these two code blocks will grow symmetrically, so we should factor them out.

When I tried I noticed that there is a reduced `Either` type for two `DScope` constructors (because we don't need the other ones *yet*).  I suggest we change this as follows:

```haskell
data PageDelegateVote = PageDelegateVote DScopeFull [User] (Maybe (AUID User))
```

This will both make our live easier now in this refactoring, and make it easier to introduce similar delegation buttons on idea space level.

fisx:
the refactoring may not be so interesting unless we can use all constructors right away.  think about this more!

fisx:
(meaning: we may want to fix this without changing the `Either`.)

PR #739, opened 2016-06-24
Misc fixes.

fisx:
Fixes the following unreported issues:

- delegate on user profile doesn't reload properly
- if you delegate to yourself and then close your node's delegatees in the graph, you close the node itself.
- delegate-to page behind topics, ideas: current delegation is rendered correctly (still not fixed: #728).

### DIFF
--- a/static/d3-aula.js
+++ b/static/d3-aula.js
@@ -266,7 +266,7 @@
         // be slightly nicer to remember which nodes were invisible
         // and recover the state before the previous click on the root
         // node.)
-        var on_click = function(d) {
+        var on_click = function(clickee) {
             var newVisibilityStatus = undefined;
             var visited = [];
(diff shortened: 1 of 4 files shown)


static/d3-aula.js
```diff
@@ -266,7 +266,7 @@
         // be slightly nicer to remember which nodes were invisible
         // and recover the state before the previous click on the root
         // node.)
-        var on_click = function(d) {
+        var on_click = function(clickee) {
```
andorp: This should be called `clickEvent`.

fisx: this is the node that is clicked on, not the event.  if anything, we could call it delegate.  i will change that.

andorp:
LGTM

Issue #733, opened 2016-06-24
`src/Types.hs:1704:38-55: Non-exhaustive patterns in lambda`


fisx:
I got this when generating 30 demo delegations on DScopeGlobal.  When using the built-in defaults, everything works fine.

PR #742, opened 2016-06-24
Fix leaves in delegation graph.

andorp:
Fixes: #733


andorp:
I've tried a simpler approach:

````haskell
fixLeaves :: Ord k => [(k,k,[k])] -> [(k,k,[k])]
fixLeaves nodes = nodes <> leaves
  where
    leaves = (\k -> (k,k,[])) <$> ((\(_,_,ks) -> ks) =<< nodes)
````

but it gives an invalid SCC graph: 
````haskell
showSCC <$> stronglyConnComp (fixLeaves [(1,1,[2,3]),(3,3,[1])])
["3","2","1","[1,3]"]
````

fisx:
thanks!

Issue #698, opened 2016-06-21
delegating likes


fisx:
current behavior: delegating does not apply to likes (quorum votes).

desired behavior: needs clarification.

fisx:
note that you can't un-like currently.  we probably don't want to have "delegate likes" without "withdraw likes".

PR #749, opened 2016-06-27
Delegating likes

andorp:
Fixes: #698

### DIFF
--- a/src/Action.hs
+++ b/src/Action.hs
@@ -531,33 +531,50 @@ moveIdeaToTopic ideaId moveIdea = do
 
 -- * Vote Handling
 
+delegatedOperationOnIdea
+    :: ActionM m
+    => (AUID User -> AUID Idea -> EQuery (Maybe (User, a)))
+    -> (User -> User -> m ())
+    -> AUID Idea
+    -> m ()
+delegatedOperationOnIdea getOperationResult operation ideaId = do
+    -- Like idea
+    user <- currentUser
+    let scope = DScopeIdeaId ideaId
+    operation user user
+    equery (votingPower (user ^. _Id) scope)
+        >>= filterM hasDoneItAlready
+        >>= mapM_ (operation user)
+  where
+    hasDoneItAlready :: ActionM m => User -> m Bool
+    hasDoneItAlready delegatee = equery $ do
+        let delegateeId = delegatee ^. _Id
+        mlike <- getOperationResult delegateeId ideaId
+        pure $ case mlike of
+            Nothing -> True
+            Just (user', _result) -> delegateeId /= (user' ^. _Id)
+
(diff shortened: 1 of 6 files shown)


src/Persistent/Pure.hs
```diff
@@ -737,11 +737,13 @@ findComment :: CommentKey -> MQuery Comment
...
 
 instance FromProto IdeaLike where
-    fromProto () = IdeaLike
+    fromProto p m = IdeaLike m (_protoIdeaLikeDelegate p)
 
 -- FIXME: Assumption: the given @AUID Idea@ MUST be in the DB.
-addLikeToIdea :: AUID Idea -> AddDb IdeaLike
-addLikeToIdea iid = addDb' (mkIdeaVoteLikeKey iid) (dbIdeaMap . at iid . _Just . ideaLikes)
+addLikeToIdea :: AUID Idea -> User -> AddDb IdeaLike
+addLikeToIdea iid user =
```
fisx: is user the delegatee or the delegate?  this function could use a line of documentation.

fisx: also, if one of `delegate` or `delegatee` is a good name for something, then `user` probably isn't.

src/Action.hs
```diff
@@ -531,33 +531,50 @@ moveIdeaToTopic ideaId moveIdea = do
...
 -- * Vote Handling
 
+delegatedOperationOnIdea
+    :: ActionM m
+    => (AUID User -> AUID Idea -> EQuery (Maybe (User, a)))
+    -> (User -> User -> m ())
+    -> AUID Idea
+    -> m ()
+delegatedOperationOnIdea getOperationResult operation ideaId = do
+    -- Like idea
```
fisx: what does this comment mean?

src/Action.hs
```diff
@@ -531,33 +531,50 @@ moveIdeaToTopic ideaId moveIdea = do
 
 -- * Vote Handling
 
+delegatedOperationOnIdea
```
fisx: write a haddock comment?

fisx:
LGTM otherwise.

src/Persistent/Pure.hs
andorp: The user who does the like. It can be delegate and delegatee for this idea.

andorp: Oh I see, sorry.

src/Action.hs
fisx: arguments:

1. get the delegate value (together with an operation result that is ignored).  rename to `getDelegate`.
1. run a new operation with delegate and delegatee on
1. this idea.

delegatee is current user.

fisx: correction: *delegate* (the one who casts the vote) is current user!

PR #751, opened 2016-06-27
User upload page styling

### LINKED ISSUE #601: Show upload constraints in upload form
User Profile – edit profile: The size limit and file format constraints for uploading pictures are not given on the form. We should add them underneath the upload button.

A NTH: There could be a preview of the current profile pic (if there is any) left of the upload field for the new pic

fisx:
Fixes #601

### DIFF
--- a/src/Action.hs
+++ b/src/Action.hs
@@ -836,7 +836,12 @@ instance HasUILabel PhaseShiftResult where
 -- * files
 
 class Monad m => ActionAvatar m where
-    readImageFile :: FilePath -> m (Either String DynamicImage)
+    -- Yield a 'Nothing' if the image file is empty, a 'Just . Left' if the image cannot be parsed,
+    -- and a 'Just . Right' image value otherwise.  The 'Maybe' is necessary to work around the fact
+    -- that digestive-functors returns an empty file if the user is not uploading anything.
+    --
+    -- FIXME: figure out how to do that better with digestive-functors.
+    readImageFile :: FilePath -> m (Maybe (Either String DynamicImage))
     savePngImageFile :: FilePath -> DynamicImage -> m ()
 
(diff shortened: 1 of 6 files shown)


src/Action.hs
```diff
@@ -836,7 +836,12 @@ instance HasUILabel PhaseShiftResult where
 -- * files
 
 class Monad m => ActionAvatar m where
-    readImageFile :: FilePath -> m (Either String DynamicImage)
+    -- Yield a 'Nothing' if the image file is empty, a 'Just . Left' if the image cannot be parsed,
+    -- and a 'Just . Right' image value otherwise.  The 'Maybe' is necessary to work around the fact
+    -- that digestive-functors returns an empty file if the user is not uploading anything.
+    --
+    -- FIXME: figure out how to do that better with digestive-functors.
+    readImageFile :: FilePath -> m (Maybe (Either String DynamicImage))
```
andorp: I would prefer an ADT here, but for the time being with the comment is ok.

src/Frontend/Constant.hs
```diff
@@ -35,3 +35,8 @@ avatarDefaultSize = 100
 
 avatarExtraSizes :: [Int]
 avatarExtraSizes = [64, 300]
+
+-- | This is just for showing the information.  To enforce this limit, set up a reverse proxie like
+-- nginx and configure it accordingly.  See ./docs/deployment.md for details.
+avatarMaxByteSize :: ST
+avatarMaxByteSize = "10MB"
```
andorp: This should also be part of the config file, as the adminsitrators of aula can adjust aula and nginx without recompile.

andorp:
Two minor comments. LGTM

fisx:
@localgrr can you work on the styling here?  login as admin, select "profil" from main menu, then 'profil bearbeiten" (edit profile).

also, could you do it on another branch based on this one?

PR #753, opened 2016-06-27
2016 06 cc minor aesthetic

localgrr:
Some little bits to make things look nicer.


fisx:
undid one glitch and one issue fixed elsewhere, otherwise LGTM.  will merge when green.

PR #777, opened 2016-06-29
Fix test suite

fisx:


tests/DelegationSpec.hs
```diff
@@ -45,10 +46,11 @@ universeSize = UniverseSize
 
 spec :: Spec
 spec = do
-    runner   <- runIO createActionRunner
-    persist  <- runIO Persistent.mkRunPersistInMemory
-    uni      <- runIO $ unNat (runner persist) (mkUniverse universeSize)
-    snapshot <- runIO $ unNat (runner persist) getDBSnapShot
+    let catchAll (SomeException msg) = error . ("DelegationSpec.hs: " <>) . show $ msg
```
andorp: This could be the part of `runIO'`.

tests/DelegationSpec.hs
```diff
@@ -68,7 +70,7 @@ spec = do
             Nothing -> error "No idea with topic is found."
             Just i  -> (i ^. _Id, fromJust (i ^? ideaLocation . _IdeaLocationTopic . _2))
         Just ideaspace = find (has _ClassSpace) $ unIdeaSpaces uni
-    let noChecks (CheckNumVotingPower{}) = False
+    let noChecks (CheckVotingPower{}) = False
```
andorp: indentation

tests/DelegationSpec.hs
fisx: then we couln't use `runIO'` elsewhere to throw something else, like the name of another test module.  or return, if we don't want to throw anything.

PR #778, opened 2016-06-29
Refactor Types module

fisx:
Split up `Types` into `Prelude`, `Core`, `Instances.*`, `Util`.  `Core` is still quite big, and because of the cyclical deps I refrained from making type-centered modules (`User`, `Idea`, ...).  But I think it's still an improvement.  I also removed (very few) helper functions.


fisx:
more or less done, but waiting for #777 to land first.

it would be interesting compare the compile time of 2a61e2f and ded2ce1 in some more controlled setting than travis, too.

fisx:
rebased.

fisx:
A quick test on my machine showed no significant speed-up in compilation time.  I would like to merge the changes anyway for readability.

andorp:
This is a big change, with lot of shuffling, but I have not spotted out any bad change.

andorp:
LGTM

PR #790, opened 2016-06-30
Delegation network, again.

### LINKED ISSUE #690: delegation network browser nagivation widget
the current mechanics shows two buttons "aufklappen" and "zuklappen" (open new level, close level, resp.).  there is an easier way to do that.

1. the "close" button goes next to the menu and becomes a little icon (perhaps `[X]` or something).  if there are several close-able levels, we have several `[X]`.  this has the additional benefit that we can close arbitrarily many levels with one click.
1. the "open" button goes away.  instead, the new level opens always, and shows a special entry "no selection", which means that the parent level is selected.  so if you select a topic, the ideas menu will pop open, but set itself to "no selection", and the topic graph will be shown.
1. if you click `[X]`, all the current level and all below will be closed.  if you select "no selection", all levels below will be closed.

the "load" button may not be necessary any more after this.  every new selection can trigger a reload.

### LINKED ISSUE #783: Delegation graph: Optimization of navigation between scopes
As discussed:

- there should be an automatic expand when a scope other than „all“ is selected in a scope select list
- there should be an „x“ next to the lowest scope select to close it
- the graph should automatically reload on scope change in the select list
- NTH: the „wild ideas space“ is treated as a topic and the wild ideas are visible when the space was selected in the topic scope select list

---

From the pad:

Beauftragungsnetzwerk https://aula-stage1.liqd.net/delegation/view?scope
Die Buttons "Aufklappen" und "Zuklappen" sind bei der Menüführung unintuitiv. Ich habe 5 Minuten gebraucht, bis ich raus hatte, wie man die Ideenräume eingrenzen kann.   ***Alexa: War bei mir auch so***
Ist es möglich, beide Buttons zu entfernen und die Funktionalität "Aufklappen" automatisch bei anklicken eines Menüpunkts auszuführen?
Dazu ist die erste Dropdown ja überflüssig, weil sie grundsätzlich nur einen Menüpunkt enthält.
Auch "Anzeigen" sollte möglichst automatisch bei der Wahl des Ideenraums erfolgen. Es ist sonst verwirrend und umständlich zu bedienen.

fisx:
Fixes #783.  Fixes #690.

### DIFF
--- a/src/Frontend/Page/Delegation.hs
+++ b/src/Frontend/Page/Delegation.hs
@@ -147,6 +147,7 @@ instance Aeson.ToJSON DScopeForest where
 
 instance Page PageDelegationNetwork where
     isAuthorized = userPage -- FIXME who needs to see this
+    isResponsive _ = False
     extraFooterElems _ = do
         script_ [src_ $ U.TopStatic "third-party/d3/d3.js"]
         -- FIXME: move the following two under static-src and sass control, resp.?
--- a/static/d3-aula.js
+++ b/static/d3-aula.js
@@ -80,7 +80,7 @@
             select.enter()
                 .append("select")
                 .attr("name", function(d) { return d.ancestors[d.ancestors.length - 1]; })
-                .on("change", function(d) { current = this.value; update(); });
+                .on("change", function(d) { document.location.href = "/delegation/view?scope=" + this.value; });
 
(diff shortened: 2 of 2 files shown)


fisx:
only one trivial change outside of d3-aula.js.  merging.

Issue #700, opened 2016-06-21
what to do with distinction between 'global' and 'idea space "school"'?


fisx:
in delegation, it makes a difference:

1. 'global' means all idea spaces, topics, ideas.
1. 'idea space "school"' means all ideas and topics that affect the entire school and are therefore created in the corresp. idea space.

options:

1. cheat, and pretend on the UI that there is no such distinction.
1. make it very obvious and make it a topic in the manual, so that teachers will be able to figure this out.

the current status is that we sometimes make the distinction, but we should also check everything again to make sure we're consistent.

PR #791, opened 2016-06-30
Unify Global and SchoolSpace delegation scopes.

andorp:
Fixes.#700.


fisx:
it turns out this is the wrong approach.  entirely my fault, sorry!  we should not have "global" at all, but end the forest on the idea space level.

the current solution would naturally lead to a situation where school-space ideas are dangling off the "global" node in the dscope tree.  not what we want!

sorry again..

fisx:
(please cherry-pick ddf9882 before closing, though.)

andorp:
Closing in favor of #795 .

PR #794, opened 2016-06-30
Render implicit delegation on the delegation view page.

andorp:


src/Persistent/Pure.hs
```diff
@@ -709,8 +710,20 @@ scopeHiearchy = \case
             IdeaLocationSpace s    -> DScopeIdeaSpace s
             IdeaLocationTopic _s t -> DScopeTopicId   t)
 
-findDelegationsByScope :: DScope -> Query [(Delegate (AUID User), DScope, [Delegatee (AUID User)])]
-findDelegationsByScope = views dbDelegations . Data.Delegation.findDelegationsByScope
+findDelegationsByScope :: DScope -> Query [(Delegate (AUID User), [(DScope, Delegatee (AUID User))])]
```
fisx: it seems to me you are nesting the list here just to flatten it out again later above.  i think i'll revert that.

fisx:
LGTM, except for one non-trivial change.

@andorp could you take another look if i got this right?  Please merge if you and travis are ok.

fisx:
On second thoughts, I will merge this.  (Please take another look anyway.)

src/Persistent/Pure.hs
```diff
@@ -710,20 +710,12 @@ scopeHiearchy = \case
...
-    hiearchy <- scopeHiearchy scope
-
-    let addScopeDeleg m scope' = do
-            m' <- Map.fromList <$> Persistent.Pure.findDelegationsByScope scope'
-            pure $ Map.unionWith (++) m m'
-
-    Map.toList <$> foldM addScopeDeleg Map.empty hiearchy
+findImplicitDelegationsByScope :: DScope -> EQuery [(Delegate (AUID User), DScope, [Delegatee (AUID User)])]
+findImplicitDelegationsByScope scope =
+    mconcat <$> (scopeHiearchy scope >>= mapM Persistent.Pure.findDelegationsByScope)
```
andorp: It won't work. The first element of the triplets has to be unique, otherwise the graph rendering does not work well, for that reason was there the Map mumbo jumbo.

For this input `[ (1,s0,[2,3,4]), (2,s0,[4,7,1]), (1,s2,[])]` the Data.Graph will count no outgoing edges for the node `1`, which leads misscalculation of the voting power, the D3 draw different graph tha we use for the computation of the voting power.

andorp: I'll write a version with `Map` here.

PR #795, opened 2016-06-30
Remove global delegation scope

### LINKED ISSUE #700: what to do with distinction between 'global' and 'idea space "school"'?

andorp:
Fixes #700.


HLint.hs
```diff
@@ -16,3 +16,4 @@ ignore "Use const"
 ignore "Use fmap"
 ignore "Use list literal"
 ignore "Use record patterns"
+ignore "Use =<<"
```
fisx: Perhaps we can contain that to one module?  I have no strong opinion here, but it would be nice to keep reminded by hlint if we bump into this again, and decide on a case-by-case basis.

src/Frontend/Page/Delegation.hs
```diff
@@ -130,19 +130,21 @@ topicDelegation tid = formPageHandlerCalcMsgM
...
   deriving (Eq, Show, Read)
 
-instance Aeson.ToJSON DScopeTree where
-    toJSON (DScopeTree t) = f t
+-- TODO: Fix JSON instance
+instance Aeson.ToJSON DScopeForest where
+    toJSON (DScopeForest ts) = f ts
       where
-        f (Tree.Node dscope chldrn) = Aeson.object
+        f [] = error "impossible"
```
fisx: that's not really impossible, though, right?  isn't it a legitimate corner case to have no dscopes and no delegations?

fisx: ah, that's probably the TODO's job to sort out.  (-:

fisx:
LGTM.

can be merged when green.  (please not the nit-picks, as usual.)

PR #797, opened 2016-07-01
Encode assumption for delegates in return type of findDelegationsByScope.

andorp:


fisx:
what assumption is encoded?  what problem does is solved by this PR?

you made that change before, and i took it back.  so yes, i started the edit war, but it would be a more productive one if you gave some reasons for keeping it going, or debunked reasons for my last move.

fisx:
ah, the discussion went on in #794.  found it!  (-:

anyway i'm not convinced.  are you saying i'm breaking the assumption, or are you saying i am hiding it, and thus make the code less robust?

i haven't looked back at the code right now, but i thought that your and my output of `findDelegationsByScope` are identical up to the grouping / flatttening.

src/Frontend/Page/Delegation.hs
```diff
@@ -181,22 +182,18 @@ viewDelegationNetwork (fromMaybe (DScopeIdeaSpace SchoolSpace) -> scope) = do
...
     let (delegationGraph, _vertexToGraphNode, nodeToVertex) = graphFromEdges graphNodes
     let uidToVertex = fromJust . nodeToVertex
     let graphComponents = stronglyConnComp graphNodes
 
-    -- Count number of inbound delegation edges in local 'DScope'.  This is not the 'votingPower'
-    -- because it does not take into account delegations from surrounding 'DScope's, but it is much
-    -- cheaper to calculate.  (FUTUREWORK: it would be nice to have the actual voting power here,
-    -- but that should go together with showing the implicit (inherited) delegation edges in the
-    -- graph as well, e.g. as dashed lines.)
+    -- Count number of inbound delegation edges in implicit 'DScope'.
```
fisx: An implicit `DScope` could be an ancestor to the local one.  I think what you want to say here is

```haskell
    -- Count voting power (number of inbound delegation edges in the local 'DScope' and all its ancestors).
```

src/Frontend/Page/Delegation.hs
```diff
@@ -181,22 +182,18 @@ viewDelegationNetwork (fromMaybe (DScopeIdeaSpace SchoolSpace) -> scope) = do
...
 delegationInfos scope = do
-    delegations <- findImplicitDelegationsByScope scope
+    delegations <- Map.toList <$> findImplicitDelegationsByScope scope
 
     -- Create delegations
-    let mkGraphNode (de, _, dees) = (unDelegate de, unDelegate de, unDelegatee <$> dees)
+    let mkGraphNode (de, dees) = (unDelegate de, unDelegate de, unDelegatee . snd <$> dees)
 
     -- Build graphs and graph handler functions
     let graphNodes = fixLeaves $ mkGraphNode <$> delegations
```
fisx: ok, got it.  if we didn't have `delegations` in the form preferred by you, this is where would have to group it at the lastest.

so why don't we do it here?  make it an independent function, obviously, but call that function here?  then we wouldn't have to flatten it back below.

PR #800, opened 2016-07-02
empty.

fisx:


exec/RenderHtml.hs
```diff
@@ -138,6 +139,7 @@ refreshSamples = do
 
 -- | Call tidy if available; generate either pretty-printed html or an error message if the html is
 -- invalid.
+-- FIXME: Configurable language instead of fixed `DE`
```
fisx: Please eliminate this `FIXME` and use `whereToGetTheLangValue` instead of `DE` (which is an implicit `FIXME`).

fisx:
Thanks, very useful stuff!

Apart from my last comment: as discussed offline, let's move the last two commits to another PR and land the rest now.

tests/AulaTests.hs
```diff
@@ -45,11 +46,24 @@ import Logger (LogLevel(..), nullLog)
 import Persistent (mkMetaInfo)
 
 
--- | 'runIO" with better errors.
+-- | 'runIO' with better errors.
```
np: The function is called `runIO'` so I guess `@runIO'@` would be better.

fisx: haddocks allows `'runIO''`, I think, but what I really want to say here is that `runIO'` is like `runIO`, but with better errors.  So I'm actually referring to `runIO`, not `runIO'`.

i'm confused now.  (-:

tests/Frontend/CoreSpec.hs
```diff
@@ -158,12 +158,14 @@ selectValue :: (Show a, Eq a) => ST -> View (Html ()) -> [(a, LT.Text)] -> a ->
...
+                                   , "in values", show xs
+                                   , "and choices", show choices
+                                   ]
   where
-    ref'    = absoluteRef ref v
-    value i = ref' <> "." <> i
-    choices = fieldInputChoice ref v
-    test (_, sx :: Html (), _) = showValue x == renderText sx
+    value i = absoluteRef ref v <> "." <> i
+    choices = (\(t, h, b) -> (t, renderHtmlDefaultH h, b)) <$> fieldInputChoice ref v
```
np: `_2 %~ renderHtmlDefaultH`

tests/Frontend/CoreSpec.hs
```diff
@@ -392,8 +393,9 @@ postToForm (FormTest g c check) = do
         payload       <- pick (arbFormPagePayload page)
         (v, mpayload) <- run $ simulateForm c page ctx payload
         case mpayload of
-            Nothing       -> fail $ unwords
-                                ("Form validation has failed:" : map show (viewErrors v))
+            Nothing -> fail $ unwords
+                ("Form validation has failed:" :
+                    (show . Control.Arrow.second renderHtmlDefaultH <$> viewErrors v))
```
np: `_2 %~ renderHtmlDefaultH`

tests/AulaTests.hs
np: Oops you're right.

np:
LGTM

PR #802, opened 2016-07-03
upgrade to servant-0.7, lts-6.3.  Custom SetCookie for AuLa

np:


fisx:
forgot to commit `serveFAction`?

fisx:
nope, just waiting for liqd/thentos/pull/508

fisx:
(we should make liqd/thentos#508 a new version and depend on that here.  which means we also need to upgrade to servant 0.7.)

fisx:
when i guild on my machine:

```
src/Persistent/Pure.hs:304:19:
    Not in scope: ‘throwError’
    Perhaps you meant ‘throwError500’ (imported from Servant.Missing)
```

is there a lower version bound missing somewhere in package.yaml?

fisx:
```
src/Action/Implementation.hs:26:8:
    Ambiguous module name ‘Crypto.Random’:
      it was found in multiple packages:
      crypto-api-0.13.2@crypt_Cf9PpCUM2HxIcKjYUCdWjU crypto-random-0.0.9@crypt_DVWUcfzeifIK6JODez5ilW cryptonite-0.7@crypt_Jpop98m0R6wALlgoxV81Nx
```

this is most likely a stale-sandbox issue.  i'll take a look.

fisx:
this may also be due to the fact that i am running in ghci, and the cabal constraints have no effect.  i will try to fix it with `-XPackageImports` if that is the case.

fisx:
ok, cabal build works for me.  `-XPackageImports` should work, I'll test and commit that in a moment.

> 09:22 < npouillard> I'd be interested to if it works for you to use https://www.stackage.org/lts-6.3/cabal.config but with uri-bytestring ==0.1.9.2

asks for `--force-reinstalls`, which isn't that surprising.  i don't think it's interesting to take out all conflicting libraries by hand, so i'll try the `--force-reinstalls`.

fisx:
LGTM, thanks!

Please merge when green.

I'll post the outcome of the `cabal.config` experiment here later, but I expect no interesting problems there.

fisx:
build via cabal with lts-6.3 freeze file was successful!

np:
We should commit that file then!

On July 7, 2016 11:54:11 AM GMT+02:00, fisx <notifications@github.com> wrote:
>build via cabal with lts-6.3 freeze file was successful!
>
>---
>You are receiving this because you modified the open/close state.
>Reply to this email directly or view it on GitHub:
>https://github.com/liqd/aula/pull/802#issuecomment-231034639

-- 
Sent from my Android device with K-9 Mail. Please excuse my brevity.

fisx:
i'm slightly against it.  `stack.yaml` documents which stack resolver will work, and the few people who insist on using cabal but don't want to fiddle with the constraints can just download it themselves.  plus it would only be another redundant build file that we need to keep in sync with the rest.

PR #806, opened 2016-07-04
Delegation network, again.

### LINKED ISSUE #731: delegation network: canvas size must adapt to number of nodes.
See here: https://github.com/liqd/aula/pull/727/commits/76aaae0c435006da69fd340311b878a0c8567c06

`globalGraphHeight` needs to be a function that takes an array of nodes from `force`, and calculates a pixel number based on their total voting power.  resize can probably take place in `updateWidget`, and may have to adjust both `force` and `svg`.  alternatively, `svg`, may have a scroll bar and fixed height, but that way we risk nested scroll bars.

fisx:
Fixes #731.

### DIFF
--- a/static/d3-aula.js
+++ b/static/d3-aula.js
@@ -374,7 +374,7 @@
             d.visibleByClick = true;
         };
 
-        graph.nodes.forEach(makeAllVisible);;
+        graph.nodes.forEach(makeAllVisible);
 
         var force = d3.layout.force()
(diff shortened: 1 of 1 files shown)


fisx:
it's done, but i also want to implement another way of calculating node sizes and smuggle it in here.

fisx:
(opened #808 for the node size thing.)

PR #808, opened 2016-07-04
Make node sizes more evenly distributed between min and max.

fisx:
One NTH feature, one issue fix.


fisx:
this approach is problematic because it makes nodes with the same voting power have different sizes.  we should fix that before putting this to use.  (take the median for all nodes of one size?)

PR #831, opened 2016-07-06
Delegation network, again.

### LINKED ISSUE #825: delegation network: show delegatees and delegates of users matching search expressions.
(optionally limit the recursion depth?)

### LINKED ISSUE #827: delegation network: wording in menus.
'*' should be "alle Themen" (there is now only level once we have resolved #826.

fisx:
Fixes #827.  Fixes #825.

The user instructions have changed a bit again:

1. nodes that have hidden edges now look a little transparent so you know you have to click on them.  1. if you click on a node, all direct neighbours are toggled.  you have to keep clicking if you want to see more.
1. if you search for users, you only see matching nodes, but you can see by their transparency which ones to click on.
1. if you click on a node, all neighbors are made visible if and only if there is at least one invisible neighbor.  (before a click changed the visibility of all neighbors to the opposite of an arbitrary neighbor.)

i tried to unfold neighbors during the search, but it didn't work in the time box i had.

### DIFF
--- a/src/Persistent/Pure.hs
+++ b/src/Persistent/Pure.hs
@@ -157,7 +157,7 @@ import Control.Lens
 import Control.Monad.Except (MonadError, ExceptT(ExceptT), runExceptT, throwError)
 import Control.Monad.Reader (MonadReader, runReader, asks)
 import Control.Monad.State (MonadState, gets, put)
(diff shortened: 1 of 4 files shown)


fisx:
ready for review!

andorp:
LGTM

PR #836, opened 2016-07-07
Fix docker commit in travis config.

fisx:


fisx:
@np like this?

fisx:
(got the wrong commit, though...)

PR #838, opened 2016-07-07
Switch to thentos-frontend-session

np:


fisx:
cool, thanks!  now we should start releasing thentos-* on hackage and get rid of the git submodule, but let's do that in another PR.

PR #841, opened 2016-07-07
Delegations on user profile

andorp:
The delegation buttons on profile should work differently if the user opens his/her own profile.

### DIFF
--- a/src/Frontend.hs
+++ b/src/Frontend.hs
@@ -165,7 +165,7 @@ type AulaMain =
   :<|> "admin" :> AulaAdmin
 
        -- delegation network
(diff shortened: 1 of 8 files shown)


src/Frontend.hs
```diff
@@ -165,7 +165,7 @@ type AulaMain =
   :<|> "admin" :> AulaAdmin
 
        -- delegation network
-  :<|> "delegation" :> "edit" :> FormHandler PageDelegateVote
+  :<|> "delegation" :> "edit" :> DScope ::> FormHandler PageDelegateVote
```
fisx: cool.  i totally forgot about this end-point.  (-:

fisx:
thanks!  merging, but note i opened an issue.

Issue #678, opened 2016-06-17
MultiUser can Report itself on the User Profile page.


andorp:
See attached screenshot with user `fazg`

![2016-06-17-122934_1600x900_scrot](https://cloud.githubusercontent.com/assets/3465327/16148094/9ccdb9ec-3487-11e6-95b3-b4ef05fdb33a.png)

![2016-06-17-123024_1600x900_scrot](https://cloud.githubusercontent.com/assets/3465327/16148100/a025eefc-3487-11e6-8dd4-a1b52f61d413.png)

fisx:
this may happen to students with several classes.  some multiple-personalities thing where a student that has a class A can report on all students that have a class other than A.

fisx:
(this is meant by the term "MultiUser" in the title :-)

fisx:
this will be resolved with #828.

andorp:
I would have this opened till we resolve it.

PR #844, opened 2016-07-11
User cannot report himself

andorp:
Fixes: #678

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -2,7 +2,7 @@ SHELL=/bin/bash
 EXEC=`test -d .stack-work/ && echo "stack exec --" || echo "cabal exec --"`
 HLINT=$(EXEC) hlint
 AULA_SOURCES=-isrc -itests -idist/build/autogen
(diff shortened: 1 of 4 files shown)


Makefile
```diff
@@ -2,7 +2,7 @@ SHELL=/bin/bash
 EXEC=`test -d .stack-work/ && echo "stack exec --" || echo "cabal exec --"`
 HLINT=$(EXEC) hlint
 AULA_SOURCES=-isrc -itests -idist/build/autogen
-FULL_SOURCES=$(AULA_SOURCES) -i$(THENTOS_ROOT_PATH)/thentos-frontend-session/src/ $(THENTOS_ROOT_PATH)/thentos-frontend-session/tests/
+FULL_SOURCES=$(AULA_SOURCES) -i$(THENTOS_ROOT_PATH)/thentos-frontend-session/src/
```
fisx: why?

(it's not really that important because i am about to release thentos-frontend-session on hackage and this will go away anyway, but this is one of those changes that could use more documentation.)

andorp: `make test-repl` had a problem, it wanted to load a non-existing directory.

Issue #816, opened 2016-07-05
students should be able to delegate in jury phase, but never after voting phase.


fisx:

fisx:
not sure this is an issue.  what it means, and what we should check we do right:

- vote delegation on topic level is possible in phases refinement, jury, voting.
- not possible in result phase.

if that's what we already to, we can close this issue.

PR #847, opened 2016-07-12
No delegation in voting phase.

andorp:
Fixes: #816

### DIFF
--- a/src/Access.hs
+++ b/src/Access.hs
@@ -190,6 +190,7 @@ isStudent u = isJust $ find (has _Student) (u ^.. userRoles)
 
 delegationCapabilities :: User -> Maybe User -> Maybe IdeaSpace -> Maybe Phase -> [Capability]
 -- guards: no delegation in these situations
+delegationCapabilities _ _ _ (Just PhaseVoting{})              = []
 delegationCapabilities _ _ _ (Just PhaseResult)                = []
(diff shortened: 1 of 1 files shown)


src/Access.hs
```diff
@@ -190,6 +190,7 @@ isStudent u = isJust $ find (has _Student) (u ^.. userRoles)
 
 delegationCapabilities :: User -> Maybe User -> Maybe IdeaSpace -> Maybe Phase -> [Capability]
 -- guards: no delegation in these situations
+delegationCapabilities _ _ _ (Just PhaseVoting{})              = []
```
fisx: nope.  "never *after* voting phase" does not imply "never *in* voting phase".  we explicitly want to allow students to delegate during voting phase.

i'm not sure this issue was ever reproduced, or whether it was just a reminder to check we got it right.  i think we may have run into an independent (and already fixed) problem with the `CanDelegate*` confusion during the presentation.

fisx:
I'm closing this since it doesn't contain anything we want to keep.

src/Access.hs
andorp: Ok, in this case, the original issue has no sense. As the current implementation forbids to delegate your vote only in result phase. The issue reported the current behavior as the desired one. Probably the usage of godadmin produced this.

Issue #829, opened 2016-07-05
Make `PageStaticTermsOfUse` markdown-editable for admin.


fisx:
We will get some default terms of use, but we want the admin to be able to, say, change a three-strikes rule about posting offensive content to a one-strike rule in the middle of the school year.

Issue #199, opened 2016-03-12
Bootstrap admin user


andorp:
After changing to real persistent, there should be a tool to create admin user.

fisx:
this is a more general problem, and we should also offer some default terms of use.

this could be resolved by tweaking and renaming `genInitialTestDb`.

fisx:
(default school space also needs to be created there.)

fisx:
also, fill in some default terms of use.

PR #849, opened 2016-07-12
Configurable "terms of use" page.

andorp:
Fixes: #829


tests/Frontend/CoreSpec.hs
```diff
@@ -408,10 +408,11 @@ postToForm (FormTest g c check) = do
         page <- pick g
         ctx <- pick (arbFormPagePayloadCtx page)
         mpayload <- pick (arbFormPageInvalidPayload page)
-        forM_ mpayload
-            (\payload -> do
+        case mpayload of
+            Nothing -> liftIO $ pendingWith "*In*valid form input is not defined for this."
```
fisx: :+1:

src/Frontend/Page/Static.hs
```diff
@@ -48,8 +52,9 @@ instance Page PageStaticTermsOfUse where
...
-        h1_ "Nutzungsbedingungen"
-        p_  "Lorem ipsum"
-        p_  "Duis aumet"
-        p_  "Ut wisi"
+    toHtml p@(PageStaticTermsOfUse terms) = semanticDiv p .
+        div_ [class_ "text-markdown"] $
+            terms ^. html
+
+termsOfUse :: ActionM m => m PageStaticTermsOfUse
+termsOfUse = PageStaticTermsOfUse <$> query Persistent.termsOfUse
```
fisx: the "terms of use" page shouldn't be in this module any more.

fisx: more general remark: this module makes a good case against structuring the page modules along the sections "template" and "page".  we should have sections for pages, and those sections should contain first the page instances and then the handler.

fisx:
did you test this?  doesn't work for me yet:

- the initial text in the edit form is empty (instead of the admittedly boring template text we have in the static page)
- the page doesn't show the content i entered, but the static page.
- the static page html file hasn't been removed from `/static/'.

fisx:
"initial text" issue can be postponed to #199.

andorp:
> the page doesn't show the content i entered, but the static page.

I think it is a caching issue as I checked it and I couldn't reproduce it.

> the static page html file hasn't been removed from `/static/'

The original text was defined in a handler, there is no file to delete.

fisx:
rename `Frontend.Page.Static` to `Frontend.Page.Info` as discussed?

fisx:
thanks.  and sorry for the sloppy review...

PR #850, opened 2016-07-13
Get thentos from hackage

fisx:
this eliminates the dependency on a thentos git repo being locally installed.


fisx:
related: https://github.com/liqd/aula-docker/pull/3

i had to move the git submodule in the docker repo to this branch in order to the docker image to build.  we should probably, in that order:

- merge this branch;
- update the docker branch to carry aula-master again
- merge the docker branch

not sure though.  it should probably just work as is, too.

fisx:
i'll also fix the module name "thentos.frontend.session", and this:

https://github.com/liqd/thentos/blob/master/thentos-cookie-session/src/Thentos/Frontend/Session.hs#L80

thanks @andorp

andorp:
LGTM

PR #874, opened 2016-07-19
Extra info for delegation buttons

### LINKED ISSUE #870: show current delegatee underneath delegate button
If I have already delegated my vote, the person I delegated to should be shown as a text underneath the delegation button on the topic overview and in my own profile. The syntax should be: „Derzeit beauftragt: <username>“

fisx:
Fixes #870.  Also, lots of refactoring.

### DIFF
--- a/src/Frontend/Page/Topic.hs
+++ b/src/Frontend/Page/Topic.hs
@@ -225,9 +225,10 @@ viewTopicHeaderDiv now ctx topic tab delegation = do
                       if isNothing delegation
                             then "Stimme beauftragen"
                             else "Beauftragung ändern"
-                    unless (isNothing delegation) . p_ $
-                        "Derzeit beauftragt: " <>
-                        delegation ^. _Just . delegationFullTo . userLogin . unUserLogin . html
+                    unless (isNothing delegation) $ do
+                        let delegate = delegation ^?! _Just . delegationFullTo
+                        p_ . a_ [href_ $ U.viewUserProfile delegate] $
+                            "Derzeit beauftragt: " <> delegate ^. userLogin . unUserLogin . html
 
             case phase of
                 PhaseWildIdea{}   -> createIdeaButton
--- a/src/Frontend/Page/Topic.hs
+++ b/src/Frontend/Page/Topic.hs
@@ -2,7 +2,6 @@
 {-# LANGUAGE FlexibleContexts    #-}
 {-# LANGUAGE FlexibleInstances   #-}
 {-# LANGUAGE LambdaCase          #-}
-{-# LANGUAGE LambdaCase          #-}
 {-# LANGUAGE OverloadedStrings   #-}
 {-# LANGUAGE Rank2Types          #-}
 {-# LANGUAGE ScopedTypeVariables #-}
--- a/src/Frontend/Page/Topic.hs
+++ b/src/Frontend/Page/Topic.hs
@@ -2,6 +2,7 @@
 {-# LANGUAGE FlexibleContexts    #-}
 {-# LANGUAGE FlexibleInstances   #-}
 {-# LANGUAGE LambdaCase          #-}
+{-# LANGUAGE LambdaCase          #-}
 {-# LANGUAGE OverloadedStrings   #-}
 {-# LANGUAGE Rank2Types          #-}
 {-# LANGUAGE ScopedTypeVariables #-}
@@ -225,8 +226,7 @@ viewTopicHeaderDiv now ctx topic tab delegation = do
                       if isNothing delegation
                             then "Stimme beauftragen"
                             else "Beauftragung ändern"
-                    unless (isNothing delegation) $ do
-                        let delegate = delegation ^?! _Just . delegationFullTo
+                    forM_ delegation $ \(view delegationFullTo -> delegate) -> do
                         p_ . a_ [href_ $ U.viewUserProfile delegate] $
                             "Derzeit beauftragt: " <> delegate ^. userLogin . unUserLogin . html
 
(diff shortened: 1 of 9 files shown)


src/Frontend/Page/User.hs
```diff
@@ -356,7 +350,7 @@ createdIdeas userId ideasQuery = do
                     (mapM getIdeaStats
                      =<< filter visibleByCurrentUser
                          <$> findIdeasByUserId userId))
-        ds <- commonIdeaSpaceDelegations cUser user
+        ds <- commonIdeaSpaceDelegations cUser user >>= mapM delegationFull
```
andorp: This could be resource consuming load. Loading the same user repeatedly. A function could collect all the user ids, load them, reuse the references.

andorp: But I am ok with this solution for the time being.

fisx: thanks.  i did consider optimizing that, but since it's just a `Map` operation and doesn't go to disk, i am strongly in favor of optimizing this after profiling.

src/Frontend/Page/User.hs
```diff
@@ -238,52 +240,44 @@ commonIdeaSpaceDelegations delegatee delegate = do
...
-                then any (\d -> d ^. delegationScope == dscope &&
-                                d ^. delegationFrom == delegatee ^. _Id) delegations
-                else Delegation dscope (delegatee ^. _Id) (delegate ^. _Id)
-                     `elem`
-                     delegations
+delegationButtons :: Monad m => User -> User -> [DelegationFull] -> HtmlT m ()
+delegationButtons visiting visited delegations = do
+    let ownProfile = isOwnProfile visiting visited
+        isActiveDelegation = isJust . activeDelegation
+        activeDelegation dscope = listToMaybe . (`filter` delegations) $
```
andorp: `not . null` would be better.

src/Frontend/Page/User.hs
```diff
@@ -238,52 +240,44 @@ commonIdeaSpaceDelegations delegatee delegate = do
...
 -- For the own profile clicking on the delegation button, the delegate selection
 -- page is opened.
 --
 -- For other's profile, clicking on the delegation buttons mark the owner of
 -- the profile as the delegate of the current user.
-delegationButtons :: Monad m => User -> User -> [Delegation] -> HtmlT m ()
-delegationButtons delegatee delegate delegations = do
-    let ownProfile = isOwnProfile delegatee delegate
-        isActiveDelegation dscope =
-            if ownProfile
```
andorp: We should keep the `if` version of the `isActiveDelegation` function, it is more readable.

src/Frontend/Page/User.hs
```diff
@@ -238,52 +240,44 @@ commonIdeaSpaceDelegations delegatee delegate = do
...
+                butGet (U.createDelegation dscope)
+                    ("Deine Beauftragung für " <> uilabel ispace)
+            (False, True) ->
+                butPost (U.withdrawDelegationOnIdeaSpace visited ispace)
+                    ("Beauftragung für " <> uilabel ispace <> " entziehen")
+            (False, False) ->
+                butPost (U.delegateVoteOnIdeaSpace visited ispace)
+                    ("Für " <> uilabel ispace <> " beauftragen")
+        br_ []
+        (`mapM_` activeDelegation dscope) $ \(DelegationFull _ _ delegate) ->
```
andorp: This is the `forM_` please use that.

src/Frontend/Page/User.hs
fisx: I disagree.  the `Maybe` makes it explicit that we expect the list to be no longer than 1.

src/Frontend/Page/Topic.hs
```diff
@@ -216,14 +217,18 @@ viewTopicHeaderDiv now ctx topic tab delegation = do
...
-                        ] $ do
+                delegateVoteButton = when (any (`elem` caps) [CanDelegateInClass, CanDelegateInSchool]) $ do
+                    a_ [ class_ "btn-cta heroic-cta m-large"
+                       , href_ $ U.createDelegation (DScopeTopicId topicId)
+                       ] $ do
                       i_ [class_ "icon-bullhorn"] nil
                       if isNothing delegation
                             then "Stimme beauftragen"
                             else "Beauftragung ändern"
+                    unless (isNothing delegation) $ do
```
andorp: `when (isJust delegation)`

andorp: or `mapM_`

src/Frontend/Page/User.hs
fisx: I disagree.  We need both `isActiveDelegation` and `activeDelegation`, and I also think readability improved with my change.

src/Frontend/Page/User.hs
fisx: will do, thanks!

src/Frontend/Page/User.hs
andorp: I agree. Please use `find` instead of `listToMaybe . filter pred`. Shouldn't we add this as an hlint rule?

src/Frontend/Page/User.hs
andorp: I have no strong opinion here, probably it is matter of taste and personal style. Leave as it is.

src/Frontend/Page/User.hs
fisx: i'd be interested in watching you write an hlint rule, sure.  (-:

will fix this time.

src/Frontend/Page/Topic.hs
```diff
@@ -2,6 +2,7 @@
 {-# LANGUAGE FlexibleContexts    #-}
 {-# LANGUAGE FlexibleInstances   #-}
 {-# LANGUAGE LambdaCase          #-}
+{-# LANGUAGE LambdaCase          #-}
```
andorp: Duplicated langauge pragma.

andorp:
LGTM

Issue #846, opened 2016-07-12
Remove user login informations from the login page.


andorp:

fisx:
(make this depend on a config-file flag or config constant somewhere, say, `_devMode`.  use same flag for #845.)

andorp:
Good idea.

PR #889, opened 2016-07-19
Dev mode

andorp:
Fixes: #845 
Fixes: #846

### DIFF
--- a/src/Frontend/Core.hs
+++ b/src/Frontend/Core.hs
@@ -642,8 +642,8 @@ cancelButton p = a_ [class_ "btn", href_ $ redirectOf p ()] "Abbrechen"
 -- the entire sequence of requests the client presents the same cookie, not only over those
 -- sub-sequences where the user is logged-in.)
 data Frame body
-    = Frame { _frameUser :: User, _frameBody :: body, _frameMessages :: [StatusMessage] }
-    | PublicFrame               { _frameBody :: body, _frameMessages :: [StatusMessage] }
+    = Frame { _frameUser :: User, _frameBody :: body, _frameMessages :: [StatusMessage], _frameDevMode :: Bool }
+    | PublicFrame               { _frameBody :: body, _frameMessages :: [StatusMessage], _frameDevMode :: Bool }
   deriving (Show, Read, Functor)
 
 -- | Check authorization of a page action.  Returns a new action.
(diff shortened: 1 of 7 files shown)


src/Frontend/Core.hs
```diff
@@ -642,8 +642,8 @@ cancelButton p = a_ [class_ "btn", href_ $ redirectOf p ()] "Abbrechen"
 -- the entire sequence of requests the client presents the same cookie, not only over those
 -- sub-sequences where the user is logged-in.)
 data Frame body
-    = Frame { _frameUser :: User, _frameBody :: body, _frameMessages :: [StatusMessage] }
-    | PublicFrame               { _frameBody :: body, _frameMessages :: [StatusMessage] }
+    = Frame { _frameUser :: User, _frameBody :: body, _frameMessages :: [StatusMessage], _frameDevMode :: Bool }
+    | PublicFrame               { _frameBody :: body, _frameMessages :: [StatusMessage], _frameDevMode :: Bool }
```
fisx: It's looking more and more like the `_frameUser` should be a `Maybe` and there should be only one constructor, but let's not try that change here.

src/Frontend/Core.hs
```diff
@@ -850,20 +850,21 @@ renderStatusMessage :: (Monad m) => StatusMessage -> HtmlT m ()
...
             span_ [class_ "main-footer-blurb"] $ do
                 "Made with \x2665 by Liqd"
-                replicateM_ 5 $ toHtmlRaw nbsp
-                toHtml Config.releaseVersion
-                replicateM_ 5 $ toHtmlRaw nbsp
-                a_ [Lucid.onclick_ "createPageSample()"]
-                    "[create page sample]"  -- see 'Frontend.createPageSamples" for an explanation.
+                when devmode $ do
+                    replicateM_ 5 $ toHtmlRaw nbsp
+                    toHtml Config.releaseVersion
```
fisx: i think the release should be in the production page as well.

fisx: i just saw that the ticket says otherwise.  i disagree, but ok, never mind.

src/Frontend/Core.hs
andorp: I would say we could introduce a `FrameKind` type which is isomorph to the Maybe type.

Issue #694, opened 2016-06-21
do not show unapplicable tabs in user profile


fisx:
for example, principles have no delegation tabs or created ideas.  should be controlled via capabilities.

PR #890, opened 2016-07-20
Do not render inappropriate tabs on user profile page.

andorp:
Fixes: #694

### DIFF
--- a/src/Frontend/Page/User.hs
+++ b/src/Frontend/Page/User.hs
@@ -47,6 +47,9 @@ import qualified Text.Digestive.Lucid.Html5 as DF
 
 -- * misc
 
+canDelegate :: CapCtx -> Bool
+canDelegate ctx = CanDelegate `elem` capabilities ctx
+
 profileContext :: (ActionPersist m, ActionUserHandler m) => User -> m CapCtx
 profileContext user = set capCtxUserProfile (Just user)
                     . set capCtxDelegateTo  (Just user)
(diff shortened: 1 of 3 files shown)


src/Types/Util.hs
```diff
@@ -201,6 +201,9 @@ isAdmin = (`hasRole` Admin)
 isStudent :: User -> Bool
 isStudent u = isJust $ find (has _Student) (u ^.. userRoles)
 
+canCreateIdeas :: User -> Bool
+canCreateIdeas = isStudent
```
fisx: `|| isModerator`.

src/Types/Util.hs
```diff
@@ -201,6 +201,9 @@ isAdmin = (`hasRole` Admin)
 isStudent :: User -> Bool
 isStudent u = isJust $ find (has _Student) (u ^.. userRoles)
```
fisx: `warn = isJust . find ==> any`

(not related to this PR, but we might as well "fix" it instead of keeping track of it.)

src/Frontend/Page/User.hs
```diff
@@ -296,15 +299,18 @@ data UserProfileTab
...
 -- could be factored out.
-userProfileTab :: Monad m => UserProfileTab -> User -> HtmlT m ()
-userProfileTab activeTab user = do
+userProfileTab :: Monad m => CapCtx -> UserProfileTab -> User -> HtmlT m ()
+userProfileTab ctx activeTab user = when (canCreateIdeas' || canDelegate') $ do
     div_ [class_ "heroic-tabs is-responsive"] $ allTabs False
     select_ [class_ "heroic-tabs-dropdown", onchange_ "window.location = this.value"] $ allTabs True
   where
+    canCreateIdeas' = canCreateIdeas user
+    canDelegate'    = canDelegate ctx
```
fisx: i am generally against these trivial name introductions, but ok.

andorp: We use those in 5 different places, I would like to keep them as it is.

src/Types/Util.hs
```diff
@@ -199,10 +198,10 @@ isAdmin :: User -> Bool
 isAdmin = (`hasRole` Admin)
 
 isStudent :: User -> Bool
-isStudent u = isJust $ find (has _Student) (u ^.. userRoles)
+isStudent u = any (has _Student) (u ^.. userRoles)
 
 canCreateIdeas :: User -> Bool
-canCreateIdeas = isStudent
+canCreateIdeas u = any ((||) <$> has _Student <*> has _Moderator) (u ^.. userRoles)
```
fisx: :+1:  (-:

Issue #895, opened 2016-07-20
Nicht gewonnen icon appears always on view idea page.


andorp:
![2016-07-20-160638_1600x900_scrot](https://cloud.githubusercontent.com/assets/3465327/16989339/20e29e3e-4e94-11e6-818e-8b7813e212c5.png)

Issue #786, opened 2016-06-30
make 'like' revokable


fisx:
there are a few FIXMEs about this in the code, too.

PR #896, opened 2016-07-20
Revoke like

andorp:
Fixes: #786
Fixes: #895

### DIFF
--- a/src/Persistent/Api.hs
+++ b/src/Persistent/Api.hs
@@ -65,7 +65,7 @@ $(makeAcidic ''AulaData
     , 'addIdeaVoteResult
     , 'revokeWinnerStatus
     , 'addLikeToIdea
-    , 'dislikeIdea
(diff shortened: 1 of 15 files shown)


src/Persistent/Api.hs
```diff
@@ -65,6 +65,7 @@ $(makeAcidic ''AulaData
     , 'addIdeaVoteResult
     , 'revokeWinnerStatus
     , 'addLikeToIdea
+    , 'removeLikeFromIdea
```
fisx: Why is it not `OnIdea` any more here?  Just to resolve the name conflict?  Or because it feels the same?

fisx: (it's not relevant any more in the later commits.)

fisx:
LGTM!

I only edited some comments, so I won't wait for travis.  Please take a look if you disagree with any of my changes.

PR #898, opened 2016-07-20
Fix dropdown padding on user menu

localgrr:
Fixes a problem I made with the main dropdown padding being all weird


fisx:
LGTM, thanks!

Issue #708, opened 2016-06-22
caching and cookie management


fisx:
1. i experienced weird behavior that could only be fixed by deleting all cookies, and i turned out to have collected a lot of them.
1. we have not thought about caching at all so far.  there may be some performance to be gained, especially if we use some hashing-naming-scheme for the `/static` content, but there may also be some hidden issues with stale pages that are delivered from cache, but shouldn't.

fisx:
related, or duplicate of: #283.

fisx:
> Link „Profil anzeigen“ in user menu in main nav redirects to idea space overview
> https://aula-stage1.liqd.net/user/18/ideas

moritz, can you clear all you aula-related cookies and try to reproduce this?
i suspect it won't work.  this is a known bug.
Moritz: Link worked as expected again, when I tested it this morning

fisx:
this may be resolved by now.  not sure, though.

fisx:
lots of packages on hackage:

- http://hackage.haskell.org/package/wai-middleware-preprocessor
    more for compiling purescript dynamically on delivery

- http://hackage.haskell.org/package/wai-middleware-etag
    *almost* what we want, but doesn't watch file changes

- http://hackage.haskell.org/package/wai-middleware-static-caching
    probably good for inspiration, but doesn't help with cache busting.

- http://hackage.haskell.org/package/wai-middleware-caching
    only server-side

- http://hackage.haskell.org/package/wai-static-cache
    only server-side

- http://hackage.haskell.org/package/wai-middleware-cache
    only server-side

nothing that does browser cache busting.

fisx:
goal:

- cache everything under /static
- do not cache anything else
- every reference to a cached resource should end with `[?|&]last_changed=<epoch_time>`

we could write a more sophisticated (and much slower) middleware that parses all of html, css, js, and whatever else i forgot, but...

fisx:
This has been resolved in #735.

(I opened #921 to keep the last two comments in scope.)

Issue #707, opened 2016-06-22
race condition in image upload


fisx:
The new image sometime does not appear after the upload, but on refresh it appears.

fisx:
related: #436.

np:
I'm not sure if this is a form of caching or not. If this is a caching issue we can add a query parameter with the timestamp or the hash of image.

np:
We can't reproduce it anymore.

fisx:
irc::

> 15:32 < npou> damn, I got the race condition just now while uploading my profile pic

fisx:
It is *possible* that this has been resolved in #920, but in order to prove that we should:

1. remove the `disableCaching` middleware.
1. reproduce the issue in a test case.
1. add `disableCaching` again and confirm that the issue is gone.

andorp:
We can not reproduce this with the test suite. There is no hard evidence how this happen. We reenable caching for the service to save energy. @fisx has an idea how to invalidate cache on the client side. Closing for now.

fisx:
related?:

- http://stackoverflow.com/questions/174348/will-web-browsers-cache-content-over-https
- http://stackoverflow.com/questions/8155064/how-to-programmatically-empty-browser-cache
- http://stackoverflow.com/questions/1011605/clear-the-cache-in-javascript

PR #920, opened 2016-07-26
Browser caching

fisx:
Related #708.

Possibly related: #707.


fisx:
so i think this may get rid of #707, BUT:

- in order to be sure we need to have a test case to reproduce it.
- we should test what that does to mobile users with cheap connections.  that's a big chunk of our target audience.

andorp:
LGTM

PR #922, opened 2016-07-27
Bootstrap aula env

andorp:
Fixes: #199

### DIFF
--- a/src/Frontend.hs
+++ b/src/Frontend.hs
@@ -73,15 +73,6 @@ runFrontend cfg = do
     let logMsg = log ^. msgDaemonSend
     withPersist logMsg cfg (runFrontend' cfg logMsg)
 
-runBoostrap :: Config -> Action () -> IO ()
-runBoostrap cfg action = do
-    log <- logDaemon (cfg ^. logging)
-    void $ log ^. start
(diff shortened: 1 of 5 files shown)


.gitignore
```diff
@@ -23,3 +23,5 @@ aula.yaml*
 /static/avatars/*.png
 /hpc
 cookie-jar
+aulaEventLog.json
```
fisx: that's probably related to something you did to your private aula.yaml?  i would like to keep that out of the repo, but i know i'm not consistent that way.  feel free to ignore me.  (-:

src/Frontend.hs
```diff
@@ -73,6 +73,15 @@ runFrontend cfg = do
     let logMsg = log ^. msgDaemonSend
     withPersist logMsg cfg (runFrontend' cfg logMsg)
 
+runBoostrap :: Config -> Action () -> IO ()
+runBoostrap cfg action = do
```
fisx: Why is this function in `src/Frontend.hs`?  Shouldn't it be in the executable module?

fisx:
LGTM!  Heh, your command line parser is so small I couldn't find it for a moment.  :-)

If you have tested manually and answered my above comment (ideally as a haddocks line), you can merge this.

PR #930, opened 2016-07-29
Fix docker build script.

fisx:
I'm basing this change on my half-baked bash theories, but I think this was a bug.  It may never have been triggered so far, though.

### DIFF
--- a/.travis/docker-build.sh
+++ b/.travis/docker-build.sh
@@ -28,10 +28,13 @@ done
 
 stack install --fast --pedantic --allow-different-user \
     --test --test-arguments "--qc-max-success=${QC_MAX_SUCCESS} --qc-max-size=${QC_MAX_SIZE}" \
-    --coverage \
     aula
 
 # FIXME: Coveralls coverage
+#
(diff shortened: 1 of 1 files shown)


.travis/docker-build.sh
```diff
@@ -9,29 +9,32 @@ cd /liqd/aula
...
+
+        *)
+            ;;
+    esac
     shift
-    ;;
-
-    --qc-max-size)
-    QC_MAX_SIZE="$2"
-    shift
```
andorp: The double shift is necessary, to get rid of the parameter and its value in the same round.

.travis/docker-build.sh
```diff
@@ -9,29 +9,32 @@ cd /liqd/aula
 QC_MAX_SUCCESS="100"
 QC_MAX_SIZE="30"
 
-while [[ $# > 1 ]]
-do
-key=$1
-case $key in
-    --qc-max-success)
-    QC_MAX_SUCCESS="$2"
+while [[ $# > 1 ]]; do
```
andorp: I like the indentation, please keep it.

PR #933, opened 2016-07-30
Deployment tweaks

fisx:


fisx:
as discussed off-line: aula-init-state should move data files to `$AULA_ROOT_PATH` or, if n/a, to `.`.

exec/InitState.hs
```diff
@@ -68,6 +71,28 @@ runBoostrap cfg action = do
         let runAction = mkRunAction (ActionEnv rp cfg logMsg)
         unNat (exceptToFail . runAction) action
 
+copyDir :: FilePath -> FilePath -> IO ()
+copyDir from to = unfoldTreeM dirTree (Left "") >>= copyTree
+  where
+    usefull = not . (`elem` [".", ".."])
```
fisx: it's spelled `useful`.

fisx: (I would also not have written these function to begin with, but just used `System.Process` to call the shell.  Copying files is what shells are reasonably good at.)

fisx: in fact, i am going to make this change, because now after reading and tweaking these functions for a while i get this:

```
aula-init-state: .../aula/.cabal-sandbox/share/x86_64-linux-ghc-7.10.2/aula-0.22.0/README.md: getDirectoryContents: inappropriate type (Not a directory)
```

andorp:
LGTM

exec/InitState.hs
andorp: `copyDir` assumes a directory to copy.

andorp:
LGTM

PR #938, opened 2016-08-01
Use proper default profile image.

### LINKED ISSUE #928: need default avatar image
we are currently using one of the category images, but that's not suitable for production.  i'd be perfectly fine with a gray circle, but we need *something*.

fisx:
Fixes #928.


fisx:
creating users in the prozessverwaltung doesn't generate a default profile image.  i think that's because we use the `updateAvatarByCopy` version in `Action.Implementation`, not the one in `DemoData`.  I reverted 82a154a, but that made no difference.  (somebody should probably try again anyway, i'm a bit tired today.)

PR #941, opened 2016-08-03
Bump thentos-cookie-session.

### LINKED ISSUE #932: session timeout
- check that session timeout does not kick the user out *unless she was idle* for the timeout period.
- check what's with `extendClearanceOnSessionToken`.

this may not be an issue, but it's a thing worth double-checking.

fisx:
Fixes #932.

(well, not really.  clarifies that there wasn't anything to fix.)


andorp:
LGTM

PR #942, opened 2016-08-03
Logging and threads

fisx:


fisx:
It would be good if this could get a thorough review.  I think I got it right, but if not it could be ugly to debug in production.

src/Daemon.hs
```diff
@@ -67,13 +70,28 @@ msgDaemon
...
+msgDaemon logger name computation handleException threadSafe = do
     chan <- newTChanIO
+    alreadyRunningRef :: MVar (Maybe ThreadId) <- newMVar Nothing
 
     let sendMsg = atomically . writeTChan chan
-        loop = forkIO . forever $ run `catch` handle
+        loop = do
+            alreadyRunning <- takeMVar alreadyRunningRef
+            case (threadSafe, alreadyRunning) of
+                (False, Just tid) -> do
```
andorp: Isn't it reversed? When `threadSafe` is `True` we should check the guard.

src/Daemon.hs
```diff
@@ -125,10 +143,27 @@ timeoutDaemon' logger name delay computation =
...
--- | Create a log daemon
+-- | Create a log daemon.  This is both responsible for the moderator's event log (file location
+-- given in Config.hs) and system log messages (stderr).
+--
+-- NOTE: 'logDaemon' can only be called once, and only have one thread.  Multiple calls to 'start'
+-- will return the same 'ThreadId', multiple calls to 'logDaemon' will crash.  (What would go wrong?
+-- If too many `logDaemon`s are running concurrently, they might undo each others moderator event
+-- log entries, or enter an inconsistent state.  also, stderr could be garbled if two lines are
+-- printed concurrently.)
 logDaemon :: LogConfig -> IO (MsgDaemon LogEntry)
```
andorp: This could be resolved in the future separating moderator logs from system logs with two different types and log daemons.

src/Daemon.hs
fisx: If a daemon is thread safe, then we can behave as before.  If it is *not*, we make it so here.  See call in `logDaemon` below.

andorp:
LGTM

PR #947, opened 2016-08-08
Wai metrics

andorp:


fisx:
cool.

i would like to add an option to not have ekg running.  what would be the best way?  what about making a config sub-type combining host and port, and making the _monitoring entry optional?

i'm happy to do this if you won't have time today, but i'd like to know what you think.

otherwise LGTM.

andorp:
I think it is not a big change.

fisx:
can be merged if you and travis agree.

thanks!

andorp:
LGTM

PR #951, opened 2016-08-09
Aula (login) metrics

andorp:


src/Frontend.hs
```diff
@@ -57,24 +58,26 @@ runFrontend :: Config -> IO ()
 runFrontend cfg = do
     log <- logDaemon (cfg ^. logging)
     void $ log ^. start
-    waiMetrics <- startEKG cfg
-    runFrontendWithLogger cfg (log ^. msgDaemonSend) waiMetrics
+    metrics <- startEKG cfg
+    runFrontendWithLogger cfg (log ^. msgDaemonSend) metrics
 
-startEKG :: Config -> IO (Maybe EKG.WaiMetrics)
+startEKG :: Config -> IO (Maybe (EKG.WaiMetrics, AulaMetrics))
```
fisx: would it make sense to add `WaiMetrics` as a field to `AulaMetrics`?

fisx: ah, that would have the disadvantage that it's being carried around inside the `Action`.

fisx:
LGTM!

src/Frontend.hs
andorp: And never used, or used in an inappropriate way.

PR #955, opened 2016-08-09
404 styling and encoding.

andorp:
Fixes: #146 
Fixe: #208

### DIFF
--- a/src/Frontend/Middleware.hs
+++ b/src/Frontend/Middleware.hs
@@ -28,8 +28,8 @@ import Frontend.Prelude
 import Frontend.Testing
 
 
-catch404 :: Middleware
-catch404 app req cont = app req $ \resp -> cont $ f resp
+catch404 :: Bool -> Middleware
+catch404 devMode app req cont = app req $ \resp -> cont $ f resp
   where
     f :: Response -> Response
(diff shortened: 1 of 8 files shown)


src/Frontend/Middleware.hs
```diff
@@ -37,11 +37,14 @@ catch404 devMode app req cont = app req $ \resp -> cont $ f resp
         else responseBuilder status headers builder
       where
         status  = responseStatus resp
-        headers = responseHeaders resp
         builder = Builder.byteString . cs
                 . (`runReader` whereToGetTheLangValue) . renderTextT . toHtml
                 $ PublicFrame Page404 [] devMode
 
+        setContentType ("Content-Type", "text/plain") = ("Content-Type", "text/html;charset=utf-8")
```
fisx: misleading name (even if local).  you are setting the content type *encoding*, not the *type*.  i'll change it to something.

fisx: ah, type also changes to html, sorry.  alsmost correct then, but `repair` is still better than `set` i think.

fisx:
LGTM!

can be merged if you and travis agree.

andorp:
LGTM

Issue #953, opened 2016-08-09
prune acid-state archive


fisx:
We want a daemon that goes through `state/AcidState/Archive/` and removes all but the most recent `N` snapshots.  `N` should be configurable via a field like `Config.acidStateArchiveLength`.

This is similar to, but also quite different from `logrotate(1)`.  The latter works on a single file and creates a sequence of copies of that to rotate them out of existence; we need something that works on a growing sequence of files of which old ones must be deleted.

fisx:
@andorp if you have other things to do please unassign yourself.

fisx:
pseudo-code from off-line discussion:

```haskell
pruneFiles :: FilePath -> Maybe FilePath -> Int -> IO ()
pruneFiles dir mprefix keepnum = do
  ...


_cleanupDeamon:
  _period :: Timespan
  _cleanupRules:
    - _cleanupRule:
        _directory: ...
        _stem: ...
        _keenum :: Int
    - _cleanupRule:
        _directory: ...
        _stem: ...
        _keenum :: Int
```

PR #961, opened 2016-08-10
Cleanup daemon

andorp:
Fixes: #953

### DIFF
--- a/src/Config.hs
+++ b/src/Config.hs
@@ -215,17 +215,13 @@ defaultLogConfig = LogConfig
     , _eventLogPath = "./aulaEventLog.json"
     }
 
-defaultCleanUpRule :: CleanUpRule
-defaultCleanUpRule = CleanUpRule
-    { _cleanUpDirectory = "./state/AulaData/Archive"
-    , _cleanUpPrefix    = "events"
-    , _cleanUpKeepnum   = 5
-    }
-
 defaulCleanUpConfig :: CleanUpConfig
 defaulCleanUpConfig = CleanUpConfig
     { _cleanUpInterval = TimespanMins 45
-    , _cleanUpRules    = [defaultCleanUpRule]
+    , _cleanUpRules    =
+        [ CleanUpRule "./state/AulaData/Archive" "events" 0
+        , CleanUpRule "./state/AulaData/Archive" "checkpoints" 10
+        ]
     }
 
 defaultConfig :: Config
--- a/src/Config.hs
+++ b/src/Config.hs
@@ -7,10 +7,17 @@
 {-# OPTIONS_GHC -Werror -Wall -fno-warn-orphans #-}
 
 module Config
-    ( Config(Config), ListenerConfig(..), SmtpConfig(SmtpConfig), LogConfig(..)
+    ( Config(Config)
+    , ListenerConfig(..)
+    , SmtpConfig(SmtpConfig)
+    , LogConfig(..)
+    , CleanUpConfig(..)
+    , CleanUpRule(..)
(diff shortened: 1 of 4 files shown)


src/Config.hs
```diff
@@ -124,6 +137,24 @@ data ListenerConfig = ListenerConfig
 
 makeLenses ''ListenerConfig
 
+data CleanUpRule = CleanUpRule
+    { _cleanUpDirectory :: FilePath
+    , _cleanUpPrefix    :: String
```
fisx: Is this also `FilePath`?  Egal...  (-:

src/Daemon.hs
```diff
@@ -170,3 +176,18 @@ unsafeLogDaemon cfg = do
...
                                             hPutStrLn stderr (cshow now <> " [" <> cshow level <> "] " <> cs msg)
     logMsg (LogEntryForModerator ev) = LBS.appendFile (cfg ^. eventLogPath) $ Aeson.encode ev <> cs "\n"
+
+cleanUpDaemon :: SystemLogger -> CleanUpConfig -> TimeoutDaemon
+cleanUpDaemon logger CleanUpConfig{..} =
+    timeoutDaemon' logger "CLEANUP" _cleanUpInterval (mapM_ (cleanUpDir logger) _cleanUpRules)
+
+cleanUpDir :: SystemLogger -> CleanUpRule -> IO ()
+cleanUpDir logger CleanUpRule{..} = do
+    let basePath = _cleanUpDirectory
```
fisx: o_O

fisx: (that was slightly over-reacting...  :-)

src/Daemon.hs
```diff
@@ -170,3 +176,18 @@ unsafeLogDaemon cfg = do
...
+    files <- filterM doesFileExist =<< (basePath </>) <$$> filter (_cleanUpPrefix `isPrefixOf`)
+             <$> getDirectoryContents basePath
+    mods  <- mapM getModificationTime files
+    let files' = drop _cleanUpKeepnum $ fst <$> reverseSortOn (to snd) (files `zip` mods)
+    forM_ files' $ \file -> do
+        logger . LogEntry INFO . cs $ "deleting file " <> file
```
fisx: `DEBUG`?  not sure, but i have a slight preference to `DEBUG`.

fisx: ok, `INFO` it is.

src/Types/Prelude.hs
```diff
@@ -115,6 +115,9 @@ infixr 9 <...>
 sortOn :: Ord b => Getter a b -> [a] -> [a]
 sortOn l = sortBy (compare `on` view l)
 
+reverseSortOn :: Ord b => Getter a b -> [a] -> [a]
+reverseSortOn l = sortBy ((flip compare) `on` view l)
```
fisx: This is the same as `downSortOn`, except it doesn't crash ghc.  I would like to use `downSortOn` instead of this new function and rewrite `downSortOn` accordingly if that fixes the problem.

We should confirm that the ghc crash was caused by that function, which I find surprising.

fisx:
Few comments that need a response, otherwise LGTM!

fisx:
i also think the default rules should guarantee constant memory use in `Archive`.  perhaps 10 for snapshots, 0 for transaction logs?

fisx:
LGTM.

so it was a stack / so-file consistency issue?

anyway, good to merge IMO.  thanks!

andorp:
I think the difference comes from the `--fast` option for stack install.

PR #965, opened 2016-08-11
smtp config and test email.

fisx:
- make default recipient in config mandatory
- send better test email on server start.


src/Action/Smtp.hs
```diff
@@ -25,14 +25,15 @@ module Action.Smtp
...
-import AulaPrelude
 
 import qualified Data.ByteString as SB
 
+import AulaPrelude
 import Config
 import Logger
 import Types
 
+
```
andorp: :)

andorp:
LGTM

Issue #960, opened 2016-08-10
5xx errors are not always captured.


fisx:
One error that is not caught is "Ill-formed CSRF token" from `thentos-cookie-session` package.  It is triggered in #959.

fisx:
note: this is not about *logging* the error (which may also be an issue), but about *styling* it.

come to think of it, i think it's both.

PR #970, opened 2016-08-11
Log action exceptions

andorp:
Fixes: #960


src/Action.hs
```diff
@@ -239,6 +238,19 @@ data ActionExcept
 
 makePrisms ''ActionExcept
 
+instance LogMessage ActionExcept where
+    logLevel = \case
+        ActionExcept e
+          | errHTTPCode e < 500 -> NOLOG
```
andorp: This should be `DEBUG`.

fisx:
LGTM.

PR #975, opened 2016-08-17
Allow creation of overlapping classes.

fisx:


fisx:
this won't do yet.

we (have to) allow for equal names of different students, and if two students are created without login name, the created login names will be mutated until there is no conflict.

this means that in the case of classes, we have no obvious way of knowing whether a student has been created before for another class, or whether she simply has a similar name.

POSSIBLE FIX: change the `create class` button to `create students`, and allow a field that contains a comma-separated list of class names that the student participates in.

PR #979, opened 2016-08-24
Logfiles

fisx:
logging to stdout has its disadvantages, so we stop doing that here.

### DIFF
--- a/src/Config.hs
+++ b/src/Config.hs
@@ -345,9 +345,8 @@ logmotd cfg wd = do
     now <- getCurrentTime
     name <- getProgName
     aulaLog (cfg ^. logging) . LogEntry INFO . ST.unlines $
-        [ ""
-        , cshow now
-        , "this is " <> cs name <> "!"
+        [ "starting " <> cs name
+        , ""
         , "\nrelease:"
         , cs Config.releaseVersion
         , "\nroot path:"
--- a/src/Config.hs
+++ b/src/Config.hs
@@ -253,8 +253,10 @@ sanitize = exposedUrl %~ (\u -> if "/" `isSuffixOf` u then init u else u)
 data WarnMissing = DontWarnMissing | WarnMissing | CrashMissing
   deriving (Eq, Show)
 
-readConfig :: (LogEntry -> IO ()) -> WarnMissing -> IO Config
-readConfig logger warnMissing = sanitize <$> (configFilePath >>= maybe (errr msgAulaPathNotSet >> dflt) decodeFileDflt)
+-- | In case of @WarnMissing :: WarnMissing@, print the warning to stderr.  (We don't have logging
+-- configured yet.)
+readConfig :: WarnMissing -> IO Config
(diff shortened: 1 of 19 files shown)


fisx:
well, i'm almost there i guess.  you can start reading if you want to, i'll post another Errata.

src/Logger.hs
```diff
@@ -20,7 +20,5 @@ data LogEntry =
   | LogEntryForModerator EventLogItemCold
   deriving (Eq, Show)
 
-type SendLogMsg = LogEntry -> IO ()
```
andorp: I would like to keep this type synonym. There is similar namings constructions in the persistent module.

src/Config.hs
```diff
@@ -333,3 +336,22 @@ checkPathExistsAndIsEmpty path = do
     isempty <- null <$> getDirectoryContentsNoDots path
     unless isempty . throwIO . ErrorCall $
         show path <> " does not exist, is not a directory, or is not empty."
+
+
+-- * motd
```
andorp: What does `motd` stand for?

src/Logger.hs
fisx: well, that just means those constructions should be eliminated as well...  :-)

hm....  i'm thinking...

fisx: 1. the type is as long as the synonym.
1. it contains the structure that tells what it means, instead of a name that we cannot trust and that is more ambiguous to begin with.
1. if we have the synonym, we risk not using it and mixing the type with the synonym in different places in the code, which makes reading and refactoring harder.

what are your reasons for keeping it?

src/Config.hs
fisx: see last Errata.

src/Logger.hs
fisx: one more reason: it has become much less widely used to other code improvements.

andorp: * Documentary types, as we except only logging here.
 * More safe in case of refactor.
 * In my opinion converting type synonyms to newtype are better than removing it.

src/Config.hs
```diff
@@ -340,9 +340,9 @@ checkPathExistsAndIsEmpty path = do
 
 -- * motd
 
+-- | Log the message (motto) of the day (like /etc/motd).
 logmotd :: Config -> FilePath -> IO ()
```
andorp: Should we call it as `logConfig`?

fisx:
I still think 17e797a makes things worse, but if you like it we can keep it.

src/Config.hs
fisx: config is only part of it, and `logConfig` is much less informative than `motd` (well ok, at least if you know unix long enough for this to be a word.)  i want to keep the name.

src/Config.hs
andorp: Ok, let's keep it.

andorp:
Ok, let's revert 17e797a , if we need more complex logging we can create a ADT. Please revert and merge.

fisx:
changed my mind again about 17e797a.  sorry!  (-:

PR #984, opened 2016-09-01
csv upload tweaks etc.

fisx:
- text encoding issues.
- if csv files contain empty lines, those should not trigger bogus user creation.
- ~~tweak avatar image browser caching.~~
- tweak avatar image size selection.
- devops niceties.

### DIFF
--- a/.travis/docker-build.sh
+++ b/.travis/docker-build.sh
@@ -3,6 +3,8 @@
 /liqd/aula/.travis/docker-link-stack-workdir.sh
 /liqd/aula/docker/selenium.sh restart
 
+apt-get install libicu-dev --yes
+
 # Change to the source directory which is attached as docker volume
 cd /liqd/aula
 
--- a/aula.cabal
+++ b/aula.cabal
@@ -184,6 +184,7 @@ library
     , string-conversions
     , template-haskell
     , text
+    , text-icu
     , time
     , transformers
     , vector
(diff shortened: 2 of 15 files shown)


fisx:
Caching of avatar breaks test suite.  We could play with the 'applicationCacheEnabled' capability in hs-webdriver, but I guess the failure means something.  Some perfromance problems (due to bad wlan) have been reported in practice, so I'm not sure how to proceed here.  I'll ask poldi!  Let's see how the performance turns out in practice.

.travis/docker-build.sh
```diff
@@ -3,6 +3,8 @@
 /liqd/aula/.travis/docker-link-stack-workdir.sh
 /liqd/aula/docker/selenium.sh restart
 
+apt-get install libicu-dev --yes
```
andorp: This should be part of the aula-docker dockerfile. But it is ok for now.

andorp:
unsafePermormIO should be avoided, we can fix that in a separate PR. LGTM

.travis/docker-build.sh
fisx: I agree, of course; it was just easier to do it here.

fisx:
I'm curious: I half-reasoned out that in this case `unsafePerformIO` is not a problem.  Of course it's always better to not have to do the reasoning, but have you thought of anything specific that can go wrong here if we don't refactor it away?

Anyway, thanks for the review, well-spotted.  I'll merge, then.

Issue #991, opened 2016-10-17
user profile tabs: "delegateTo" and "delegateFrom" labels are flipped.


fisx:

andorp:
I checked this. for me seems the tabs are right...

andorp:
Probably the URL is missleading...
![2016-10-18-125635_1600x900_scrot](https://cloud.githubusercontent.com/assets/3465327/19475077/91dd6d8c-9532-11e6-9c0f-9305d8a23e65.png)

andorp:
We can ask poly to provide an URL if the issue persist on their side.

fisx:
i'm also confused, so we should probably add a comment to the source like this:

- delegatees give their vote to me, so i vote for them: "für wen stimme ich ab?" ("for whom do i vote?")
- delegates get their vote from me, so they vote for me: "wer stimmt für mich ab?" ("who votes for me?")

PR #993, opened 2016-10-18
Flip delegation tab names on the user profile page.

andorp:
Fixes: #991

### DIFF
--- a/src/Data/Delegation.hs
+++ b/src/Data/Delegation.hs
@@ -58,9 +58,13 @@ type S = DScope
 
 -- * types
 
+-- | Delegatees give their vote to me, so i vote for them:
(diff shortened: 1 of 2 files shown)


fisx:
LGTM!  can be merged when travis is green.

fisx:
(i wouldn't have added the explanation in two different places for fear of diverging changes, but on second thoughts, the fact that it's easier to find may outweigh this hypothetical problem.)

fisx:
thanks!

andorp:
> (i wouldn't have added the explanation in two different places for fear of diverging changes, but on second thoughts, the fact that it's easier to find may outweigh this hypothetical problem.)

I thought those are the two occurences, even it would be better to use the `Delegate` and `Delegatee` newtype everywhere where it applies, but that would be a bigger change.

Issue #945, opened 2016-08-07
Could not find module ‘Paths_aula’


lwm:
Coming to the final step after the [Getting Started with Docker](https://github.com/liqd/aula#getting-started-with-docker) section, I got the following error:

```
root@f845c1d5515b:/liqd/aula# make aula-server 
`test -d .stack-work/ && echo "stack exec --" || echo "cabal exec --"` runhaskell -j5 -isrc -itests -iexec -idist/build/autogen ./exec/Aula.hs

src/Config.hs:72:18:
    Could not find module ‘Paths_aula’
    it is a hidden module in the package ‘aula-0.24.4@aula_Bw0IDGj4DsfJC1QtgFP2Co’
    Use -v to see a list of the files searched for.
make: *** [aula-server] Error 1
```

fisx:
thanks for reporting this.

i think the issue is that `stack install` does not keep the build artifacts under `./dist` where `runhaskell` is looking for them.  so you can try any of these two things:

1. call the `$HOME/.local/bin/aula-server` executable instead of `make aula-server`
1. run `stack build ...` instead of (or after) the `stack install ...` from the instructions.

let me know if that works and i'll fix the README.  of course another PR would be appreciated as well.  (-:

PR #997, opened 2016-11-14
Remove Ideaspaces.

andorp:
Fix: #945

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,7 +1,8 @@
 SHELL=/bin/bash
 EXEC=`test -d .stack-work/ && echo "stack exec --allow-different-user --" || echo "cabal exec --"`
 HLINT=$(EXEC) hlint
-AULA_SOURCES=-isrc -itests -iexec -idist/build/autogen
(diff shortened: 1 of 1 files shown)


andorp:
The `SchoolClass` can be found in the following data definitions:

 - Role
 - IdeaSpace
 
The `IdeaSpace` can be found in the following data definitions:

 - Topic
 - DScope
 - IdeaLocation

To remove an idea space, the implicit references from that idea space have to be cleared from the types listed above. My suggestion is

 - to create a default location and set all the referenced entities (like: user, topic, etc) to that location,
 - admin could add/change the user's role,
 - students has to set their ideaspace related delegations again after new idea spaces are created

andorp:
Correction based on the issue linked: Only ideas and topics has to be removed. Which makes life easier.

fisx:
`IdeaSpace`s should not be affected by this change.  we only want to delete ideas and topics.  so @andorp your above comment is off-topic (but still interesting).

PR #1000, opened 2016-11-21
class names with dashes

### LINKED ISSUE #999: class names with dashes are not supported, but allowed.
to reproduce, create a class with name `wef-wef` and arbitrary student data.  then try to `bearbeiten` that class.  you will get a 404.

fisx:
Fixes #999.


fisx:
tested manually.  i'll also test that a class created on v0.31.0 with dashes will be intact after the upgrade.

fisx:
> i'll also test that a class created on v0.31.0 with dashes will be intact after the upgrade.

done.  success!

andorp:
LGTM

Issue #1002, opened 2016-11-29
Rename and delete `IdeaSpace`s.  [5]


fisx:
- "delete" in this case means "destroy", i.e. there is no way of recovering deleted idea spaces.
- renaming of idea spaces breaks existing URLs: the renamed idea space cannot be reached any more under links saved anywhere in the text or outside of the aula instance.

approach 1:

1. Change ideaspace data type to carry AUID
2. Add admin page for renaming
3. Extend admin page for renaming with deleting

approach 2:

Global rename of the string directly, like with `userLogin.  This way there is no need for `AUID`.  Drawback: Need to find `IdeaSpace` by hand; old URL does not work any more.

CAVEATS:

Violation of hard assumptions: (1) user always has an idea space; (2) names for idea spaces will never change.  if student ends up without idea space, she will often have zero roles left.  admin can change that on request.

fisx:
*orphans* are users who have no roles.  since some users only have, say, a student role that depends on a school class, deleting that class will create an orphan.

two options:

1. fail to delete the class, and report a list of users that would become orphans.
1. move to a new page, list all soon-to-be orphans with a popup menu for role assignment, and force the admin to assign new roles to all students (not including the one to be deleted, obviously).  submit of this form will delete the class without creating orphans.

the latter is NTH.

fisx:
btw will this make an acid-state migration necessary?  if not i would like to deploy this once together with #1001 soon when it's done.

fisx:
@rittermo i updated the description to say that we do not support recovering of deleted ideas.  do you remember if we have agreed on that yet?  if not we should get it confirmed.  thanks!  (-:

fisx:
@rittermo (added another point).

fisx:
Fixed in #1011.

PR #1008, opened 2016-11-30
Rename `IdeaSpace`s

np:


np:
Work in progress towards #1002

np:
This has only the renaming part but this can be merged.

fisx:
what do you think about this change?:

```
@@ -477,7 +477,7 @@ newtype ClassName = ClassName { _unClassName :: ST }
 -- every school year.)
 data SchoolClass = SchoolClass
     { _classSchoolYear :: Int -- ^ e.g. 2015
-    , _className       :: ST  -- ^ e.g. "7a"
+    , _className       :: ClassName
     }
 
```

fisx:
The following should morally hold, but doesn't technically:

```haskell
fromList [Student (Just c), Student Nothing] === fromList [Student (Just c)]
```

Should we canonicalize this?  Or do you think it's not an issue?  (We should not use equality on role *sets* for authorization checks anyway, I guess.)

np:
@fisx 

Yes `_className :: ClassName` is planned.
I agree that ideally this should hold but I would not bother.

fisx:
@np ack to both answers, thanks.

fisx:
I've played with the styling and moved a validator to module `Frontend.Validation`.  The rest LGTM.  I'll merge when travis is green.

fisx:
thanks!

PR #1012, opened 2016-12-05
Fix #1010, no longer `reload` after changing a location in custom.js

### LINKED ISSUE #1010: jsRedirectOnClickConfirm does not redirect
So far it is only used where deleting an idea. It reloads but stays on the same page.

np:


fisx:
fixes things on firefox and chromium, and seems to follow the consensus on stackoverflow.  good catch, thanks!

PR #1014, opened 2016-12-06
Deriving `Data` for most types

np:


np:
Could we derive the `CoDelegationMap` from the `DelegationMap`?

andorp:
I would use the `setDelegation` functions, generate a list of `(userA, space, userB)` triplet and fold over the list.

np:
@andorp do you think that `gshrink` is correct or is it safer to leave it out.

fisx:
> @andorp do you think that gshrink is correct or is it safer to leave it out.

I guess the time to worry about this more is when we get failing test cases?  :-)

fisx:
LGTM!

I've fixed the `Delegations` shrinking issue after all.  Can be merged when travis agrees.

fisx:
thanks!

PR #1011, opened 2016-12-05
Handle removing school classes (issue #1002 part 2)

np:

PR #1016, opened 2016-12-07
Moving an idea should update comment keys

np:


np:
The new commits only start at "Refactoring to have a single place to change the location of an idea".
If we merge #1011 first, I'll rebase this one for clarity.

fisx:
I rebased, like all the changes, nit-picked a little, and will merge when travis is green.  Thanks!

PR #1019, opened 2016-12-09
Do not write standard "class" before group names

### LINKED ISSUE #1004: Do not write standard "class" before group names  [6]
Requires change of `instance HasUILabel SchoolClass` (if i recall correctly).

Please keep the `{From,To}HttpApiData` instance intact (related: #999).

np:
Fixes #1004

### DIFF
--- a/src/Types/Core.hs
+++ b/src/Types/Core.hs
@@ -808,7 +808,7 @@ instance Ord IdeaSpace where
 instance HasUILabel IdeaSpace where
     uilabel = \case
         SchoolSpace    -> "Schule"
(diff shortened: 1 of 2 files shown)


src/Types/Core.hs
```diff
@@ -808,7 +808,7 @@ instance Ord IdeaSpace where
 instance HasUILabel IdeaSpace where
     uilabel = \case
         SchoolSpace    -> "Schule"
-        (ClassSpace c) -> "Klasse " <> uilabel c
+        ClassSpace c   -> uilabel c
```
fisx: we should probably rule out "Schule" as a class name then.  i'll fix this.

PR #1021, opened 2016-12-12
Migration 1

### LINKED ISSUE #1020: Renaming users does not work (for me)
I get a 500. Internally the server complains about an empty CSRF token.

@fisx can you confirm? Or should I simply dig myself into that.

np:
This also fixes #1020


np:
I've tested the migration from the previous release.

fisx:
Thanks!  I was going to keep using TH, which would have meant migrating most of the types nested in `AulaData`.  This is better.  :-)

PR #1028, opened 2017-01-16
Initial passwords can now be exported in Xlsx

### LINKED ISSUE #1005: Printable list of initial passwords  [7]
- We need a template xmlx file to start with this.
- Options are HTML or xlsx (https://hackage.haskell.org/package/xlsx) output.

np:
Fixes #1005.

Things to consider:

* The end points names for csv and xlsx are not ideal (the csv endpoint changed)
* I propose to change the behavior of the export of students which no longer have an initial password. Currently they are not in the list at all and I propose to include them with an empty (or constant) value in the password column. The rational is to avoid getting an empty file when all students have their hashed password.


fisx:
> I propose to change the behavior of the export of students which no longer have an initial password. 

good idea.  i'll get a decision on that.

fisx:
- [ ] users with hashed passwords are already listed, let's keep it like that.  we should just say "<Passwort geschützt>" instead of "" in the password column.
- [ ] filename should not be "xlsx".   Ideally it would be "Passwoerter <school class>.xlsx".
- [ ] email column should be removed (it makes the entire table too wide when printing).
- [ ] sheet name isn't set when i open downloaded document in libreoffice.

I'll also get a decision on whether we want to keep the csv button.  It's nice we still have it, but it may be more distracting than helpful for the target audience.

fisx:
Otherwise LGTM, thanks!  (-:  I am happy about the xlsx-templater pull request.

np:
Your points in the same order:

* Users with a hashed password are not currently listed in master (https://github.com/liqd/aula/blob/master/src/Frontend/Page/Admin.hs#L1142), however they are in my commit since this easier to test since I have most users with hashed passwords. I'll add the "<Passwort geschützt>".

* About the filename, I try with a `Content-Disposition: inline; filename="Passwoerter.xlsx"`, if I can't make it work, I'll simply change the end point name.

* We can remove the email column simply by removing it from the template file, I'll do it.

* I've only tested with libreoffice and the sheet name shows up at the usual location of the UI.

fisx:
cool.

weird about the missing sheet name.  please ignore this point for now and i'll investigate if i have the PR back.

np:
The filename suggestion does not work for me. Additionally using curl, I don't see the headers being added. I think one of these two is happening:

* For some reason these response headers are not added
* They are added but only show up under some circumstances (query headers?) AND chrome fails to interpret the header

Does this issue rings a bell?
I can investigate more.

fisx:
> Does this issue rings a bell?

very faintly.  i could look into this, but i think i like your idea of just renaming the end-point best at this point

np:
I've fixed the remaining issues. Total time tracking on this issue 13 hours.

src/Types/Core.hs
```diff
@@ -635,14 +635,15 @@ data CSVZIP
...
+type ContentDisposition = "Content-Disposition"  -- appease hlint v1.9.22
+type AttachmentHeaders a = Headers '[Header ContentDisposition String] a
 
-instance MimeRender CSVZIP a => MimeRender CSVZIP (CsvHeaders a) where
+instance MimeRender CSVZIP a => MimeRender CSVZIP (AttachmentHeaders a) where
     mimeRender proxy (Headers v _) = mimeRender proxy v
 
-csvZipHeaders :: String -> a -> CsvHeaders a
-csvZipHeaders filename = addHeader $ "attachment; filename=" <> filename <> ".zip"
+-- FIXME Escaping?
```
fisx: `filename` is coming from the user.  i'm going to fix this now.

Makefile
```diff
@@ -94,6 +94,21 @@ fix-content:
 	curl -b cookie-jar -XPOST $(AULA_MK_CONTENT_URL)/api/manage-state/fix
 	rm -f cookie-jar
 
+testrenameclass:
```
fisx: i get this in the logfile when running this:

```
2017-02-03 15:44:58.427813 UTC [ERROR] [QN6NaFSUP6b9Y66FCvHn9vuo] 500 Ill-formatted CSRF Token CsrfToken {fromCsrfToken = ""}
```

cookie-jar looks fine, from what i can tell.

fisx:
I'm not getting headers in the xlsx file that show the class name; neither in the sheet name in libreoffice nor in the PDF export.  I wonder what's different from what you're doing?  I will let you know if we need to fix this, perhaps it's good enough as it is.

fisx:
Thanks!

PR #1029, opened 2017-01-16
2017 01 16 np reporter name

np:
This needs a bit of translation and potentially to extend the confirmation text.


fisx:
I don't get a confirmation dialog, and I don't see any code about that.  (It wasn't there before and you suggested to add it, correct?)

Also translation and one small change, otherwise LGTM.

fisx:
> I don't get a confirmation dialog, and I don't see any code about that. (It wasn't there before and you suggested to add it, correct?)

what's with this?

fisx:
I'm still interested in what became of the notification popup, but I'm going to merge this.  It works fine with the small layout fix I made.

fisx:
...  and thanks again (-:

